PR #1, opened 2016-06-17
add nginx setup

jamesgorrie:
To be able to run `media-atom-maker.gutoosl.co.uk`
We could just use `media-maker.`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 application.conf
 logs
+.idea/
+target/
--- a/nginx-mappings.yml
+++ b/nginx-mappings.yml
@@ -1,0 +1,4 @@
+name: media-atom-maker
+mappings:
+  - prefix: media-atom-maker
+    port: 9001


akash1810:
> We could just use `media-maker`.

`media-atom` would fit better as it deals with the whole lifecycle of the atom?

jamesgorrie:
> media-atom

I personally don't like using the word atom, as that is part of the underlying technology - not really descriptive of what this is actually doing. Similar to [quiz builder](https://github.com/guardian/ten-four_quiz-builder).

paulmr:
:+1: I prefer media-maker to media-atom, although perhaps it is a bit confusing in the sense that it doesn't actually create media. I think my preference is still media-atom-maker, but we need to make a decision so frankly I am happy with any of them!

clloyd:
Are there plans for it to be a user facing tool, or more of an api with a frontend for people when they need to manually tweak stuff, because `media-atom-maker` could make sense in that situation.

akash1810:
> Are there plans for it to be a user facing tool

Yep.

majority rules - `media-atom-maker`


👍

clloyd:
If it's user facing then I don't think we should use `atom`

paulmr:
I think we can always change it later if it becomes a problem.

paulmr:
Also, use master now instead of `pmr/initial-version`!

akash1810:
@clloyd this would be for DEV nginx setup. The PROD cname will have a nicer/clearer entry (which also makes me question why I raised the initial concern.... 😜)

clloyd:
@akash1810 DEV and PROD need to match or I'll forget them :D

paulmr:
@clloyd agreed, but at least if we change it when going to prod, only devs will notice the change, and they can handle it. :)

PR #3, opened 2016-06-29
Dev stack

jamesgorrie:
Just adding the ability to create the DEV stack to run locally.
Dynamo is all we need for the time being.

### DIFF
--- a/cloudformation/convert-yaml-to-json.rb
+++ b/cloudformation/convert-yaml-to-json.rb
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+
+require 'json'
+require 'yaml'
+
+if ARGV.size != 1
+  puts "usage: convert.rb <file.yml>"
+  exit(1)
+end
+
+input_filename = ARGV[0]
+output_filename = input_filename.sub(/(yml|yaml)$/, 'json')
+
+input_file = File.open(input_filename, 'r')
+input_yml = input_file.read
+input_file.close
+
+output_json = JSON.pretty_generate(YAML::load(input_yml))
+
+output_file = File.open(output_filename, 'w+')
+output_file.write(output_json)
+output_file.close
--- a/cloudformation/media-atom-maker-dev.yml
+++ b/cloudformation/media-atom-maker-dev.yml
@@ -45,6 +45,9 @@ Outputs:
   EditsDynamoTable:
     Value:
       Ref: "MediaAtomsDynamoTable"
-  HostKeys:
-      Value:
-        Ref: "HostKeys"
+  AwsId:
+    Value:
+      Ref: "HostKeys"
+  AwsSecret:
+    Value:
+      Fn::GetAtt: ["HostKeys", "SecretAccessKey"]
--- a/cloudformation/media-atom-maker-dev.yml
+++ b/cloudformation/media-atom-maker-dev.yml
@@ -19,13 +19,7 @@ Resources:
                 Resource:
                   - Fn::Join:
                     - ""
-                    -
-                      - "arn:aws:dynamodb:"
-                      - Ref: "AWS::Region"
-                      - ":"
-                      - Ref: "AWS::AccountId"
-                      - ":table/"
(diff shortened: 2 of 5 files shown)


jamesgorrie:
@paulmr @akash1810

cloudformation/convert-yaml-to-json.rb
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env ruby
```
akash1810: ♦️ 😮

cloudformation/media-atom-maker-dev.yml
```diff
@@ -0,0 +1,53 @@
...
+        - AttributeName: "id"
+          AttributeType: "S"
+      KeySchema:
+        - AttributeName: "id"
+          KeyType: "HASH"
+      ProvisionedThroughput:
+        ReadCapacityUnits: "10"
+        WriteCapacityUnits: "5"
+Outputs:
+  EditsDynamoTable:
```
akash1810: should the `HostKeys` be outputted too? As I don't think the secret is viewable from within the web console.

cloudformation/media-atom-maker-dev.yml
```diff
@@ -0,0 +1,53 @@
...
+        - PolicyName:   "Group-Policy"
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: "Allow"
+                Action: ["dynamodb:*"]
+                Resource:
+                  - Fn::Join:
+                    - ""
+                    -
```
akash1810: can do the `Fn::Join` on one line to avoid this annoying feature of yaml:

```yaml
Resource:
  Fn::Join: ['', ['arn:aws:dynamodb:', {Ref: AWS::Region}, ':', {Ref: AWS::AccountId}, ':table/', {Ref: AWS:: MediaAtomsDynamoTable} ]]
```

cloudformation/convert-yaml-to-json.rb
jamesgorrie: Built in yaml and json libs

cloudformation/media-atom-maker-dev.yml
jamesgorrie: I actually preferred it this way as it's relatively readable. Happy to change.

cloudformation/media-atom-maker-dev.yml
jamesgorrie: Good call

cloudformation/media-atom-maker-dev.yml
jamesgorrie: What about this:
```YAML
- Fn::Join:
  - ""
  - ["arn:aws:dynamodb:", {Ref: "AWS::Region"}, ":", {Ref: "AWS::AccountId"}, ":table/", {Ref: "MediaAtomsDynamoTable"}]
```

Which feels like it keeps the line length down and says "Join with this ("") on that (blahblah)"

akash1810: nice!

cloudformation/media-atom-maker-dev.yml
akash1810: I think you'd need another block for the secret as its behind a `Fn::GetAtt` http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-iam-accesskey.html ?

akash1810:
question on whether you'd need to output the secret too, otherwise 👍

jamesgorrie:
>  output the secret

Which secret?

akash1810:
`AWS::IAM::AccessKey` creates a AccessKeyId/SecretAccessKey for auth `Ref::HostKeys` will return the  AccessKeyId. We'd need the SecretAccessKey too to talk to AWS right?

jamesgorrie:
@akash1810 Good point

akash1810:
👍

paulmr:
Nice work 👍

PR #4, opened 2016-06-30
use Cats's `Xor` instead of returning unit and throwing exceptions

paulmr:
Provides better code-documentation of the errors that might result from the API interactions, and makes the code easier to read due to less surprises.

Thanks @philwills for the suggestion. (Would appreciate you casting your eye over this to make sure I have done it right ...)


sihil:
Definitely a good idea - I've not used Cats Xor yet so would be interested in hearing how you found it. For the same reason I'm not sure I'm the best person to give this a plus one so will leave it to @philwills

philwills:
👍

paulmr:
Thanks @sihil and @philwills

PR #8, opened 2016-07-13
[WIP] Publish to preview stream 

Reettaphant:
- Publish media atom to preview kinesis stream when new content is added to it

@paulmr

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import com.gu.atom.publish.AtomPublisher
+import com.gu.atom.publish.{LiveAtomPublisher, PreviewAtomPublisher}
 import com.gu.contentatom.thrift.{ ContentAtomEvent, EventType }
 import com.gu.contentatom.thrift.atom.media.Asset
 import com.gu.contentatom.thrift.{ Atom, AtomData }
@@ -23,7 +23,8 @@ import scala.util.{ Success, Failure }
 import data.JsonConversions._
 
 class Api @Inject() (val dataStore: DataStore,
-                     val publisher: AtomPublisher,
+                     val livePublisher: LiveAtomPublisher,
+                     val previewPublisher: PreviewAtomPublisher,
                      val conf: Configuration,
                      val authActions: AuthActions)
     extends AtomController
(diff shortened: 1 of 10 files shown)


app/controllers/Api.scala
```diff
@@ -69,7 +71,16 @@ class Api @Inject() (val dataStore: DataStore,
...
+          _ => {
+            Created(s"updated atom $atomId").withHeaders("Location" -> atomUrl(atom.id))
+
+            val event = ContentAtomEvent(atom, EventType.Update, now())
+
+            previewPublisher.publishAtomEvent(event) match {
+              case Success(_)  => NoContent
+              case Failure(err) => InternalServerError(jsonError(s"could not publish: ${err.toString}"))
+            }
+          }
```
jamesgorrie: Should we be doing this in `createMediaAtom` too?

Reettaphant: Yes I think so. I forgot all about that.

paulmr:
Let's add the new liveStream and previewStream settings to the `reference.conf` file as a pointer to users that they need to be filled in.

Also we will need to update the configuration files on S3 that are used for deployment. (Just stating that as a reminder for after this is deployed, not something we need to do as part of the PR).

app/controllers/Api.scala
```diff
@@ -69,7 +80,16 @@ class Api @Inject() (val dataStore: DataStore,
 
         dataStore.updateMediaAtom(newAtom).fold(
           err => InternalServerError(err.msg),
-          _ => Created(s"updated atom $atomId").withHeaders("Location" -> atomUrl(atom.id))
+          _ => {
+            
+            val event = ContentAtomEvent(atom, EventType.Update, now())
```
paulmr: I think this should be `newAtom`? Current it is publishing the state of the atom before the asset is added.

test/ApiSpec.scala
```diff
@@ -121,11 +128,21 @@ class ApiSpec
...
       }
     }
-    "call out to publisher to publish an atom" in withApi() { api =>
+    "call out to live publisher to publish an atom" in withApi() { api =>
       val result = call(api.publishAtom("1"), requestWithCookies(api))
       status(result) mustEqual NO_CONTENT
     }
-    "call report failure if publisher fails" in withApi(publisher = failingMockPublisher) { api =>
+
+    "call out to preview publisher when adding an asset" in withApi() { api =>
```
paulmr: Let's add a test here to catch the situation mentioned in the above comment: so we can use the mock publisher `verify()` to confirm that the atom update includes the new asset.

paulmr:
Nice work! :+1:

PR #9, opened 2016-07-13
Ignore auth in dev environment

kelvin-chappell:
/cc @paulmr

### DIFF
--- a/app/di.scala
+++ b/app/di.scala
@@ -1,9 +1,13 @@
 import com.google.inject.AbstractModule
+import com.google.inject.Stage.DEVELOPMENT
 import com.gu.pandomainauth.action.AuthActions
 
 class Module extends AbstractModule {
   def configure() = {
-    bind(classOf[AuthActions])
-    .to(classOf[controllers.PanDomainAuthActions])
+    if (currentStage() == DEVELOPMENT) {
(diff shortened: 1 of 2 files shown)


app/di.scala
```diff
@@ -1,12 +1,19 @@
...
 import com.gu.atom.publish.AtomPublisher
 import com.gu.pandomainauth.action.AuthActions
 import data.AtomPublisherProvider
 
 class Module extends AbstractModule {
   def configure() = {
-    bind(classOf[AuthActions])
-    .to(classOf[controllers.PanDomainAuthActions])
+    
+    if (currentStage() == DEVELOPMENT) {
```
paulmr: do we need to do anything to make sure that this is `PRODUCTION` on CODE?

paulmr:
as long as this doesn't disable panda on CODE, :+1:

app/di.scala
kelvin-chappell: I'll check.

jamesgorrie:
Why is this a good idea? Aren't we opening our selves to breaking things in PROD that we won't catch in Dev?

kelvin-chappell:
@jamesgorrie I think you might be right.  There's a greater risk of building unsecured endpoints etc with this change.  We should keep dev as similar to prod as possible.  I'll take @paulmr's advice and get access to to the S3 bucket instead.

Issue #7, opened 2016-07-13
AsyncHttpClient error during tests


paulmr:
Panda isn't being shutdown properly during API tests, resulting in this error message during testing:

```
[error] o.a.DefaultAsyncHttpClient - AsyncHttpClient.close() hasn't been invoked, which may produce file descriptor leaks
[error] o.a.DefaultAsyncHttpClient - AsyncHttpClient.close() hasn't been invoked, which may produce file descriptor leaks
[error] o.a.DefaultAsyncHttpClient - AsyncHttpClient.close() hasn't been invoked, which may produce file descriptor leaks
```

paulmr:
See PR #10

PR #10, opened 2016-07-13
Only create one Injector during testing, and shut it down properly

paulmr:
This reorganises the tests so that we only create one Injector for each test, while still allowing us to override the bindings for some tests (you can only override once, so you can't just add overrides to the existing module).

In addition, this runs shtudown code after each test, which fixes issue #7.

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -30,7 +30,7 @@ class Api @Inject() (val dataStore: DataStore,
     extends AtomController
     with MediaAtomImplicits {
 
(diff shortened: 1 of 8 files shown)


paulmr:
I know it's a bit cheeky but I am going to just go ahead and merge this shortly as it makes the tests much better so I would really to get it in before master diverges so much that this becomes unmergable!

paulmr:
I will let @Reettaphant merge PR#8 first though as my stuff will cause conflicts with hers so I will fix them first before merging this.

PR #11, opened 2016-07-18
Add initial reindex endpoint

paulmr:
This is an initial implementation of the required reindex endpoint, which will allow [CAPI Floodgate](https://floodgate.capi.gutools.co.uk/) to trigger the atom maker to resend all of it's atoms onto a separate re-indexing stream.

This is part of the the contract that CAPI defines for incoming content sources, and which is described [here](https://github.com/guardian/floodgate/blob/master/docs/implementing-a-content-source.md).

This PR is a minimal implementation, it does not include the optional items (cancelling a request and limiting the query by start and end date). However, what's included here should be sufficient for an MVP.

I have included support for the API authentication option (as described in the doc).

The implementation of the reindexing is quite simple, as it just pushes a list of atoms to kinesis. Most of the work (including reporting status of an ongoing job, and reporting an error if a new job is requested while one is already in progress) is done in the API endpoints which are still part of the main application (not yet separated out into a separate library). Creating a separate `atom-manager-play` library is still a very good idea I think to allow reuse of these standard endpoints.

### DIFF
--- a/app/util/AWS.scala
+++ b/app/util/AWS.scala
@@ -34,6 +34,8 @@ class AWSConfig @Inject() (config: Configuration) {
 
   lazy val kinesisStreamName = config.getString("aws.kinesis.streamName").get
 
+  lazy val kinesisReindexStreamName = config.getString("aws.kinesis.reindexStreamName").get
+
   lazy val kinesisClient = region.createClient(
     classOf[AmazonKinesisClient],
     credProvider,
--- a/app/controllers/ReindexController.scala
+++ b/app/controllers/ReindexController.scala
@@ -2,6 +2,8 @@ package controllers
 
 import akka.actor.{ ActorSystem, Props }
 import com.gu.atom.publish.AtomReindexer
+import com.gu.contentatom.thrift.{ ContentAtomEvent, EventType }
+import play.api.Configuration
 
 import play.api.mvc._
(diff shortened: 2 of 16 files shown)


app/util/AWS.scala
```diff
@@ -34,6 +34,8 @@ class AWSConfig @Inject() (config: Configuration) {
 
   lazy val kinesisStreamName = config.getString("aws.kinesis.streamName").get
 
+  lazy val kinesisReindexStreamName = config.getString("aws.kinesis.reindexStreamName").get
```
LATaylor-guardian: @paulmr FYI we decided that for all atoms we would use a single kinesis stream for indexing and reindexing (which are already defined, see here: https://github.com/guardian/content-api/blob/master/porter/src/main/scala/com.gu.contentapi.porter/Config.scala#L68)

So you may just use this one. Was this what you was planning?

paulmr: Thanks @LATaylor-guardian , no I hadn't realised that.

With that in mind though, I think it still makes sense to have the option in the code to use different streams, but we can just put the same stream in the configuration file. Does that make sense?

LATaylor-guardian: Yes of course, I just stuck this comment here as it was when I was reading this that it popped into my head :)

PR #13, opened 2016-07-18
Move atom-maker to media service account

Reettaphant:


app/util/AWS.scala
```diff
@@ -24,6 +25,15 @@ class AWSConfig @Inject() (config: Configuration) {
 
   lazy val credProvider = new AWSCredentialsProviderChain(credProviders.flatten: _*)
 
+  lazy val sessionId: String = "session" + Math.random()
+
+  lazy val kinesisArn = config.getString("aws.kinesis.streamArn").getOrElse("")
+
+  lazy val atomsCredProvider = new AWSCredentialsProviderChain(
+    new ProfileCredentialsProvider("composer"),
```
akash1810: should this be `media-service` rather than `composer`?

Reettaphant: The code and prod kinesis streams are in the composer account. Having the composer credentials provider here should allow people to access the code/prod streams locally.

akash1810: ah ok

Is it possible to use AssumeRole to gain access? Just to limit the number of aws profiles needed to run the service.

jamesgorrie:
Quick one - shouldn't we be moving this to multimedia? media-service is a fairly misleading name.

app/util/AWS.scala
Reettaphant: Yes, I can do that as well.

Reettaphant: @akash1810 I can't get it working actually it doesn't seem to be the way its done in other projects. Having thought about it, it is probably not the simplest way of doing it. Being able to read from the code/prod streams in this way would require adding the assume role arn to the configuration file and filling into the cloudformation. Copying composer credentials seems simpler. The composer credentials would not be needed to run the app, because there are kinesis streams for local development in the media-service account. Getting these credentials would only be needed for reading from the code/prod streams.

README.md
```diff
@@ -3,8 +3,16 @@ This app has two functions: it is a simple Play app for editing media atoms,
 and it's a publisher that pushes media atoms into a Kinesis stream.
 
 ## Setting up dev environment
+0. Set up a dev stack by generating a `media-atom-maker-dev.json` template from the corresponding yam file
```
gidsg: typo: `.yml` file

app/util/AWS.scala
```diff
@@ -37,9 +47,17 @@ class AWSConfig @Inject() (config: Configuration) {
 
   lazy val kinesisReindexStreamName = config.getString("aws.kinesis.reindexStreamName").get
 
-  lazy val kinesisClient = region.createClient(
+  lazy val stage = config.getString("stage").getOrElse("DEV")
+  lazy val readFromComposerAccount = config.getString("readFromComposer").getOrElse("false")
```
gidsg: I think you can do `config.getBoolean` and avoid the string conversion later on.

Reettaphant:
@jamesgorrie I think that @paulmr investigated this and decided that this should live in media-service, but I'm happy to move it elsewhere as well.

paulmr:
@jamesgorrie could you say a bit more about the multimedia stack? I am not familiar with it. What else is in it? @Reettaphant and I don't currently have access to it but when we discussed it with the video team on slack a few weeks back we decided on this one, but I don't think multimedia came up as an option.

From our point of view the main goal is to get it out of the composer stack so that we don't have to give full access to composer for media-atom devs, so which one it actually goes in is not a big deal.

jamesgorrie:
> don't have to give full access to composer for media-atom devs
👍 

> a bit more about the multimedia stack

There's quite a lot in there, lots to do with the pluto stuff and other surrounding systems. Honestly, it's a bit messy, a lot of it isn't cloudformed, and a few other rough edges.

The reason I suggest it though is that:
* it is essentially a tools stack (all video stuff at the moment is tools), so inline with keeping to the stream
* pluto integration is going to have to happen at some stage, and this might make things easier
* there's a small piece around as we start to chip away the cost of the multimedia account should drop even though we are adding more services as they should be more efficient.

Although - this is only because I don't know if there is a plan to have a multimedia team i.e. video / audio / images. I would 100% say video and audio go together. Images in that lot, I personally think is tenuous.

Happy to add you to the stack to have a poke around (or add yourself in janus, and we'll get it sorted).

app/util/AWS.scala
akash1810: Does this mean the CODE/PROD instances will have a set of credentials in the config files?

akash1810:
> pluto integration is going to have to happen at some stage, and this might make things easier

Just to say, I'd envisage any integration between Pluto and a.n.other service to be done via a queue rather than talking to Pluto directly as the core of Pluto is run in the basement and not in AWS and setting up direct connect etc would likely be too much effort for the end result.

app/util/AWS.scala
Reettaphant: I'm not sure what you mean. If someone wanted to use a different kinesis stream, then they'd have to change the name and arn of the stream they were using, and if they wanted to use CODE/PROD locally, they'd have to, with this setting, get composer credentials, and set a variable in the configuration file to true.

paulmr:
My main concern about using the multimedia stack is that it isn't owned by tools. And also I think integration with pluto shouldn't be a problem because @Reettaphant has already got cross-stack access working, and as @akash1810 says, we might not need it anyway if there is an SQS queue or something similar that will mediate the access.

Add to that the fact that it shouldn't be too hard to move it later on, and for now we just need to put it somewhere so we can get it moving, so I think we can settle on the media stack for now.

Thanks all!

jamesgorrie:
this might be worth picking up with Blishen as I know there is some immanent talk about merging teams etc, so perhaps worth having that conversation before doing this and potentially having to do t again.

paulmr:
:+1:

PR #14, opened 2016-07-18
Add organization field to build.sbt.

philmcmahon:
This changes the directory that the library is published to and should make it easier to find in maven.


paulmr:
:+1:

PR #17, opened 2016-07-19
Create a separate reusable library for the play actions

paulmr:
This creates a second reusable library, this time for the play actions. This will allow consumers to decide whether they want to use all of our reusable code to jumpstart their atom manager, or to pick just the publisher/kinesis bits (e.g. if they don't use play and therefore want to implement the actual endpoints themselves).

This change also creates some intitial content for this library: I have moved the Reindex Enpoint, it's tests, and the general traits that they are both built upon, into the library.


philmcmahon:
👍

PR #19, opened 2016-07-21
Update media atom model

kelvin-chappell:
These are just the changes necessary to support model v2.2.0.

NB. As `activeVersion` is now optional, this is going to break any atoms currently in storage.

/cc @guardian/labs-beta

### DIFF
--- a/app/data/JsonConversions.scala
+++ b/app/data/JsonConversions.scala
@@ -2,27 +2,43 @@ package data
 
 import com.gu.contentatom.thrift._
 import com.gu.contentatom.thrift.atom.media._
-import play.api.libs.json._
 import play.api.libs.functional.syntax._
+import play.api.libs.json._
 
 object JsonConversions {
 
   implicit val atomType = Writes[AtomType](at => JsString(at.name.toLowerCase))
 
+  implicit val category = Writes[Category](category => JsString(category.name.toLowerCase))
+
   implicit val mediaAsset = (
     (__ \ "id").write[String] and
-      (__ \ "version").write[Long] and
-      (__ \ "platform").write[String] and
-      (__ \ "assetType").write[String]
+    (__ \ "version").write[Long] and
+    (__ \ "platform").write[String] and
+    (__ \ "assetType").write[String] and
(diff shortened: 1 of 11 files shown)


akash1810:
why are we making `activeVersion` optional?

kelvin-chappell:
@akash1810 Not sure: https://github.com/guardian/content-atom/commit/a0bcac7444f10cdbe4cf92c56df4bdad69690df4

app/data/JsonConversions.scala
```diff
@@ -2,35 +2,78 @@ package data
...
 
   implicit val atomDataMedia = (
     (__ \ "assets").write[Seq[Asset]] and
-      (__ \ "activeVersion").write[Long]
-  ) { mediaAtom: MediaAtom =>
-    (mediaAtom.assets, mediaAtom.activeVersion)
+    (__ \ "activeVersion").writeNullable[Long] and
+    (__ \ "title").write[String] and
+    (__ \ "category").write[Category] and
+    (__ \ "plutoProjectId").writeNullable[String] and
```
akash1810: Is this a true pluto project id? or a vidispine id? (vidispine ids look like `KP-XXXXX`)

Composer has [knowledge of vidispine ids](https://github.com/guardian/flexible-content/blob/2a2da925aa778e77de9567e6a5fa204e50e93067/flexible-content-integration/src/main/scala/com/gu/flexiblecontent/integration/integration/xmlprocessors/IdentifierProcessor.scala#L30) so if they're the same, would be good to use the same terminology for simplicity.

kelvin-chappell: @paulmr Can you answer this one?  It looks like the prop was added here: https://github.com/guardian/content-atom/commit/21247f54c677d5506e238fa2046917af2822f2c4

jamesgorrie:
@akash1810 What if we don't have an activeVersion yet?

We could potentially look at having an `active` on an asset? That way we could actually have an active YT id as well as an active URL?

jamesgorrie:
👍

PR #21, opened 2016-07-25
Support extra properties in media atoms

kelvin-chappell:
This change supports some extra properties in the editor:
* title
* category
* duration

/cc @paulmr @Reettaphant


paulmr:
Looks good to me: :+1:

PR #23, opened 2016-07-27
Set play to listen on port 9001 in dev environment

kelvin-chappell:
#20 didn't work very well.
So using a local sbt script instead.


paulmr:
:+1:

PR #25, opened 2016-07-28
Support video poster

kelvin-chappell:
Adds support to api and editor for the [posterUrl property](https://github.com/guardian/content-atom/pull/33).

Also updated some dependencies.

/cc @guardian/labs-beta @akash1810

### DIFF
--- a/app/model/ThriftUtil.scala
+++ b/app/model/ThriftUtil.scala
@@ -67,6 +67,7 @@ object ThriftUtil {
     }
     val duration = params.get("duration").map(_.head.toLong)
     val source = params.get("source").map(_.head)
+    val posterUrl = params.get("posterUrl").map(_.head)
     for {
       assets <- parseAssets(
         params.get("uri").getOrElse(Nil),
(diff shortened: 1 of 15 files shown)


app/model/ThriftUtil.scala
```diff
@@ -67,19 +67,18 @@ object ThriftUtil {
     }
     val duration = params.get("duration").map(_.head.toLong)
     val source = params.get("source").map(_.head)
+    val posterUrl = params.get("posterUrl").map(_.head).map(parseId).flatMap(_.right.toOption)
```
paulmr: do we need to use `parseId` here? I am assuming that all we want to do is check that it is a URL, and that it will never need to be distinguished as a youtube id? In that case maybe we can just apply `Url` directly.

paulmr:
:+1:

app/model/ThriftUtil.scala
kelvin-chappell: Yes you're right.  That would be simpler.  I'll do it.

akash1810:
👍

(was on leave yesterday, hence a little late to review)

PR #26, opened 2016-08-01
Use scanamo-scrooge to make DynanoDB code generic

paulmr:
This PR uses [scanamo-scrooge](https://github.com/guardian/scanamo-scrooge) to allow storing of the atom into the dynamo db, using it's fully indexable schema, based on the Thrift model. This makes it much more generic as you need only write a very small couple of lines of code to identify your atom type and everything else comes automatically out of the thrift definition.

As such, this change also moves the Dynamo-based `DataStore` implementation into the separate library for reuse by other tools.

Ideally this will depend on later versions of  `scanamo` and `scanamo-scrooge` because it needs some fixes that are PR'd for those libraries. However, until these new versions are published, I have put a couple of shims in `ScanamoUtil.scala` to get this up and running with the current versions.


philmcmahon:
👍 💯 👍

PR #27, opened 2016-08-02
Fix revision check of update

paulmr:
Now that we are storing a proper nested structure for the atom in the database, the version check no longer works because the revision is nested in the `contentChangeDetails`.

In addition to fixing this, this PR creates some tests that use a local DynamoDB for testing, which actually reproduces this issue and generally exercises code that wasn't being tested until now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ application.conf
 logs
 .idea/
 target/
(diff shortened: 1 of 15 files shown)


clloyd:
👍

PR #28, opened 2016-08-04
Store published time in atom when it is published

paulmr:
* Move the publish endpoint into the `atom-manager-play` library for re-use
* Bump the version of mockito because it was causing a memory leak problem that meant you couldn't run the tests more than a couple of times
* When publishing store the time in the `contentChangeDetails` field of the atom
* Test the same


clloyd:
👍

PR #30, opened 2016-08-05
bugfix: need to update revision after publish

paulmr:
Bug fix for a bug introduced by PR#28, need to bump ther version when doing the update otherwise it will fail.


Reettaphant:
👍

PR #31, opened 2016-08-12
Add tests to atom manager play artifact

Reettaphant:
Add tests to the atom manager play artifact so that the resources it provides for writing tests can be used by apps that use the library.


paulmr:
:+1:

PR #32, opened 2016-08-16
remove unsused file

Reettaphant:
The ReindexActor has been moved to the atom util library, but the original file was never deleted from the project.


paulmr:
:+1:

PR #33, opened 2016-08-23
fix displaying errors

Reettaphant:
Display errors correctly on the client. ResponseJSON does not exist in the error response and the client was never displaying the error alerts.


paulmr:
:+1:

PR #34, opened 2016-08-23
increase revision number of new atom when creating it instead of active version

Reettaphant:
Fixes the following bug: 

- atoms that have already been published cannot be updated

This is because the active version was being used as the revision for the new atom, causing the revision validation check to fail later.


paulmr:
:+1:

PR #35, opened 2016-08-23
Rv published to separate dynamo table

Reettaphant:
Saves published atoms to a separate dynamo table

This allows for reindexing to push atoms in the correct kinesis streams (preview/live)

It also allows for displaying published atom information on the client

### DIFF
--- a/atom-manager-play-lib/src/main/scala/com.gu.atom/play/AtomAPIActions.scala
+++ b/atom-manager-play-lib/src/main/scala/com.gu.atom/play/AtomAPIActions.scala
@@ -22,11 +22,16 @@ trait AtomAPIActions extends Controller {
 
   def publishAtom(atomId: String) = Action { implicit req =>
 
+    val revisionNumber = publishedDataStore.getAtom(atomId) match {
+      case Some(atom) => atom.contentChangeDetails.revision + 1
+      case None => 1
+    }
+
     previewDataStore.getAtom(atomId) match {
       case Some(atom) => {
         val updatedAtom = atom.copy(
           contentChangeDetails = atom.contentChangeDetails.copy(published = Some(ChangeRecord((new Date()).getTime(), None)))
-        ).withRevision(1)
+        ).withRevision(revisionNumber)
 
         saveAtom(updatedAtom)
       }
--- a/atom-manager-play-lib/src/main/scala/com.gu.atom/play/AtomAPIActions.scala
+++ b/atom-manager-play-lib/src/main/scala/com.gu.atom/play/AtomAPIActions.scala
@@ -33,13 +33,13 @@ trait AtomAPIActions extends Controller {
           contentChangeDetails = atom.contentChangeDetails.copy(published = Some(ChangeRecord((new Date()).getTime(), None)))
         ).withRevision(revisionNumber)
 
-        saveAtom(updatedAtom)
+        savePublishedAtom(updatedAtom)
       }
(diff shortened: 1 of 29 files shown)


atom-manager-play-lib/src/main/scala/com.gu.atom/play/AtomAPIActions.scala
```diff
@@ -15,30 +15,41 @@ trait AtomAPIActions extends Controller {
...
+          contentChangeDetails = atom.contentChangeDetails.copy(published = Some(ChangeRecord((new Date()).getTime(), None)))
+        ).withRevision(revisionNumber)
+
+        saveAtom(updatedAtom)
+      }
+      case None => NotFound(jsonError(s"No such atom $atomId"))
+    }
+  }
+
+  private def saveAtom(updatedAtom: Atom) = {
```
paulmr: could we change the name of this method to something like `savePublishedAtom`, or something more descriptive? Especially as it actually does the publishing as well as saving it.

atom-publisher-lib/src/main/scala/com/gu/atom/data/DataStore.scala
```diff
@@ -24,7 +24,6 @@ trait DataStore extends DataStoreResult {
   /* this will only allow the update if the version in atom is later
    * than the version stored in the database, otherwise it will report
    * it as a version conflict error */
-  def updateAtom(newAtom: Atom): DataStoreResult[Unit]
```
paulmr: why was this moved into `(Preview|Published)DataStore` traits? Is it to cover the fact that if you just have a data store you can't update an atom, you need to know whether you have a preview or a published data store?

atom-publisher-lib/src/test/scala/com/gu/atom/data/DynamoDataStoreSpec.scala
```diff
@@ -23,35 +23,47 @@ class DynamoDataStoreSpec
     with BeforeAndAfterAll
     with AtomImplicitsGeneral {
   val tableName = "atom-test-table"
+  val publishedTableName = "published-atom-test-table"
 
-  type FixtureParam = DynamoDataStore[MediaAtom]
+  type FixtureParam = (PreviewDynamoDataStore[MediaAtom], PublishedDynamoDataStore[MediaAtom])
```
paulmr: could we make this a case class with some named members, just to make the code a bit more readable (to avoid referencing `_1` etc)

Something like:

```scala
case class DataStores(preview: PreviewDynamoDataStore[MediaAtom],
  published: PublishedDynamoDataStore[MediaAtom])
type FixtureParam = DataStores
```

paulmr:
I just had some minor comments inline, but otherwise, this is great stuff: 👍

atom-publisher-lib/src/main/scala/com/gu/atom/data/DataStore.scala
Reettaphant: Yes, I thought that it would make sense since they both have their own ways of updating atoms.

paulmr: if they have the same api but different implementations I think it makes sense to keep the method description in the parent interface. But I that's just my opinion, I am happy either way!

Reettaphant: Yes, I think that makes sense. I'll change it now.

PR #36, opened 2016-08-24
wait for async call() to complete in AtomAPIActions test

paulmr:
without this fix there is a race condition, that if you check for interactions before waiting for the api call to complete, you have random test failures if the check happens before the completion.


Reettaphant:
👍

PR #39, opened 2016-08-31
remove version from play lib build.sbt

Reettaphant:
Removes the version from play library build.sbt so that the version stays in sync with publisher.


paulmr:
This is the correct thing to do now. For the record though: we need to move these libraries out into their own repos soon, at which point we'll need our own version here again.

👍

PR #40, opened 2016-09-01
add poster url to media atom json

Reettaphant:
Add poster url to media atom json so that it is displayed on the client.


paulmr:
👍

PR #43, opened 2016-09-05
publish tests as well

Reettaphant:
Publish atom publisher tests as well so that the tests will have access to AtomData


paulmr:
:+1:

PR #44, opened 2016-09-05
fix url field validations

Reettaphant:
Fix url form field validations:
Poster url and asset url should accept https urls.


steppenwells:
You might want to rename the poster-url-validation.js file to url-validation.js

I suspect the old youtube url validator might be unused now, if so is it worth deleting?

Otherwise looks sensible 👍

Issue #41, opened 2016-09-02
Duplicate asset version


Reettaphant:
The app allows for adding the same asset version for different assets on the same platform. This means that one of the assets with the duplicate version would never be used. There should be checks on the client and on the server to make sure this does not happen.

mbarton:
@Reettaphant I think we need to support this behaviour since that is how different format/quality (mp4, m3u8 etc) for self-hosted videos is encoded

PR #45, opened 2016-09-06
Asset version validation

Reettaphant:
- Adds validation for media atom asset version to avoid problems described here: https://github.com/guardian/media-atom-maker/issues/41


paulmr:
:+1:

PR #46, opened 2016-09-06
UI improvements

Reettaphant:
Makes to improvements to the UI: 

- Redirects to the atom page when a new atom is created
- Removes some styling to make the atom embed link easier to copy


paulmr:
👍

PR #47, opened 2016-09-14
Remove server side activeAsset Incrementing

clloyd:
This fixes the unpredictable activeAsset version when clicking save. It will either use the value supplied from the client, or default to the highest version in the improbable case it doesn't have an active asset.

@kelvin-chappell @steppenwells


steppenwells:
👍

clloyd:
@Reettaphant This was the fix FYI

PR #58, opened 2016-10-06
Add beta route and React frontend

ShaunYearStrong:
Adds a `/beta` route that points at a simple React app, managed with jspm.


public/beta/jspm.config.js
```diff
@@ -0,0 +1,572 @@
...
+  paths: {
+    "npm:": "jspm_packages/npm/",
+    "github:": "jspm_packages/github/"
+  },
+  browserConfig: {
+    "baseURL": "/assets",
+    "paths": {
+      "media-atom-maker/": "beta/"
+    },
+    "bundles": {
```
clloyd: As you're (correctly) not including the `bundle.js` file in the repo then we also shouldn't check in this reference here

clloyd: You can run `jspm unbundle` to remove

clloyd:
Probably should hold off though until we have a reliable build process

PR #60, opened 2016-10-07
set allowfullscreen by its presence

akash1810:
the presence of allowfullscreen makes its value true, we don't need to assign it to empty string


akash1810:
closing as this was more of a vanity change and don't want to fix the failing test (`Attribute name "allowfullscreen" associated with an element type "iframe" must be followed by the ' = ' character.`)

PR #63, opened 2016-10-12
Atom list and single page

ShaunYearStrong:
Atom List page (/video/atoms):

picture 5

Single Atom page (/video/atoms/:id):

picture 6


public/video-ui/src/services/pandaReqwest.js
```diff
@@ -0,0 +1,28 @@
...
+
+export function pandaReqwest(reqwestBody) {
+  return Q.Promise(function(resolve, reject) {
+    Reqwest(reqwestBody)
+        .then(res => {
+          resolve(res)
+        })
+        .fail(err => {
+          if (err.status == 419) {
+            const reauthUrl = '/reauth';
```
clloyd: I think we need to implement this route?

ShaunYearStrong: Oops, right you are!

ShaunYearStrong: Fixed in commit 1fbb044

app/controllers/VideoUIApp.scala
```diff
@@ -2,11 +2,15 @@ package controllers
...
 import play.api.mvc._
 import javax.inject._
+
 import com.gu.pandomainauth.action.AuthActions
 
+
 class VideoUIApp @Inject() (val authActions: AuthActions)
   extends AtomController {
 
+  import authActions.{ AuthAction }
```
clloyd: Did IntelliJ insert that import here? Seems a really odd place

PR #65, opened 2016-10-13
Some simple UI improvements

ShaunYearStrong:
Adds some base grid-style styles to the list

picture 10


public/video-ui/src/components/Atoms/AtomItem.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import {Link} from 'react-router';
+
+export default class AtomItem extends React.Component {
```
clloyd: I'm not sure I'd use the name `Atom` here, it's a video we're representing.

public/video-ui/src/components/Atoms/AtomItem.js
```diff
@@ -0,0 +1,21 @@
...
+
+  render() {
+    return(
+        <li className="grid__item">
```
clloyd: With BEM wouldn't this be `atom__` or `video__`

public/video-ui/src/components/Header.js
```diff
@@ -8,16 +8,20 @@ export default class Header extends React.Component {
...
-            <span className="home-logo__text-large">Back to</span>
-            <span className="home-logo__text-small">Atoms</span>
-          </Link>
+        <header className="topbar">
 
-          <div className="header__children">
-            <nav className="links">
-              <IndexLink activeClassName="links__item--active" className="links__item top-toolbar__item--highlight" to="/video">Media Atoms</IndexLink>
+          <div className="topbar__container">
+            <Link to="/video/atoms" className="home-link" title="Home">
```
clloyd: `topbar__homelink`?

public/video-ui/src/components/Atoms/AtomItem.js
ShaunYearStrong: I just made it a bit more generic as it's kind of a grid view. But as we're only using it for videos/media atoms here, it makes sense to be a bit more specific

clloyd: 👍

public/video-ui/src/components/Atoms/AtomItem.js
ShaunYearStrong: Yep, agreed. I'll reformat this inside the redux PR as it'll cause a lot of conflicts/involve a lot of redundant work otherwise.

PR #66, opened 2016-10-17
Add redux

ShaunYearStrong:
Adds redux to the flow we already had.


public/video-ui/src/actions/VideoActions/getVideo.js
```diff
@@ -0,0 +1,37 @@
...
+  return {
+    type:       'VIDEO_GET_RECIEVE',
+    video:      video,
+    receivedAt: Date.now()
+  };
+}
+
+function errorRecievingVideo(error) {
+  return {
+    type:       'VIDEO_ERROR',
```
clloyd: This could just be the general `SHOW_ERROR` from below, saves having to implement a load of different error reducers

public/video-ui/src/util/storeAccessor.js
```diff
@@ -0,0 +1,9 @@
+let savedStore;
```
clloyd: Lets not have this storeAccessor until we need it, hopefully we never will!

ShaunYearStrong:
@clloyd Fixed those in commits 91bf6fc and d7e91e4

PR #67, opened 2016-10-18
Create and save videos

ShaunYearStrong:
Note: Saving will not work as the API doesn't accept JSON.


public/video-ui/src/actions/VideoActions/createVideo.js
```diff
@@ -0,0 +1,55 @@
...
+    receivedAt: Date.now()
+  };
+}
+
+function recieveVideoCreate(video, refreshSections) {
+  browserHistory.push('/video/videos/' + video.id);
+  return {
+    type: 'VIDEO_CREATE_RECEIVE',
+    video: video,
+    refreshSections: refreshSections,
```
clloyd: I don't think this is used?

public/video-ui/src/components/VideoEdit/VideoEdit.js
```diff
@@ -0,0 +1,27 @@
...
+    super(props);
+  }
+
+  render () {
+    if (!this.props.video) {
+      console.log('VideoEdit loaded without video provided');
+      return false;
+    }
+
+    if (this.props.video.data) {
```
clloyd: Could this be combined into the above if statement so no need to wrap main the actual return in an if statement

PR #68, opened 2016-10-19
Video form and validation

ShaunYearStrong:
First run at adding new fields and adding validation using [redux-form](http://redux-form.com/6.1.1/)

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -27,7 +27,6 @@ class VideoDisplay extends React.Component {
         <div className="container">
           <form className="form">
             <VideoEdit video={this.props.video || {}} updateVideo={this.updateVideo} />
-            {/*<SyncValidationForm />*/}
             <SaveButton onSaveClick={this.saveVideo} />
           </form>
         </div>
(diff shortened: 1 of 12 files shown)


public/video-ui/src/components/VideoEdit/formComponents/VideoCategory.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+import { videoCategories } from '../../../constants/videoCategories';
+
+export default class VideoCategorySelect extends React.Component {
+
+  constructor(props) {
```
clloyd: This can be removed.

public/video-ui/src/components/Video/Display.js
```diff
@@ -16,19 +20,44 @@ class VideoDisplay extends React.Component {
...
+    if(this.state.editable) {
+      return (
+          <div className="container">
+            <form className="form">
+              <VideoEdit video={this.props.video || {}} updateVideo={this.updateVideo}/>
+              <SaveButton onSaveClick={this.saveVideo}/>
+            </form>
+          </div>
+      )
+    } else {
```
clloyd: else not needed

PR #69, opened 2016-10-27
Instance type change

jennysivapalan:
As discussed moved instance type from m3.medium to t2.medium for Prod and  t2.small for Code.

While I was here I though it would be good to separate the min, max and desired capacity so that we can tweak them when we're ready for youtube etc.

cc @paulmr @clloyd

### DIFF
--- a/cloudformation/media-atom-maker.json
+++ b/cloudformation/media-atom-maker.json
@@ -350,7 +350,7 @@
                     "Ref": "MediaAtomLaunchConfig"
                 },
                 "MinSize": { "Fn::FindInMap": [ "StageMap", {"Ref": "Stage"}, "MinSize"]},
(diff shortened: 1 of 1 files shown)


cloudformation/media-atom-maker.json
```diff
@@ -53,6 +53,23 @@
             "Type": "String"
         }
     },
+    "Mappings": {
+        "StageMap": {
+            "PROD": {
+                "MinSize": 1,
```
clloyd: Should this be 3, 6 for both?

jennysivapalan: There is currently just 1 in production so I kept it the same. I assume we'll be bumping it up in the near future but perhaps we're not at that point?

jennysivapalan:
Applied on CODE fine.

PR #70, opened 2016-10-27
Add assets to details page and split into components

ShaunYearStrong:
picture 18

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -29,6 +29,21 @@ class VideoDisplay extends React.Component {
     });
   };
 
+  renderDetails = () => {
+    if(this.state.editable) {
+      return (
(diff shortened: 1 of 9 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -41,20 +43,27 @@ class VideoDisplay extends React.Component {
 
     if(this.state.editable) {
       return (
-          <div className="container">
-            <form className="form">
-              <VideoEdit video={this.props.video || {}} updateVideo={this.updateVideo}/>
-              <SaveButton onSaveClick={this.saveVideo}/>
-            </form>
+          <div className="video">
```
clloyd: There's quite alot of duplication here between the two different renders... could there be a seperate video sidebar render method that does the editable check... or separate component that accepts a editable prop

PR #72, opened 2016-10-28
Parse metadata fields

marialivia16:
This PR just adds the support for parsing metadata, updates some versions and adds a test.

As soon as this is merged we can implement the API for metadata.


clloyd:
:+1:

PR #71, opened 2016-10-28
Fix submission and clean up styles

ShaunYearStrong:
SCREENSHOTS AHOY!

picture 19
picture 20
picture 21
picture 22

PR #73, opened 2016-10-31
Add video search

susiecoleman:
videofilter

Adds a search bar to allow filtering by title. Currently it looks ugly as all the beautiful style is in @ShaunYearStrong's pr https://github.com/guardian/media-atom-maker/pull/71 but the correct classes are being used for the style to be consistent after the merge.

### DIFF
--- a/public/video-ui/src/components/VideoSearch/VideoSearch.js
+++ b/public/video-ui/src/components/VideoSearch/VideoSearch.js
@@ -2,10 +2,6 @@ import React from 'react'
 
 export default class VideoSearch extends React.Component {
 
-  constructor(props) {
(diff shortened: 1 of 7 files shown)


public/video-ui/src/components/VideoSearch/VideoSearch.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+export default class VideoSearch extends React.Component {
+
+  constructor(props) {
```
ShaunYearStrong: I don't think this is needed, I've taken out a few in my previous PR but there are still some hanging about.

PR #74, opened 2016-10-31
Add config file

ShaunYearStrong:


public/video-ui/src/reducers/rootReducer.js
```diff
@@ -1,11 +1,13 @@
 import { combineReducers } from 'redux';
+import config from './configReducer';
 import error from './errorReducer';
 import video from './videoReducer';
 import videos from './videosReducer';
 import { reducer as form } from 'redux-form'
 
 
 export default combineReducers({
```
susiecoleman: Just for my own curiosity is it coincidence that these are in alphabetical order or is this good style for reducer combining?

public/video-ui/src/reducers/configReducer.js
```diff
@@ -0,0 +1,11 @@
+export default function config(state = false, action) {
+  switch (action.type) {
+
+    case 'CONFIG_RECEIVED':
+        return action.config || false;
```
susiecoleman: I was just wondering why the default is false when config is an object?

public/video-ui/src/services/pandaReqwest.js
```diff
@@ -10,7 +11,8 @@ export function pandaReqwest(reqwestBody) {
         })
         .fail(err => {
           if (err.status == 419) {
-            const reauthUrl = '/reauth';
+            const store = getStore();
+            var reauthUrl = store.getState().config.reauthUrl;
```
susiecoleman: What happens if config has been set to false in the reducer or the config isn't there how does it fail?

public/video-ui/src/reducers/rootReducer.js
ShaunYearStrong: Kind of coincidence, I unconsciously did it as that's the order they're in the folder in.

public/video-ui/src/reducers/configReducer.js
ShaunYearStrong: This is just how the video was done, I agree though, we should return an empty object, then at least it's be undefined and not an error

public/video-ui/src/services/pandaReqwest.js
ShaunYearStrong: In the app.js file, config is set to either the returned config or an empty object, so this would account for that.

public/video-ui/src/reducers/configReducer.js
ShaunYearStrong: Actually, I'm taking that back, the config is set to an empty object when it's added to the store if there isn't one. We could just remove the check

PR #78, opened 2016-11-03
Bug Fix: Add more appropriate default data for creating a video.

susiecoleman:
Problem: The create video page was not rendering

Cause: The page would first try to render with `this.props.video = {}` as the blank video default data was only assigned to this.props.video after the component mounts. This caused issues as there were references to `this.props.video.data.title` in sub components of the create component.

Fix: The default is now 
``` 
this.props.video = {      
    data: {
           title: '',
           category: ''
    }
}
```

### DIFF
--- a/public/video-ui/src/components/Video/Create.js
+++ b/public/video-ui/src/components/Video/Create.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import VideoEdit from '../VideoEdit/VideoEdit';
 import SaveButton from '../utils/SaveButton';
+import {blankVideoData} from '../../constants/blankVideoData';
 
 class VideoCreate extends React.Component {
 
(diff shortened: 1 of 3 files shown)


public/video-ui/src/components/Video/Create.js
```diff
@@ -21,10 +21,17 @@ class VideoCreate extends React.Component {
   };
 
   render () {
+    const defaultVideoData = {
```
clloyd: This should probably be defined as a constants in the constants directory

ShaunYearStrong: I agree, this could then also be used in the `BLANK_VIDEO` for consistency

susiecoleman: Moved thanks

PR #79, opened 2016-11-03
Allow reverting (changing) between assets

ShaunYearStrong:
picture 29

### DIFF
--- a/public/video-ui/src/components/VideoAssets/VideoAssetItem.js
+++ b/public/video-ui/src/components/VideoAssets/VideoAssetItem.js
@@ -9,12 +9,29 @@ export default class VideoAssetItem extends React.Component {
     return store.getState().config.youtubeEmbedUrl;
   };
 
+  revertAsset = () => {
+    this.props.revertAsset(this.props.asset.version);
+  };
+
+  renderAssetVersionButton = () => {
+    if(this.props.activeAsset !== this.props.asset.version) {
+      return (
+        <button type="button" className="btn" onClick={this.revertAsset}>
+          Set as current asset
+        </button>
+      )
+    } else {
+      return false;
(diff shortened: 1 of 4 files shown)


public/video-ui/src/services/VideosApi.js
```diff
@@ -40,6 +40,13 @@ export default {
     })
   },
 
+  revertAsset: (version, videoId) => {
+    return pandaReqwest({
+      url: '/api/atom/' + videoId + '/revert/' + version,
+      method: 'post'
```
akash1810: a `PUT` would suit this endpoint more imo, but not related to this PR

public/video-ui/src/components/VideoAssets/VideoAssetItem.js
```diff
@@ -9,12 +9,29 @@ export default class VideoAssetItem extends React.Component {
...
+
+  renderAssetVersionButton = () => {
+    if(this.props.activeAsset !== this.props.asset.version) {
+      return (
+        <button type="button" className="btn" onClick={this.revertAsset}>
+          Set as current asset
+        </button>
+      )
+    } else {
+      return false;
```
akash1810: don't know much react - would `false` get rendered below?

public/video-ui/src/services/VideosApi.js
ShaunYearStrong: Yep, agreed!

public/video-ui/src/components/VideoAssets/VideoAssetItem.js
ShaunYearStrong: Nah, if the render function doesn't return anything, then nothing happens

PR #88, opened 2016-11-11
update video api service

akash1810:
The api was recently updated, changing POSTs to PUTs. These changes were missed.


akash1810:
cc @ShaunYearStrong

PR #91, opened 2016-11-15
edit yt category client side

akash1810:
not linked to datastore yet as api needs updating, but this will be done in a different PR

yt-cat

### DIFF
--- a/public/video-ui/src/actions/YoutubeActions/getCategories.js
+++ b/public/video-ui/src/actions/YoutubeActions/getCategories.js
@@ -1,0 +1,24 @@
+import YoutubeApi from '../../services/YoutubeApi';
+
+function requestCategories() {
+  return {
+    type: 'YT_CATEGORIES_GET_REQUEST',
+    receivedAt: Date.now()
+  };
+}
+
+function receiveCategories(categories) {
+  return {
+    type: 'YT_CATEGORIES_GET_RECEIVE',
+    categories: categories,
+    receivedAt: Date.now()
+  };
+}
+
+export function getCategories() {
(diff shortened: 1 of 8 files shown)


public/video-ui/src/actions/YoutubeActions/getCategories.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+function receiveCategories(categories) {
+  return {
+    type: 'YT_CATEGORIES_GET_RECEIVE',
+    categories: categories,
+    receivedAt: Date.now()
+  };
+}
+
```
ShaunYearStrong: We should add something for errors here, as in the other actions.

public/video-ui/src/components/VideoEdit/formComponents/YoutubeCategory.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+
+class YoutubeCategorySelect extends React.Component {
+
+  hasCategories () {
```
ShaunYearStrong: doesn't this need arrow function syntax to bind to `this` correctly?

public/video-ui/src/actions/YoutubeActions/getCategories.js
akash1810: Done.

public/video-ui/src/components/VideoEdit/formComponents/YoutubeCategory.js
akash1810: Fixed now.

PR #93, opened 2016-11-17
Update Youtube metadata endpoint

marialivia16:
This endpoint is found under `/api/atom/:id/updateMetadata` and it updates the following metadata fields: `description`, `categoryId`, `tags` and `license`.

NOTE: At the moment the `commentsEnabled` flag is not updated because YT doesn't provide proper support for this action. The closest thing I found is something called [`isPublic`](https://developers.google.com/youtube/v3/docs/commentThreads#snippet.isPublic) - if we decide to go with that, I'll add it in a different PR.


app/util/Youtube.scala
```diff
@@ -47,3 +56,24 @@ case class YouTubeVideoCategoryApi (config: YouTubeConfig) {
...
 }
+
+case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends YouTubeBuilder {
+  def updateMetadata(id: String, metadata: JsValue): Option[Video] =
+      youtube.videos().list("snippet, status").setId(id).execute().getItems.asScala.toList.headOption match {
+      case Some(video) =>
+        val snippet = video.getSnippet
+        snippet.setTitle(snippet.getTitle)
+        snippet.setCategoryId((metadata \ "categoryId").asOpt[String].fold(snippet.getCategoryId)(identity))
+        snippet.setChannelId((metadata \ "channelId").asOpt[String].fold(snippet.getChannelId)(identity))
```
akash1810: we can't change the channel id of a video, so we can remove this line

conf/routes
```diff
@@ -17,6 +17,7 @@ GET     /api2/atom/:id                  controllers.Api2.getAtom(id)
 PUT     /api2/atom/:id/asset            controllers.Api2.addAsset(id)
 
 GET     /api/youtube/categories         controllers.YoutubeCategory.list()
+PUT     /api/youtube/videos/:id/update  controllers.YoutubeVideos.update(id)
```
akash1810: this endpoint may be a little confusing as `id` is an atom id, rather than a youtube video id

marialivia16: Do you think it's better to keep this `id` as the youtube video id and add the logic we discussed one level up?

app/util/Youtube.scala
```diff
@@ -47,3 +56,24 @@ case class YouTubeVideoCategoryApi (config: YouTubeConfig) {
...
   }
 }
+
+case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends YouTubeBuilder {
+  def updateMetadata(id: String, metadata: JsValue): Option[Video] =
+      youtube.videos().list("snippet, status").setId(id).execute().getItems.asScala.toList.headOption match {
+      case Some(video) =>
+        val snippet = video.getSnippet
+        snippet.setTitle(snippet.getTitle)
+        snippet.setCategoryId((metadata \ "categoryId").asOpt[String].fold(snippet.getCategoryId)(identity))
```
akash1810: may want to deserialise `metadata` to a more friendly type than `JsValue`

conf/routes
marialivia16: It is confusing to use the `atomId` in this case as this endpoint is doing operations on youtube videos. But as you said, changing it to the youtube video id would mean updating any video.

akash1810: maybe `/api/atom/:id/updateMetadata`? We'd still want to be able to update metadata for non-youtube atoms.

marialivia16: Then I'll move this in the `Api2.scala` controller, because this is about updating the metadata for all platforms, it makes no sense to sit in a file called `YoutubeVideo.scala`.

app/controllers/Api2.scala
```diff
@@ -5,10 +5,12 @@ import javax.inject.Inject
...
 import com.gu.atom.play.AtomAPIActions
 import com.gu.atom.publish.{PreviewAtomPublisher, LiveAtomPublisher}
+import com.gu.contentatom.thrift.atom.media.Platform
 import com.gu.pandahmac.HMACAuthActions
-import model.commands.AddAssetCommand
+import model.UpdatedMetadata
+import model.commands.{UpdateMetadataCommand, AddAssetCommand}
 import play.api.Configuration
-import util.AWSConfig
+import _root_.util.{YouTubeConfig, YouTubeVideoUpdateApi, AWSConfig}
```
akash1810: why `_root_`?

app/model/commands/UpdateMetadataCommand.scala
```diff
@@ -0,0 +1,65 @@
...
+        activeAsset match {
+          case Some(asset) =>
+
+            if(asset.platform == Platform.Youtube)
+              YouTubeVideoUpdateApi(youtubeConfig).updateMetadata(asset.id, metadata)
+
+            val newMetadata = mediaAtom.metadata.map(_.copy(
+              tags = metadata.tags,
+              categoryId = metadata.categoryId,
+              license = metadata.license))
```
akash1810: the `if(asset.platform == Platform.Youtube)` check doesn't appear to be scoped to here, does this mean we're setting `license` on non youtube atoms? (though I may be reading GH incorrectly!)

app/controllers/Api2.scala
marialivia16: Oh, that's intelliJ being silly.

app/model/commands/UpdateMetadataCommand.scala
marialivia16: You're right, I was confused - as far as I know the non-youtube assets do not have metadata. I'll correct this.

PR #94, opened 2016-11-17
Moving to a new API and Atom Data Structure

currysoup:


kelvin-chappell:
This looks like it will be much easier to work with.

PR #95, opened 2016-11-17
Add CAPI usages check

clloyd:
screen shot 2016-11-17 at 16 23 48

This adds an (unstyled) check for video usage in other content. There is a new `VideoUsages` component which makes use of a CAPI proxy to access preview CAPI (`/support/capiPreview/`). This requires a key and username for capi preview in configuration

### DIFF
--- a/public/video-ui/src/components/VideoUsages/VideoUsages.js
+++ b/public/video-ui/src/components/VideoUsages/VideoUsages.js
@@ -50,12 +50,6 @@ class VideoUsages extends React.Component {
 
   renderUsages() {
 
-    const fakeUsages = [
-      'technology/2016/oct/21/unlaunched-test-article',
-      'technology/2016/oct/14/testing-witness-embed',
-      'technology/2013/sep/18/ios-7-review-can-apple-hold-off-the-attack-of-the-giant-screens',
-    ];
-
     if (!this.state.usages) {
       return (<div>Fetching Usages...</div>)
     }
@@ -66,7 +60,7 @@ class VideoUsages extends React.Component {
 
     return (
       <ul className="usages__list">
-        {fakeUsages.map(this.renderUsage)}
+        {this.state.usages.map(this.renderUsage)}
       </ul>
     )
   }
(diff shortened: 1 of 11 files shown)


public/video-ui/src/components/VideoUsages/VideoUsages.js
```diff
@@ -0,0 +1,91 @@
...
+
+  state = {
+    usages: undefined
+  }
+
+  updateUsages(videoId) {
+
+    if (!videoId) {
+      return this.setState({
+        usages: undefined
```
akash1810: I did `[]` for the yt categories, shall we stick to one pattern? (happy to update the categories)

public/video-ui/src/components/VideoUsages/VideoUsages.js
```diff
@@ -0,0 +1,91 @@
...
+    return (
+      <li className="usages__list__item">
+        {usage}
+      </li>
+    )
+  }
+
+  renderUsages() {
+
+    const fakeUsages = [
```
akash1810: so we're not talking to capi?

public/video-ui/src/components/VideoUsages/VideoUsages.js
clloyd: It depends on the situation, undefined in this case means we've never fetched a state, [] means we've fetched the state and there's no usages.

public/video-ui/src/components/VideoUsages/VideoUsages.js
clloyd: d'oh! 🍩 Will fix

clloyd:
@akash1810 can you re-review?

public/video-ui/src/services/capi.js
```diff
@@ -0,0 +1,11 @@
+import {pandaReqwest} from './pandaReqwest';
+import {getStore} from '../util/storeAccessor';
+
+export function fetchUsages(videoId) {
+  
+  const capiProxyUrl = getStore().getState().config.capiProxyUrl;
+  return pandaReqwest({
+    url: capiProxyUrl + "/atom/media/" + videoId + "/usage",
```
akash1810: minor: could use template strings here

PR #97, opened 2016-11-18
pass `--debug` flag to helper scripts to launch sbt in debug mode

akash1810:
this means you can run

`./scripts/client-dev.sh --debug` and attach a remove jvm inspector to interactively debug scala in your ide of choice.


currysoup:
:+1:

PR #98, opened 2016-11-21
Add embeddedMode config switch and simple postMessage button

clloyd:
cc @akash1810 

This looks for an `embeddedMode=true` parameter in the url and sets a config bool to true.


akash1810:
👍

PR #104, opened 2016-11-23
Active sset endpoint

Reettaphant:
Add an endpoint to media atom maker that makes an asset active if youtube has finished encoding it. The endpoint is polled by CDS.

@clloyd @akash1810

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -11,12 +11,11 @@ import com.gu.pandahmac.HMACAuthActions
 import data.JsonConversions._
 import model.commands.CommandExceptions._
 import model.commands._
-import model.UpdatedMetadata
 import play.api.Configuration
 import util.{YouTubeConfig, AWSConfig}
 import util.atom.MediaAtomImplicits
 import play.api.libs.json._
-import model.MediaAtom
+import model.{UpdatedMetadata, MediaAtom}
 
 import scala.util.{Failure, Success}
 
@@ -112,5 +111,21 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
       BadRequest("Could not read json")
     }
     Ok
+  }
+
+  def setActiveAsset(atomId: String) = APIHMACAuthAction { implicit req =>
+    req.body.asJson.map { json =>
+      try {
+        val videoId = (json \ "youtubeId").as[String]
+
+        val status: Unit = ActiveAssetCommand(atomId, videoId).process()
+
+        Ok("asset added")
+      } catch {
+        commandExceptionAsResult
+      }
+    }.getOrElse {
+      BadRequest("Could not read json")
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


app/util/Youtube.scala
```diff
@@ -98,3 +98,17 @@ case class YouTubeChannelsApi(config: YouTubeConfig) extends YouTubeBuilder {
 
   }
 }
+
+case class YouTubeVideoStatusApi(config: YouTubeConfig) extends YouTubeBuilder {
+  def get(youtubeId: String): Option[String] =
+    youtube.videos()
+      .list("processingDetails")
+      .setId(youtubeId)
```
akash1810: what happens when the youtubeId isn't in our account?

app/controllers/Api2.scala
```diff
@@ -113,4 +112,20 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
...
   }
+
+  def setActiveAsset(atomId: String) = APIHMACAuthAction { implicit req =>
+    req.body.asJson.map { json =>
+      try {
+        val videoId = (json \ "youtubeId").as[String]
+
+        val status: Unit = ActiveAssetCommand(atomId, videoId).process()
+
+        Ok("made asset " + videoId + " active in atom " + atomId)
```
currysoup: There are a few places where errors just return raw strings. If we're going to move to JSON it would be sensible to do it everywhere.

app/util/Youtube.scala
Reettaphant: There is an unauthorised error from the youtube api.

akash1810: Cool. Do we handle this scenario gracefully?

Reettaphant: Yes, I've done that now.

app/model/commands/ActiveAssetCommand.scala
```diff
@@ -0,0 +1,85 @@
...
+
+            previewDataStore.getAtom(atomId) match {
+              case Some(atom) => {
+                val mediaAtom = atom.tdata
+                val atomAssets: Seq[Asset] = mediaAtom.assets
+
+                atomAssets.find(asset => asset.id == youtubeId) match {
+                  case Some(newActiveAsset) => {
+
+                    val newAtom = atom
```
akash1810: minor: would `newAtomRevision` be more descriptive?

app/model/commands/CommandException.scala
```diff
@@ -12,6 +12,7 @@ object CommandExceptions extends Results {
   def NotYoutubeAsset = throw new CommandException("Asset is not a youtube video", 400)
   def AssetVersionConflict = throw new CommandException("Asset version conflict", 400)
   def AssetParseFailed = throw new CommandException("Failed to parse asset", 400)
+  def AssetEncodingInProcess = throw new CommandException("Asset encoding in process", 400)
```
akash1810: do you mean `InProgress`?

PR #112, opened 2016-11-25
Metadata/Thumbnail sync on publish

currysoup:


currysoup:
After far too long trying to figure out why the little thumbnail wasn't displaying in MAM i found out it's because unlisted videos don't have publicly visible thumbnails.

public/video-ui/src/services/VideosApi.js
```diff
@@ -29,7 +29,7 @@ export default {
 
   publishVideo: (videoId) => {
     return pandaReqwest({
-      url: '/api/atom/' + videoId + '/publish',
+      url: '/api2/atom/' + videoId + '/publish',
```
akash1810: should this be `atomId` rather than `videoId`?

public/video-ui/src/components/VideoAssetAdd/formComponents/VideoAssetUrl.js
```diff
@@ -6,7 +6,7 @@ import React from 'react';
 export default class VideoAssetUrl extends React.Component {
 
   onUpdateAssetUrl = (e) => {
-    const newVersion = this.props.video.activeVersion === undefined ? 0 : this.props.video.activeVersion + 1;
+    const newVersion = this.props.video.activeVersion === undefined ? 0 : Math.max(...this.props.video.assets.map(x => x.version)) + 1;
```
akash1810: why are we doing this?

currysoup: Say you have asset versions 1, 2 and 3 and you roll back the `activeVersion` to 2, then the new version generated would be 3. This causes a version conflict.

app/controllers/Api2.scala
```diff
@@ -10,13 +10,13 @@ import data.JsonConversions._
 import model.commands.CommandExceptions._
 import model.commands._
 import play.api.Configuration
-import util.{YouTubeConfig, AWSConfig}
+import _root_.util.{ YouTubeConfig, AWSConfig}
```
currysoup: 😞  "Helpful", indeed :)

public/video-ui/src/components/VideoAssetAdd/formComponents/VideoAssetUrl.js
akash1810: should this be fixed server side then?

currysoup: The pattern we've been following is the client `PUT`s a well formed asset/video/whatever and the server either accepts or rejects it.

We could move to a model where we send the data required to generate the mentioned asset/video/whatever but that would be a bit of a break from convention.

public/video-ui/src/services/VideosApi.js
currysoup: Will do in another PR.

PR #114, opened 2016-11-29
rename `name` to `title` to work with SelectBox component

akash1810:
similar to https://github.com/guardian/media-atom-maker/pull/113

Also fixed a 500 that was occurring as the API was expecting the payload to be capitalised.


currysoup:
:+1:

PR #124, opened 2016-12-05
Improve Video Grid UI

clloyd:
screen shot 2016-12-05 at 11 59 40

🐧  🐧  🐧  🐧

### DIFF
--- a/public/video-ui/src/components/Videos/VideoItem.js
+++ b/public/video-ui/src/components/Videos/VideoItem.js
@@ -3,6 +3,28 @@ import {Link} from 'react-router';
 
 export default class VideoItem extends React.Component {
 
+  renderActiveAssetName() {
+
+    if (this.props.video.activeVersion && this.props.video.assets.length) {
+      const activeAssets = this.props.video.assets.filter((asset) => asset.version === this.props.video.activeVersion)
+      if (activeAssets.length !== -1) {
+        if (activeAssets[0] && activeAssets[0].platform) {
+          return <span className="success">Active {activeAssets[0].platform} video</span>
+        }
+      }
+    }
+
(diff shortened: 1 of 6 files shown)


clloyd:
@akash1810

public/video-ui/src/components/Videos/VideoItem.js
```diff
@@ -3,19 +3,41 @@ import {Link} from 'react-router';
 
 export default class VideoItem extends React.Component {
 
+  renderActiveAssetName() {
+
+    if (this.props.video.activeVersion && this.props.video.assets.length) {
+      const activeAssets = this.props.video.assets.filter((asset) => asset.version === this.props.video.activeVersion)
+      if (activeAssets.length !== -1) {
```
akash1810: does this work? wouldn't `.filter` return an empty array if nothing can be found?

public/video-ui/styles/layout/_grid.scss
```diff
@@ -11,38 +10,96 @@
...
+  text-align: center;
 }
 
 .grid__item {
-  flex-basis: 25%;
-  padding: 5px;
+  display: inline-block;
+
+  position: relative;
+  border: 3px solid #3e3e3e;
```
akash1810: can we use a variable here?

public/video-ui/src/components/Videos/VideoItem.js
clloyd: Whoops wrong syntax should be `!activeAssets.length`

clloyd:
@akash1810 fixed your comments, but accidentally pushed a fix for select boxes showing id's rather than titles when in display mode

PR #129, opened 2016-12-05
Use new API for updating asset version

marialivia16:
Uses the new `api2` endpoint for changing the active version.


Reettaphant:
Should we display an error message to the user if the user tries to make a video that is still being processed by youtube active?

marialivia16:
@Reettaphant We are already showing the `Could not revert the asset` error when something goes wrong.

marialivia16:
@Reettaphant Also, future changes will make the error message more specific. In the meanwhile, whenever you have time we could check together that CDS still works fine with these changes before merging this. Thanks!

conf/routes
```diff
@@ -21,7 +21,7 @@ POST    /api2/atoms/:id/assets          controllers.Api2.addAsset(id)
 
 PUT     /api2/atom/:id/publish          controllers.Api2.publishMediaAtom(id)
 PUT     /api/atom/:id/update-metadata   controllers.Api2.updateMetadata(id)
-POST    /api2/atom/:id/asset-active     controllers.Api2.setActiveAsset(id)
+PUT     /api2/atom/:id/asset-active     controllers.Api2.setActiveAsset(id)
```
akash1810: This'll require a change to CDS too - https://github.com/guardian/content_delivery_system/blob/c96aa063527edef19badbc060b317c20705dadc3/CDS/scripts/js_utils/media-atom-lib.js#L127

We'd need to be careful on the deploy order.

app/controllers/Api2.scala
```diff
@@ -114,17 +114,14 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
...
   def setActiveAsset(atomId: String) = APIHMACAuthAction { implicit req =>
     req.body.asJson.map { json =>
       try {
         val videoId = (json \ "youtubeId").as[String]
-
-        val status: Unit = ActiveAssetCommand(atomId, videoId).process()
-
-        Ok("made asset " + videoId + " active in atom " + atomId)
+        val atom = ActiveAssetCommand(atomId, videoId).process()
+        Ok(Json.toJson(atom))
```
akash1810: 😍

conf/routes
marialivia16: I've created a PR with the CDS change: https://github.com/guardian/content_delivery_system/pull/50

As for the deploy order, would it be alright to deploy this first and CDS right after?

akash1810: Yep, though you may want to become familiar with the CDS deploy process beforehand as its not riff-raff - https://sites.google.com/a/guardian.co.uk/gnm-multimedia-knowledgebase/home/how-to-deploy-a-cds-code-update

Once this change goes out, CDS will break as until deployed.

PR #133, opened 2016-12-06
Send correct change details to client side

clloyd:
This PR does the following
 - Updates PublishCommand to set the correct change records and update preview before updating live
 - Modifies UpdateCommand and PublishCommand to return the updated MediaAtom after completing
 - Listens client side for the response from save and publish requests and pulls the appropriate contentChangeRecord out of the response for use client side @ShaunYearStrong

### DIFF
--- a/app/model/commands/UpdateAtomCommand.scala
+++ b/app/model/commands/UpdateAtomCommand.scala
@@ -7,20 +7,20 @@ import com.gu.atom.publish.PreviewAtomPublisher
 import com.gu.contentatom.thrift.{Atom, ContentAtomEvent, EventType}
 import model.commands.CommandExceptions._
 import util.atom.MediaAtomImplicits
-import model.MediaAtom
+import model.{ChangeRecord, MediaAtom}
 
 import scala.util.{Failure, Success}
-
 import model.commands.CommandExceptions._
+import org.joda.time.DateTime
 case class UpdateAtomCommand(id: String, atom: MediaAtom)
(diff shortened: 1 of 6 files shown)


app/model/commands/UpdateAtomCommand.scala
```diff
@@ -31,7 +31,12 @@ case class UpdateAtomCommand(id: String, atom: MediaAtom)
       AtomNotFound
     }
 
-    val details = atom.contentChangeDetails.copy(revision = existingAtom.get.contentChangeDetails.revision + 1)
+    val changeRecord = ChangeRecord(DateTime.now(), None)
```
clloyd: There's two ChangeRecord models, one wants `Long` one wants `DateTime`

clloyd: Completely agree with the abstracting, can do this when we add Users in (I have a list of Todos!)

public/video-ui/src/components/VideoPreview/VideoPreview.js
```diff
@@ -36,7 +36,7 @@ export default class VideoPreview extends React.Component {
     }
 
     return  (
-      <iframe className="video-preview__video" src={this.youtubeEmbedUrl() + activeAsset.id} allowfullscreen></iframe>
+      <iframe className="video-preview__video" src={this.youtubeEmbedUrl() + activeAsset.id} allowFullScreen></iframe>
```
clloyd: React wants a specific capitalisation

PR #136, opened 2016-12-06
Move asset list and hide old assets + some picky spacing changes.

ShaunYearStrong:
picture 48
picture 49


public/video-ui/src/components/VideoAssets/VideoAssets.js
```diff
@@ -54,12 +62,40 @@ class VideoAssets extends React.Component {
...
+    if (this.state.showAssetList) {
+      return (
+        this.props.video.assets.map(this.mapListItems)
+      )
+    }
+
+    return false;
   }
 
+  mapListItems = (asset, index) => {
```
clloyd: Is this a filter rather than a map?

PR #138, opened 2016-12-06
set duration server side

akash1810:
a PUT to `update-metadata` or a POST to `asset-active` will get the active asset for an atom and if youtube, will make a call to the yt api and save it to the atom duration.

A refresh is still needed client-side, but that'll be a different PR.


clloyd:
I say instead of refreshing client side... we just remove all trace of the field

PR #140, opened 2016-12-07
Video Display/Assets/Usages tweaks

ShaunYearStrong:
picture 50


public/video-ui/styles/layout/_video.scss
```diff
@@ -85,5 +86,9 @@
     text-transform: uppercase;
     padding: 10px;
     font-size: 17px;
+
+    &--with-btn {
```
akash1810: no rules == not needed?

PR #144, opened 2016-12-13
Restrict youtube assets to ones in the Guardian's ownership.

akash1810:
Adds a restriction allowing only youtube videos that we own to be added as an asset as only our content should be atoms.

not-gu

### DIFF
--- a/app/controllers/Youtube.scala
+++ b/app/controllers/Youtube.scala
@@ -4,7 +4,7 @@ import javax.inject._
 
 import com.gu.pandahmac.HMACAuthActions
 import play.api.libs.json.Json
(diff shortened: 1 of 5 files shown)


kelvin-chappell:
Doesn't this mean hosted video atoms won't work?
eg. https://www.theguardian.com/advertiser-content/explore-canada-food-busker-in-canada/duelling-bagels

akash1810:
@kelvin-chappell shoot yeah! Will update to only do this validation when `Category != Hosted`

akash1810:
[Updated](https://github.com/guardian/media-atom-maker/pull/144/commits/9661d08a1e13b270bd0e28e1c29a82e2dee3f22e) now @kelvin-chappell

kelvin-chappell:
👍  Thanks!

PR #145, opened 2016-12-13
Publish on asset active, reuse other commands for update/publish

clloyd:
This aims to remove the updateAtom logic found in most routes with a call to the main UpdateAtomCommand

This also adds publish when an asset is marked as active.

cc @akash1810 @currysoup


app/model/commands/ActiveAssetCommand.scala
```diff
@@ -32,57 +33,52 @@ case class ActiveAssetCommand(atomId: String, youtubeId: String)
...
+        videoStatus match {
+          case Some("succeeded") => markAssetAsActive()
           case Some(_) => AssetEncodingInProcess
           case None => NotYoutubeAsset
         }
       }
-      case e: YoutubeException => BadRequest(e.toString())
+      case e: YoutubeException => {
+        Logger.error(e.toString)
+        UnknownFailure
```
akash1810: do we need to be more specific here?

clloyd: Added much clearer commandexception

PR #146, opened 2016-12-13
make yt channel readonly in edit mode

akash1810:
As we cannot move a video between yt channels, it doesn't make sense to be able to edit it.
If we want this feature, we would need to re-upload the video to the new channel and set the old video private.


akash1810:
fixes https://github.com/guardian/media-atom-maker/projects/1#card-1117460

akash1810:
fixes https://github.com/guardian/media-atom-maker/issues/154

PR #147, opened 2016-12-14
Fix React errors caused by empty values in validation

ShaunYearStrong:


public/video-ui/src/constants/videoEditValidation.js
```diff
@@ -1,16 +1,18 @@
...
-  }
-  if (!values.category) {
-    errors.category = 'Required'
-  }
-  if (!values.posterImage) {
-    errors.posterImage = 'Required'
-  }
-  if (!values.duration) {
-    errors.duration = 'Required'
+  if (values) {
```
clloyd: Would this be nicer just to have a single `if (!values){return false}` at the top?

public/video-ui/src/actions/YoutubeActions/getCategories.js
```diff
@@ -16,6 +16,7 @@ function receiveCategories(categories) {
 }
 
 function errorReceivingCategories(error) {
+  console.error(error);
```
akash1810: is this needed?

clloyd: Yeah

public/video-ui/src/constants/videoEditValidation.js
ShaunYearStrong: Yeah actually, stops the nested `if`

PR #148, opened 2016-12-14
protect against mistakes in DEV

akash1810:
We currently have the power to edit any metadata of any video in our account in DEV. This is scary!

This PR adds some protection so that we don't, for example, change the title of a video of a public video to something like "test". Instead, we can only edit the metadata of video in a channel in config.youtube.allowedChannels.

### DIFF
--- a/app/util/Youtube.scala
+++ b/app/util/Youtube.scala
@@ -13,7 +13,7 @@ import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.youtube.YouTube
 import com.google.api.services.youtube.model.Video
 import model.{UpdatedMetadata, YouTubeChannel, YouTubeVideoCategory}
-import play.api.Configuration
+import play.api.{Configuration, Logger}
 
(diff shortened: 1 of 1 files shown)


app/util/Youtube.scala
```diff
@@ -86,13 +99,22 @@ case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends YouTubeBuilder {
       case _ => None
     }
 
-  def updateThumbnail(id: String, thumbnailUrl: URL, mimeType: String): Unit = {
-    val content = new InputStreamContent(mimeType, new BufferedInputStream(thumbnailUrl.openStream()))
-    val set = youtube.thumbnails().set(id, content).setOnBehalfOfContentOwner(config.contentOwner)
+  def updateThumbnail(id: String, thumbnailUrl: URL, mimeType: String): Option[Video] = {
+    YouTubeVideoInfoApi(config).getVideo(id, "snippet") match {
```
akash1810: we're now making an extra API call, but I think its worth it

PR #150, opened 2016-12-14
reverse sorted list to get the biggest image

akash1810:
`sortBy` in descending order to get the largest crop


currysoup:
Good catch @akash1810 , just to point out rather than `.reverse.head` we could simply do `.last`

akash1810:
`.last` is way more elegant! https://github.com/guardian/media-atom-maker/pull/152

PR #153, opened 2016-12-15
Minor update asset changes

marialivia16:
Some changes that got lost after the rebase against master 😢

### DIFF
--- a/app/model/commands/ActiveAssetCommand.scala
+++ b/app/model/commands/ActiveAssetCommand.scala
@@ -38,12 +38,12 @@ case class ActiveAssetCommand(atomId: String, youtubeId: String)
 
   def markAssetAsActive(): MediaAtom = {
     previewDataStore.getAtom(atomId) match {
-      case Some(atom) => {
+      case Some(atom) =>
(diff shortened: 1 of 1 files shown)


app/model/commands/ActiveAssetCommand.scala
```diff
@@ -69,15 +67,21 @@ case class ActiveAssetCommand(atomId: String, youtubeId: String)
...
-          case Some("succeeded") => markAssetAsActive()
 
+        /** Processing status:
+          * failed – Video processing has failed.
+          * processing – Video is currently being processed.
+          * succeeded – Video has been successfully processed.
+          * terminated – Processing information is no longer available.
+          **/
+        videoStatus match {
+          case Some(status) if status == "succeeded" || status == "terminated" => markAssetAsActive()
```
akash1810: is `terminated` a good thing? Does it represent a playable video?

marialivia16: As far as I can tell, yes. There are a lot of videos which have the status `terminated` but you can play them.

PR #156, opened 2016-12-15
[wip] use poster image as default html for new atom

akash1810:
Composer renders the default html read from CAPI. If there is no asset on the atom, this is just `<div></div>` which looks a little odd.

This changes the default html to be an image tag.

### DIFF
--- a/app/model/commands/CreateAtomCommand.scala
+++ b/app/model/commands/CreateAtomCommand.scala
@@ -37,11 +37,16 @@ case class CreateAtomCommand(data: CreateAtomCommandData)
 
   def process() = {
 
+    val defaultHtml = data.posterImage.master match {
+      case Some(image) => s"""<img src="${image.file}">"""
+      case None => "<div></div>"
+    }
+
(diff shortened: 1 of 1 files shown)


clloyd:
Hold off on merging this

clloyd:
Does it make sense to calculate the defaultHtml in the command? We already have https://github.com/guardian/media-atom-maker/blob/master/app/model/MediaAtom.scala#L37 which we should make use of?

clloyd:
I wonder if this is a more systemic problem... should we be contructing a `MediaAtom` then calling `.toThrift` rather than constructing it directly

akash1810:
> We already have https://github.com/guardian/media-atom-maker/blob/master/app/model/MediaAtom.scala#L37 which we should make use of?

As far as I could see, api2 wasn't using that... it def should though!

akash1810:
As a side note, I deployed this to CODE and it Composer wasn't reacting as I thought it would.

akash1810:
closing as really old - can revisit this later.

PR #158, opened 2016-12-16
Composer atom page

Reettaphant:
Adds the ability to create a video page in composer if a video page does not already exist. If one or more video pages exist, links to these pages are displayed instead. 

@clloyd @ShaunYearStrong


app/controllers/VideoUIApp.scala
```diff
@@ -18,13 +19,20 @@ class VideoUIApp @Inject() (val authActions: HMACAuthActions)
       case None => routes.Assets.versioned(jsFileName).toString
     }
 
+    val composerUrl = awsConfig.stage match {
+      case "PROD" => "https://composer.gutools.co.uk"
```
clloyd: Should we have this in config? It seems to be the pattern we're adopting for this type of config

public/video-ui/src/actions/VideoActions/videoUsages.js
```diff
@@ -0,0 +1,40 @@
...
+
+function requestVideoUsages(id) {
+  return {
+    type:       'VIDEO_USAGE_GET_REQUEST',
+    receivedAt: Date.now()
+  };
+}
+
+function receiveVideoUsages(usages, id) {
+  const usageObject = {};
```
clloyd: this can actually be written as 
`const usageObject = {[id]: usages.response.results}`

public/video-ui/src/components/VideoPage/VideoPage.js
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import Q from 'q';
```
clloyd: Do we need to use Q? Generally we're using `Promise` everywhere else

public/video-ui/src/components/VideoPage/VideoPage.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  pageCreate = () => {
+
+    this.setState({
+      pageCreated: true
+    });
+
+    const videoPage = {
+      elements: [
```
clloyd: This should probably be extracted out into a helper function (perhaps in a different file) that takes a video and generates the block.

public/video-ui/src/components/VideoPage/VideoPage.js
```diff
@@ -0,0 +1,127 @@
...
+        <a href= {composerUrl + '/content/' + composerIdWithUsage.composerId}>{composerIdWithUsage.usage}</a>
+      </li>
+      );
+  }
+
+  render() {
+    //If composerId exists, this means that this is the only composer page
+    //that exists, it has just been created via media-atom-maker. It may not
+    //yet be in capi, so we use this id to render it
+    if (this.props.composerId.composerId) {
```
clloyd: This naming is a bit confusing? I'd expect `this.props.composerId` to be the ID

public/video-ui/src/components/VideoPage/VideoPage.js
Reettaphant: I saw that Q was being used in the PandaReqwest service and I don't think Promise is being used here, but I replace Q with Promise if you'd like

clloyd: Maybe I'm getting confused with campaign central

clloyd: Just stick with Q

Reettaphant:
@clloyd Made some changes to address your comments.

PR #160, opened 2016-12-19
Auditing

currysoup:
Requires infrastructure updates 🍰 

New config required `aws.dynamo.auditTableName`, a table which my cloudformation script updates will setup for you.

1430d5e5f1536a66536726f415f615b6

I'm not super happy with the diffing right now, it wouldn't be super hard to make a more complete solution (which actually covers all the fields) but I don't think it's a big priority right now.


currysoup:
Updated!

abc

PR #162, opened 2016-12-20
Expiry dates

Reettaphant:
-Adds an expiry date field to media atom maker 
-Adds a poller that checks for expired atoms
-The poller marks the assets of these atoms private and marks the atom itself as private
screen shot 2016-12-20 at 15 26 59

Questions: 

-What should be the relationship between the youtube asset statuses, preview atom privacy status and published atom privacy status? 
- What is an appropriate polling interval?
- Can I have some guidance on how to make the date picker prettier? 
- Should the privacy status field for expired atoms be disabled? What if it gets disabled and someone goes to youtube and changes the status of the assets?
- We should probably add more detailed logging, but I'm not sure what we want logged.

-depends on this: https://github.com/guardian/content-atom/pull/76 and bumping atom maker library

@akash1810 @ShaunYearStrong

### DIFF
--- a/app/data/JsonConversions.scala
+++ b/app/data/JsonConversions.scala
@@ -39,7 +39,9 @@ object JsonConversions {
     (__ \ "license").writeNullable[String] and
     (__ \ "commentsEnabled").writeNullable[Boolean] and
     (__ \ "channelId").writeNullable[String] and
-    (__ \ "privacyStatus").writeNullable[PrivacyStatus]
+    (__ \ "privacyStatus").writeNullable[PrivacyStatus] and
+    (__ \ "expiryDate").writeNullable[Long]
+
   ) { metadata: Metadata =>
       (
         metadata.tags,
@@ -47,7 +49,8 @@ object JsonConversions {
         metadata.license,
         metadata.commentsEnabled,
         metadata.channelId,
-        metadata.privacyStatus
+        metadata.privacyStatus,
+        metadata.expiryDate
         )
   }
 
@@ -57,7 +60,8 @@ object JsonConversions {
     (__ \ "license").readNullable[String] and
     (__ \ "commentsEnabled").readNullable[Boolean] and
     (__ \ "channelId").readNullable[String] and
-    (__ \ "privacyStatus").readNullable[PrivacyStatus]
+    (__ \ "privacyStatus").readNullable[PrivacyStatus] and
+    (__ \ "expiryDate").readNullable[Long]
   )(Metadata.apply _)
 
   implicit val atomDataMedia = (
--- a/app/util/ExpiryPoller.scala
+++ b/app/util/ExpiryPoller.scala
@@ -18,7 +18,8 @@ case class ExpiryPoller @Inject () (
                         previewPublisher: PreviewAtomPublisher,
                         livePublisher: LiveAtomPublisher,
                         youtubeConfig: YouTubeConfig,
-                        implicit val auditDataStore: AuditDataStore
+                        implicit val auditDataStore: AuditDataStore,
+                        val awsConfig: AWSConfig
                        ){
   def start(scheduler: Scheduler): Unit = {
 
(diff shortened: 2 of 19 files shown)


app/util/ExpiryPoller.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  }
+
+  def checkExpiryDates(): Unit = {
+
+    val timeNow = new Date().getTime
+    implicit val username = PandaUser(awsConfig.expiryPollerName,
+      awsConfig.expiryPollerLastName, "expiryPoller", None)
+
+    previewDataStore.listAtoms.map(atoms => {
```
akash1810: should we filter for just Youtube atoms?

akash1810: should we also filter for atoms with an expiry date here?

app/util/Youtube.scala
```diff
@@ -122,6 +126,56 @@ case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends YouTubeBuilder {
...
+          Some(updatedAtom)
+
+        } else None
+
+      }
+      case _ => None
+    }
+  }
+
+  private def setStatusToPrivate(asset: Asset, atomId: String): Unit = {
```
akash1810: can this be a `Future` to make it async?

project/BuildVars.scala
```diff
@@ -5,7 +5,7 @@ object BuildVars {
   lazy val scroogeVersion     = "4.2.0"
   lazy val pandaVer           = "0.4.0"
   lazy val mockitoVersion     = "2.0.97-beta"
-  lazy val atomMakerVersion   = "0.1.6"
+  lazy val atomMakerVersion   = "0.1.7-SNAPSHOT"
```
akash1810: is `SNAPSHOT` correct?

public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import Picker from 'react-datepicker';
+import moment from 'moment';
+import 'react-datepicker/dist/react-datepicker.css';
```
akash1810: why do we need to do this?

public/video-ui/src/components/VideoEdit/VideoEdit.js
```diff
@@ -51,7 +53,17 @@ const VideoEdit = (props) => {
             editable={props.editable} />
 
           <Field
-            name="youtubeChannel"
+            name="videoExpiry"
+            name="expiry"
+            type="select"
+            component={VideoExpiryEdit}
+            component={YoutubeCategorySelect}
```
akash1810: does this mean the yt category can't be changed if the video has expired?

public/video-ui/src/components/VideoEdit/formComponents/VideoExpiry.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import moment from 'moment';
+import DatePickerField from '../../FormFields/DatePicker';
+
+export default class VideoExpiryEdit extends React.Component {
+
+  onUpdateExpiry = (newDate) => {
+    let newData = Object.assign({}, this.props.video, {
```
akash1810: `const`?

app/util/ExpiryPoller.scala
Reettaphant: I'd rather do these checks in the updateStatusIfExpired function so that I can use the same function for checking atoms when atom metadata gets changed.

app/data/JsonConversions.scala
```diff
@@ -39,15 +39,18 @@ object JsonConversions {
     (__ \ "license").writeNullable[String] and
     (__ \ "commentsEnabled").writeNullable[Boolean] and
     (__ \ "channelId").writeNullable[String] and
-    (__ \ "privacyStatus").writeNullable[PrivacyStatus]
+    (__ \ "privacyStatus").writeNullable[PrivacyStatus] and
+    (__ \ "expiryDate").writeNullable[Long]
```
Reettaphant: That's how it is stored in the thrift model

akash1810:
I think the build is failing because it can't locate the new version of the package 😢

PR #166, opened 2016-12-28
Clean up validation props and upgrade form library

ShaunYearStrong:


public/video-ui/src/components/VideoEdit/VideoEdit.js
```diff
@@ -58,62 +74,107 @@ const VideoEdit = (props) => {
...
               type="number"
               component={VideoDurationEdit}
-              {...props} />
+              video={props.video}
+              updateVideo={props.updateVideo}
+              editable={props.editable} />
 
-            <Field name="content-flags" component={ContentFlags} {...props} />
+            <Field
+              name="content-flags"
```
susiecoleman: Should the name be camel case?

ShaunYearStrong: Yep!

PR #170, opened 2017-01-03
fix issue with auditing when hmac api authed

akash1810:
The User of an hmac authed action has an `email` property of `""`, dynamo isn't comfortable with this and raises an exception:

`play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[AmazonDynamoDBException: One or more parameter values were invalid: An AttributeValue may not contain an empty string (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: KF11D27416N4T1945546I375DNVV4KQNSO5AEMVJF66Q9ASUAAJG)]]`

Although we're doing:

```scala
implicit val username = Option(req.user.email)
username.getOrElse("unknown user")
```

`req.user.email` is [being set to `""`](https://github.com/guardian/panda-hmac/blob/master/src/main/scala/com/gu/pandahmac/HmacAuthActions.scala#L38) so we never `OrElse`.

Fix this by pattern matching on `email` - if empty string (i.e. hmac authed) use `firstName` else (i.e. panda user) use `email`.

Pattern matching on `email` is temporary until https://github.com/guardian/panda-hmac/pull/2 is released.

NB: CDS auths with hmac and MAM API is currently 500ing because of `AmazonDynamoDBException` above... so the MAM routes in CDS are failing atm 😱.

### DIFF
--- a/app/data/AuditDataStoreProvider.scala
+++ b/app/data/AuditDataStoreProvider.scala
@@ -2,11 +2,14 @@ package data
 
 import ai.x.diff.DiffShow
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient
-import com.amazonaws.services.dynamodbv2.document.DynamoDB
-import javax.inject.{ Inject, Provider }
+import com.amazonaws.services.dynamodbv2.document.{DynamoDB, PutItemOutcome}
+import javax.inject.{Inject, Provider}
+
+import com.gu.pandomainauth.model.{User => PandaUser}
 import util.AWSConfig
 import org.joda.time.DateTime
-import model.{MediaAtom, Audit}
+import model.{Audit, MediaAtom}
+
 import scala.collection.JavaConversions._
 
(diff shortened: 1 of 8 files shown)


app/data/AuditDataStoreProvider.scala
```diff
@@ -21,18 +24,25 @@ class AuditDataStore(client: AmazonDynamoDBClient, auditDynamoTableName: String)
     db.query("atomId", id).map(Audit.fromItem).toList
   }
 
-  def auditPublish(atomId: String, username: Option[String]) = {
-    val audit = Audit(atomId, "publish", None, DateTime.now().getMillis, username.getOrElse("unknown user"))
+  private def getUsername (user: PandaUser): String = {
```
akash1810: temporary hack until https://github.com/guardian/panda-hmac/pull/2 is released, then we can pattern match like:

```scala
val username = request.user match {
  case pandaUser: PandaUser => user.email
  case hmacUser: HMACAuthenticatedService => user.name
}
```

PR #174, opened 2017-01-05
Allow for adding assets 

Reettaphant:
Fixes the form for adding new assets. 

There are still problems with creating new assets: 

- should the asset version number be calculated in the client? 
- Is the version number set correctly on the client? Currently if no asset is marked as active, the version number for the new asset is set to 0. This behaviour does not seem correct.
- The functionality for the video asset uri component does not make sense. It now contains the button for saving a new asset, and also contains logic for creating new assets and their versions numbers. It seems that there should be just one component for creating new assets.

We should decide if this functionality is going to stay in MAM and what it is going to look like. @akash1810 @clloyd

### DIFF
--- a/public/video-ui/src/actions/VideoActions/updateAsset.js
+++ b/public/video-ui/src/actions/VideoActions/updateAsset.js
@@ -1,8 +1,0 @@
-//NOTE: THIS DOESN'T SAVE THE ASSET, ONLY UPDATES THE CLIENT STATE. USE saveVideo TO SAVE
-export function updateAsset(asset) {
-  return {
-    type:       'ASSET_UPDATE_REQUEST',
-    asset:      asset,
-    receivedAt: Date.now()
-  };
-}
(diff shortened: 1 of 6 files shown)


akash1810:
Another point to consider for adding assets.

We currently do not allow the channel to be changed once the atom has been created. However we don't validate the asset being added is on that channel. So we could enter a scenario where the channel of the atom and the channel of the active asset are different. Although this doesn't break anything, it is confusing.

We should probably make the add asset form a separate endpoint with explicit instructions/warnings for this?

mbarton:
I'd vote for removing it from the UI and having a dev script that replicates what CDS does to add it

Reettaphant:
I think what @mbarton suggests makes sense. If we are going to have alternative way of uploading videos, then it is easier to use if it follows a similar workflow than currently existing tools. If we have this alternative tool, is there still a scenario where we'd have to upload assets directly in MAM?

PR #177, opened 2017-01-10
Small UX tweaks to the banners

mr-mr:
A couple of frontend tweaks for my first PR to MAM!

1. Moved the green bar to the bottom left column. 
2. Tweaked the copy of both bars: The green one to say this will publish to yt and blue I just removed the copy for now as it's duplicated when in composer.

image

cc @akash1810 @ShaunYearStrong

### DIFF
--- a/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
+++ b/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
@@ -10,7 +10,6 @@ export default class VideoSelectBar extends React.Component {
 
     return (
       <div className="bar info-bar">
-        <span className="bar__message"></span>
         <button type="button" className="bar__button" onClick={this.props.onSelectVideo}>Select this Video</button>
       </div>
     )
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
```diff
@@ -10,7 +10,7 @@ export default class VideoSelectBar extends React.Component {
 
     return (
       <div className="bar info-bar">
-        <span className="bar__message">You are in an embedded version of the media atom maker</span>
+        <span className="bar__message"></span>
```
akash1810: do we need this `span` now that its empty?

mr-mr: @akash1810 I didn't want to remove it entirely incase anyone had an objection / suggestion of other copy

ShaunYearStrong: I think we should remove the empty span for now, it can always be added again if needed later.

akash1810: none from me; we could remove it and get feedback from Editorial.

mr-mr:
All good @akash1810 @ShaunYearStrong ?

PR #178, opened 2017-01-10
Revert "Temporary error message change"

akash1810:
Reverts guardian/media-atom-maker#157

Making a revert commit because too lazy to branch 😄 

CDS is only looking at the [response code](https://github.com/guardian/content_delivery_system/blob/master/CDS/scripts/js_utils/media-atom/media-atom.js#L60-L61) now; anything other than a 200 and it continues polling. So this message can be reverted.


akash1810:
cc @marialivia16

PR #180, opened 2017-01-10
HTTP 500 alarm and webhook

mbarton:
This PR updates the cloud formation JSON to include an alarm on HTTP 500 responses which routes through to a webhook via SNS.

When deployed with the right parameters, this hooks up media-atom-maker to pager duty.

### DIFF
--- a/cloudformation/media-atom-maker.json
+++ b/cloudformation/media-atom-maker.json
@@ -55,6 +55,14 @@
         "AuditTable": {
             "Description": "Name of the audit dynamo table",
             "Type": "String"
+        },
+        "AlertWebhook": {
+          "Description": "Where CloudWatch alerts are sent",
+          "Type": "String"
+        },
+        "AlertActive": {
+          "Description": "Whether to send CloudWatch alerts to the AlertWebhook",
+          "Type": "String"
         }
     },
     "Mappings": {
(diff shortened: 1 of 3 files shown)


cloudformation/media-atom-maker.json
```diff
@@ -55,6 +55,14 @@
         "AuditTable": {
             "Description": "Name of the audit dynamo table",
             "Type": "String"
+        },
+        "AlertWebhook": {
+          "Description": "Where CloudWatch alerts are sent",
+          "Type": "String"
+        },
+        "AlertActive": {
```
akash1810: when would we make this false?

mbarton: I don't know. I was copying how the grid does it.

Happy to remove it if you want?

akash1810: ah ok. It'll probably be best to keep all alarms on for now. We can revisit later once the app has proven stability.

akash1810: is an alarm that can be turned off, worthy of being an alarm?

if a tree falls in the forest...

mbarton: to avoid the philosophical debate, I've removed it :)

Issue #187, opened 2017-01-11
Errors appear forever in the front-end


mbarton:
- Create a new video but only fill in the title, leaving the other fields blank
- A red error appears at the top of the screen
- Click the homepage button

Expected: the error clears and the homepage is displayed
Actual: the error remains and the homepage is displayed

PR #188, opened 2017-01-11
Improved server side JSON validation

mbarton:
Refactored out JSON validation in `Api2` into a single method, which returns a `BadRequest` response detailing the errors.

For example, missing out fields in the `create` UI no longer causes a 500 but returns a list of the missing fields:

```
/posterImage -> ValidationError(List(error.path.missing),WrappedArray())
/youtubeCategoryId -> ValidationError(List(error.path.missing),WrappedArray())
/channelId -> ValidationError(List(error.path.missing),WrappedArray())
```

Also update the front-end to clear out saving/publishing/searching state when an error occurs. This stops the save button from spinning forever.

NB: the error still shows up in red at the top forever (#187)

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -17,6 +17,7 @@ import model.{MediaAtom, UpdatedMetadata}
 import play.api.Configuration
 import util.atom.MediaAtomImplicits
 import play.api.libs.json._
+import play.api.libs.functional.syntax._
 import play.api.mvc.{AnyContent, Result}
 
 class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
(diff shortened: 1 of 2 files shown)


app/controllers/Api2.scala
```diff
@@ -70,69 +70,47 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
...
     implicit val user = req.user
-    try {
-      val updatedAtom = PublishAtomCommand(id).process()
-      Ok(Json.toJson(updatedAtom))
-    } catch {
-      commandExceptionAsResult
-    }
+
+    val updatedAtom = PublishAtomCommand(id).process()
+    Ok(Json.toJson(updatedAtom))
```
akash1810: as we've removed the `try/catch` what happens when `PublishAtomCommand(id).process()` do we get a 500 and a sensible message still?

app/controllers/Api2.scala
```diff
@@ -141,35 +119,47 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
...
+      val atom = command.process()
+      Ok(Json.toJson(atom))
     }
   }
 
   def getAuditTrailForAtomId(id: String) = APIHMACAuthAction { implicit req =>
     Ok(Json.toJson(auditDataStore.getAuditTrailForAtomId(id)))
   }
+
+  private def parse[T](raw: UserRequest[AnyContent])(fn: T => Result)(implicit reads: Reads[T]): Result = try {
```
akash1810: nice

app/controllers/Api2.scala
mbarton: Oops sorry. I've put back the try/catch

app/controllers/Api2.scala
```diff
@@ -2,23 +2,23 @@ package controllers
 
 import javax.inject.Inject
 
+import _root_.util.{AWSConfig, ExpiryPoller, YouTubeConfig, YouTubeVideoUpdateApi}
```
akash1810: `_root_` silly intellij!

mbarton: Yeah. I double checked it was like that before I changed things

PR #189, opened 2017-01-12
adds link to feedback form

akash1810:
screen shot 2017-01-12 at 09 22 43

### DIFF
--- a/public/video-ui/src/components/Header.js
+++ b/public/video-ui/src/components/Header.js
@@ -21,7 +21,7 @@ export default class Header extends React.Component {
               <a className="topbar__nav-link"
                  target="_blank"
                  href="https://goo.gl/forms/0KoeGOW64584Bydm2">
-                Give us feedback?
+                Give feedback
               </a>
             </nav>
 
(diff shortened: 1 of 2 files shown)


mr-mr:
👍
Nice. Brings up a question about the green being used on the edit screen buttons. Should we move to the yellow as the primary colour?

public/video-ui/src/components/Header.js
```diff
@@ -17,11 +17,18 @@ export default class Header extends React.Component {
           </div>
 
           <div className="topbar__container">
+            <nav className="topbar__nav topbar__feedback">
+              <a className="topbar__nav-link"
+                 target="_blank"
+                 href="https://goo.gl/forms/0KoeGOW64584Bydm2">
+                Give us feedback?
```
mr-mr: Can we just say Give feedback, rather than ask a question.

akash1810:
@mr-mr yellow was chosen as the logo is yellow and they both sit in the top bar. Not too fussed on colour choice, but have changed to `brandColor` (green); can always change later.

public/video-ui/src/components/Header.js
akash1810: Done.

mr-mr:
I should have been more explicit! I like the yellow, it works better on the grey!
I'll look at the colours as a whole this morning.

PR #192, opened 2017-01-13
Select video bar changes

mr-mr:
A tweak to the select video bar when choosing a video.
Making it stand out a bit more, and pulling in a couple of elements to ensure ed's are picking the correct video quickly. Thanks @clloyd for the react input!

image

No image example
image

cc @akash1810 @ShaunYearStrong @Reettaphant

### DIFF
--- a/public/video-ui/styles/components/_bar.scss
+++ b/public/video-ui/styles/components/_bar.scss
@@ -21,14 +21,21 @@
     }
   }
 
+  &__image img {
+    width: 160px;
+    height: auto;
+    margin-right: 15px;
+  }
+
   &__message {
     margin-left: 5px;
   }
 }
 
 .info-bar {
-  background-color: $cBlue32;
-  border-bottom: 1px solid darken($cBlue32, 10);
+  background-color: $color700Grey;
+  text-align: left;
+  display: inline-flex;
 }
 
 .error-bar {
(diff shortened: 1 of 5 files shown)


public/video-ui/styles/components/_bar.scss
```diff
@@ -21,14 +21,31 @@
...
+      margin-right: 15px;
+
+  }
+
+  &__image-placeholder {
+      background-color: $color800Grey;
+      width: 160px;
+      height: 90px;
+      text-align: center;
+      padding-top: 32px;
```
akash1810: Any reason for 32? Its a bit of an odd number...

mr-mr:
This change was to reflect the fact that bars have different purposes, and should separate either by distance or visually. I agree on the vertical height, I'll attempt to make it thinner.

public/video-ui/styles/components/_bar.scss
mr-mr: It's to have the 'No Image' title in the centre of the blank image box. I'll upload an example screenshot

paperboyo:
Good yellow! Isn't the font in buttons and search box wrong (Helvetica instead of Agate), though?

mr-mr:
Small steps @paperboyo small steps

mr-mr:
@akash1810 
Thinner up top now. Can always improve!
image

PR #195, opened 2017-01-17
Increase number of prod instances and add sentry integration

mbarton:
- Increases the number of instances in production to 3
  - This is for resilience although it would not address configuration problems on start-up (this will be done separately)

- Adds integration with sentry.io


cloudformation/media-atom-maker.json
```diff
@@ -523,6 +523,27 @@
...
+            "Namespace": "AWS/ELB",
+            "MetricName": "UnHealthyHostCount",
+            "Dimensions": [
+              {
+                "Name": "LoadBalancerName",
+                "Value": { "Ref": "MediaAtomMakerLoadBalancer" }
+              }
+            ],
+            "Period": "300",
+            "EvaluationPeriods": "1",
```
akash1810: is this too often?

mbarton: As with all the alarms, we should knock it down if it gets spammy. I think it's best to be high whilst in the trial

PR #196, opened 2017-01-17
Fix alarm so it does not fire as quickly

mbarton:
The unhealthy alarm was firing during deploys. This replaces it with a Route53 end-point check.

PS @akash1810 your review comment on #195 was completely correct sorry!

cc @jennysivapalan @akash1810


mbarton:
Closing in favour of removing the alarm (since it's going to fire naturally as part of deploys) and replacing it with a ping to the app via Route53

mbarton:
@akash1810 yes, agree. I've updated it to make them a parameter

mbarton:
I've increased the tolerance on the 4XX and latency alarms cc @jennysivapalan.

The 4XX alarm needed increasing because it was firing due to vulnerability scans.
The latency alarm needed increasing because uploading thumbnails to YouTube on publish was causing it to fire.

We could consider changing the code to return from the request before the YouTube upload is finished to keep things snappy for the users.

PR #197, opened 2017-01-18
Fix search results to be less confusing

### LINKED ISSUE #191: 'Active Assets' different between search and front-page
Reproducible with this atom: https://media-atom-maker.gutools.co.uk/video/videos/dc17f761-bbab-46d8-bd83-342af359f5b6

When it appears on the front-page, it correctly shows as active:

3fd0d324-238d-4729-a2af-0147270b8d62

When it appears in a search, it doesn't:

fad2dfa3-e8f9-4656-8a00-ba84e7907867

I think this could be related to the atom not appearing in CAPI correctly.

mbarton:
Currently, video search results are returned without an image and marked as 'No Active Assets' even if they have some. It's kind of confusing and threw me off on a tangent when debugging some other things.

The reason (I think) is that search is powered by CAPI results which are then massaged into the right shape client side. This process was not including the images, which is now fixed. The fix also hides the 'active assets' text when displaying results from the search.

**Before**

1d1da401-422f-43ec-9b49-c5bbe3a9751e

**After**

cbd6f85a-b6a1-419c-9a45-8dd90a792bf7

**Notes to reviewers**

I've hidden the 'active assets' text by passing `activeVersion` as -1 from search results. This is 
smelly but was the best I could think of without touching too much code. Any other suggestions very welcome!

cc @ShaunYearStrong @akash1810 @Reettaphant 
Fixes #191

### DIFF
--- a/public/video-ui/src/actions/VideoActions/getVideos.js
+++ b/public/video-ui/src/actions/VideoActions/getVideos.js
@@ -54,6 +54,7 @@ export function getVideos() {
 function adaptCapiAtom(atom) {
   const ret = {
     id: atom.id,
+    activeVersion: -1, // not known
     title: atom.data.media.title,
     contentChangeDetails: atom.contentChangeDetails,
     assets: atom.data.media.assets
--- a/public/video-ui/src/actions/VideoActions/getVideos.js
+++ b/public/video-ui/src/actions/VideoActions/getVideos.js
@@ -51,22 +51,33 @@ export function getVideos() {
   };
 }
 
+function adaptCapiAtom(atom) {
+  const ret = {
+    id: atom.id,
+    title: atom.data.media.title,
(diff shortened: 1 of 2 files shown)


public/video-ui/src/actions/VideoActions/getVideos.js
```diff
@@ -51,22 +51,34 @@ export function getVideos() {
   };
 }
 
+function adaptCapiAtom(atom) {
+  const ret = {
+    id: atom.id,
+    activeVersion: -1, // not known
```
akash1810: why `-1` over `undefined`?

mbarton: There are atoms in the DB that don't have that property (I think they are ones that have just been created and not published?) so I wanted to make an explicit distinction between those and atoms that have come from search results.

PR #198, opened 2017-01-18
No unpub embed

Reettaphant:
- Don't allow for embedding unpublished media atoms from composer and creating composer pages for unpublished atoms from media-atom-maker
- There are plans for adding unpublished icons to media-atoms, so we don't have to change the way that media-atoms appear in search now

### DIFF
--- a/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
+++ b/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
@@ -20,6 +20,14 @@ export default class VideoSelectBar extends React.Component {
 
   }
 
+  renderEmbedButton() {
+    return <button type="button" className="bar__button" onClick={this.props.onSelectVideo}>Select this Video</button>
+  }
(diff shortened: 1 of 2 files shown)


public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
```diff
@@ -15,19 +15,45 @@ export default class VideoSelectBar extends React.Component {
     return <div className="bar__image-placeholder">No Image</div>
   }
 
+  isVideoPublished() {
+    console.log('is video published ', this.props.video);
```
mbarton: Remove?

Reettaphant:
There is still a way for unpublished atoms to end up to published pages. Because composer embeds the preview versions of atoms, a user would have to edit the media atom and not publish those changes to accomplish this. Fixing this would require making changes to composer, I think it is best to talk about preview/published atoms in general before making these changes.

PR #206, opened 2017-01-20
Use versioned config bucket in S3

mbarton:
Ronseal. Point media-atom-maker at a new S3 bucket that is just for config and has versioning enabled.

I will delete the configuration from the old S3 bucket once this PR is merged.


jennysivapalan:
👍 📁

PR #207, opened 2017-01-20
Use amigo and riff-raff.yaml

jennysivapalan:
This sets us up to use Amigo and so be able to update the AMI regularly. Projects that use Amigo tend to use a Debian file so I think I've done the right steps to get that working for us too. I've kept the `.tgz` in for now as I think it will help with the step process to get this out.

I've also updated the project to use `riff-raff.yaml` as the current `deploy.json` did not allow to just do S3 upload which is what I needed to check this was going to work. So I made the move to update the new format.

Steps to get this into production:
- [ ] Apply the cloudformation and use an AMI with Ed tool ssh access
- [ ] Merge PR and run a deployment
- [ ] Remove reference to `.tgz` - separate PR


cloudformation/media-atom-maker.json
```diff
@@ -357,7 +338,9 @@
                                 "aws.dynamo.auditTableName=", { "Ref": "AuditTable" }, "\n",
                                 "EOF\n",
 
-                                "systemctl start media-atom-maker\n"
+                                "aws s3 cp s3://atom-maker-dist/media-service/", {"Ref":"Stage"}, "/media-atom-maker/media-atom-maker.deb /tmp",
```
jennysivapalan: I'm not that happy with current structure of the dist folder - feels like we have two too many folders. I won't change it now but I'm happy to sort this out a later date. cc/ @akash1810 @mbarton @Reettaphant

jennysivapalan:
This doesn't quite work as the config files don't get loaded in the right place. I will address that as a separate PR and then come back to this.

jennysivapalan:
Closing - will re-open once I fixed configuration

PR #208, opened 2017-01-24
keep query string when redirecting

akash1810:
to keep embedded mode working

`embeddedMode` is triggered with the `?embeddedMode=true` query string. Keep it when we redirect.

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -43,7 +43,7 @@ class Support @Inject() (val authActions: HMACAuthActions,
     }
   }
 
-  def legacyVideosEndpointRedirect(path: String) = Action {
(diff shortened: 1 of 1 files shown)


akash1810:
cc @clloyd

app/controllers/Support.scala
```diff
@@ -43,7 +43,7 @@ class Support @Inject() (val authActions: HMACAuthActions,
     }
   }
 
-  def legacyVideosEndpointRedirect(path: String) = Action {
-    Redirect(s"/$path")
+  def legacyVideosEndpointRedirect(path: String) = Action { request =>
+    Redirect(s"/$path?${request.rawQueryString}")
```
akash1810: `?` is needed

PR #212, opened 2017-01-25
Clearer publishing logic

Reettaphant:
Clears publishing logic: 

- Fixes a bug that prevented the publish action from pushing changes to youtube
- Updating the atom title now also updates the youtube title
- Pressing the publish button is the only way to publish an atom: atom will no longer be published when an atom is added to it
- Publishing is the only way of syncing with youtube (for some reason the update metadata method used by cds did this as well

### DIFF
--- a/app/model/Metadata.scala
+++ b/app/model/Metadata.scala
@@ -4,7 +4,7 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
 case class UpdatedMetadata (
-  title: String,
+  title: Option[String],
   description: Option[String],
   tags: Option[List[String]],
   categoryId: Option[String],
@@ -16,7 +16,7 @@ case class UpdatedMetadata (
 
 object UpdatedMetadata {
   implicit val metadataRead: Reads[UpdatedMetadata] = (
-    (__ \ "title").read[String] ~
+    (__ \ "title").readNullable[String] ~
     (__ \ "description").readNullable[String] ~
     (__ \ "tags").readNullable[List[String]] ~
     (__ \ "categoryId").readNullable[String] ~
--- a/app/model/Metadata.scala
+++ b/app/model/Metadata.scala
@@ -4,6 +4,7 @@ import play.api.libs.functional.syntax._
 import play.api.libs.json._
 
 case class UpdatedMetadata (
+  title: String,
   description: Option[String],
   tags: Option[List[String]],
   categoryId: Option[String],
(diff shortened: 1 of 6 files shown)


app/model/commands/ActiveAssetCommand.scala
```diff
@@ -58,8 +59,10 @@ case class ActiveAssetCommand(atomId: String, youtubeId: String)
                 duration = ytAssetDuration
               ))
 
-            UpdateAtomCommand(atomId, MediaAtom.fromThrift(updatedAtom)).process()
-            PublishAtomCommand(atomId).process()
+            publishedDataStore.getAtom(atomId) match {
```
mbarton: "Pressing the publish button is the only way to publish an atom: atom will no longer be published when an atom is added to it"

Does that cover this code as well? My naive reading of it is that we re-publish an already published atom if the asset is changed

app/model/commands/PublishAtomCommand.scala
```diff
@@ -50,14 +50,25 @@ case class PublishAtomCommand(id: String)(implicit val previewDataStore: Preview
...
+              atom.title,
+              atom.description,
+              Some(atom.tags),
+              atom.youtubeCategoryId,
+              atom.license,
+              atom.privacyStatus,
+              atom.expiryDate,
+              atom.plutoProjectId))
+          }
+          case None =>
```
mbarton: Do we need to handle this case? Maybe `flatMap` into the search for the assetId as well to avoid calling `.get`

Reettaphant: This would be the case when the atom has no video so we won't have to make any changes to youtube.

app/model/Metadata.scala
```diff
@@ -15,6 +16,7 @@ case class UpdatedMetadata (
 
 object UpdatedMetadata {
   implicit val metadataRead: Reads[UpdatedMetadata] = (
+    (__ \ "title").read[String] ~
```
akash1810: The `update-metadata` is used by CDS to [set the plutoId](https://github.com/guardian/content_delivery_system/blob/master/CDS/scripts/js_utils/media-atom/media-atom.js#L126-L129). It sends a payload of `{plutoId: 'KP-xxxxx'}`.

Can this be `readNullable[String]` so that cds doesn't need to send the title in the payload too?

Alternatively, we can get rid of the `update-metadata` endpoint, in place of a `set-pluto-id` one?

Reettaphant: I've made the title into an option now.

PR #213, opened 2017-01-25
Set inactive assets to private when publishing the atom

### LINKED ISSUE #211: Mark old assets as private
If a new asset is activated, the video for the previously active asset should be set as private

mbarton:
When the active asset is changed, make all other assets private so that there is one single 'source of truth' for the atom that is publicly visible.

Fixes #211

### DIFF
--- a/app/util/Youtube.scala
+++ b/app/util/Youtube.scala
@@ -7,7 +7,6 @@ import java.util.Date
 import javax.inject.{Inject, Singleton}
 
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
-import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.http.InputStreamContent
 import com.google.api.client.http.javanet.NetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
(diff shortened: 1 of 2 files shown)


app/util/Youtube.scala
```diff
@@ -151,28 +149,26 @@ case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends YouTubeBuilder {
     }
   }
 
-  private def setStatusToPrivate(asset: Asset, atomId: String): Unit = {
-    if (asset.platform == Youtube) {
-      YouTubeVideoInfoApi(config).getVideo(asset.id, "snippet,status") match {
-        case Some(video) => {
-          protectAgainstMistakesInDev(video)
+  def setStatusToPrivate(id: String, atomId: String): Unit = {
```
akash1810: could we make this return a `Future` to make it async?

mbarton: Yeah, it will then make the requests to YouTube in parallel which is nice. Did you have any other reason for wanting it?

mbarton: Just discussed this and agreed to address it later on as part of making more of a service for asset management

PR #223, opened 2017-01-27
Add more application logging

mbarton:
This PR adds lots of logging statements that should make it easier to track down problems.

It also disables the default logging from the YouTube API client - the noise from which made it difficult to track other things down.


mbarton:

jennysivapalan:
👍 🌴

PR #224, opened 2017-01-27
Disable sentry in dev

mbarton:
Only catch and send JS errors to sentry if we are running in CODE or PROD

cc @akash1810 @Reettaphant 
Fixes #224

### DIFF
--- a/public/video-ui/src/app.js
+++ b/public/video-ui/src/app.js
@@ -29,7 +29,8 @@ const store = configureStore();
 const config = extractConfigFromPage();
 
 // publish uncaught errors to sentry.io
-Raven.config(config.ravenUrl).install();
+if(config.stage !== 'DEV')
+  Raven.config(config.ravenUrl).install();
 
 setStore(store);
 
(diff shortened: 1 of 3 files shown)


public/video-ui/src/app.js
```diff
@@ -29,7 +29,8 @@ const store = configureStore();
 const config = extractConfigFromPage();
 
 // publish uncaught errors to sentry.io
-Raven.config(config.ravenUrl).install();
+if(config.stage !== 'DEV')
```
akash1810: Can we limit to PROD only as we sometime release feature branches to CODE.

mbarton: Done

public/video-ui/src/app.js
```diff
@@ -29,7 +29,8 @@ const store = configureStore();
 const config = extractConfigFromPage();
 
 // publish uncaught errors to sentry.io
-Raven.config(config.ravenUrl).install();
+if(config.stage !== 'PROD')
```
akash1810: should this be `=== 'PROD'` ?

mbarton: Oops yes it should

PR #226, opened 2017-01-27
Refactor private config and application.conf

jennysivapalan:
In order to move .deb we need to stop copying in the application.conf and restarting the application. Also the way we currently pull in dev configuration is quite different to our other applications. This PR looks at moving this project to similar configuration as other Guardian repos.

Each environment will have a `atom-maker.private.conf` which is stored in S3. On start up an instance will download the configuration file and move it `/etc/gu`. The `application.conf` references this file to ensure properties are loaded in. Each environment also has a `[stage].public.conf` which can be used to for properties that can be checked in. As you can see they are quite light, perhaps some of the private config can be moved in but I'd like to pair with someone on that.

As we use one DEV stack I've put the properties for those into the DEV version `atom-maker.private.conf`. So we can remove all the bits that try to generate the properties.

To do before merging:
- [ ] Add `atom-maker.private.conf` to S3 for PROD


README.md
```diff
@@ -17,13 +17,10 @@ publisher that pushes media atoms into a Kinesis stream.
...
-0. Run `./scripts/setup-dev-conf` to generate an `application.conf` file. You need to pass to this script the name of the stack that you want to use
-   (usually this is the default `media-atom-maker-dev` stack),
-   aws profile for panda (e.g. `media-service`) and the domain for the app
-   (`local.dev-gutools.co.uk`)
-0. If you need your own dev-stack, you can create it by using the
-   `media-atom-maker-dev.yml` template and by generating the `application.conf`
-   file from the sample `reference.conf` file in the cloudformation folder
+0. You'll need to get the private configuration information from the S3 bucket (atom-maker-conf) and put it in your /etc/gu folder. This will be referenced in `application.conf`
+ on start up.
+`aws s3 cp s3://atom-maker-conf/DEV/atom-maker.private.conf /etc/gu/atom-maker.private.conf --profile media-service`
```
akash1810: would `media-atom-maker.private.conf` be better, as there's less chance of clashing with any config from `atom workshop`.

cloudformation/media-atom-maker.json
```diff
@@ -345,17 +344,16 @@
...
+                                "mkdir /etc/gu \n",
+                                "aws s3 cp 's3://atom-maker-conf/",
                                 {
-                                    "Ref": "Stage"
+                                  "Ref": "Stage"
                                 },
-                                ".application.conf' ",
-                                "/home/media-service/media-atom-maker/conf/application.conf\n",
-                                "cat >> /home/media-service/media-atom-maker/conf/application.conf<<'EOF'\n",
-                                "aws.dynamo.tableName=", { "Ref": "AtomMakerTable" }, "\n",
```
akash1810: The only argument I can think of against removing these is that they are passed in as params - if the params change but the file in s3 doesn't then bad things happen. We should probably only worry about it if it becomes an issue, however...

README.md
jennysivapalan: yes - I followed convention of the bucket but yes think being more specific with conf file name would be better.

cloudformation/media-atom-maker.json
jennysivapalan: I think it's the same for both, you'd have to recycle the instance to get the new dynamo config? 
I'll remove the params from the cloudformation as they won't be used anymore.

jennysivapalan: Turns out these parameters are used in this cloudformation. I'll do a similarapproach to above but add it to the private conf. In the long run we could put the creation of the dynamo tables in this cloudformation. Then if we want to create a new stack we just have one place to look.

jennysivapalan:
Have made the changes and tested on CODE. @akash1810 mind giving this a 👍 and I'll try get this out to PROD later today?

cloudformation/media-atom-maker.json
```diff
@@ -345,17 +344,23 @@
...
+                                "aws s3 cp 's3://atom-maker-conf/",
                                 {
-                                    "Ref": "Stage"
+                                  "Ref": "Stage"
                                 },
-                                ".application.conf' ",
-                                "/home/media-service/media-atom-maker/conf/application.conf\n",
-                                "cat >> /home/media-service/media-atom-maker/conf/application.conf<<'EOF'\n",
+                                "/media-atom-maker.private.conf' /tmp\n",
+                                "cat >> /tmp/media-atom-maker.private.conf<<'EOF'\n",
```
akash1810: minor question: why cat to the existing file rather than a new file and `include` it?

jennysivapalan: errm because I copied the previous implementation 😊 
Is the way you mentioned a preferred method?

akash1810: I only mention it on the back of the newline problem we had earlier.

Happy either way.

PR #228, opened 2017-01-30
Add ‘preview’ option to embedded mode

mbarton:
At the moment, both Pluto and Composer use the same `embeddedMode=true` flag when iframing the media-atom-maker. Previous changes only allowing the selection of published videos mean that the atom must be published before it can be selected from within Pluto.

This change adds two states to the flag: `live` and `preview`. In live mode, only published atoms can be selected. In preview mode, anything can be selected. Pluto should be updated to use the `preview` mode, with Composer using the `live` mode.

The change is backwards compatible, so that `embeddedMode=true` is equivalent to `embeddedMode=live`. This means Composer does not immediately need to be changed.


public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
```diff
@@ -16,44 +16,43 @@ export default class VideoSelectBar extends React.Component {
...
+
+    const isPublished = this.props.video && this.props.video.contentChangeDetails && this.props.video.contentChangeDetails.published;
+    const embedButton = <button type="button" className="bar__button" onClick={this.props.onSelectVideo}>Select this Video</button>;
+
+    switch(this.props.embeddedMode) {
+      case "preview":
+        return embedButton;
+
+      case "live":
+      case "true":
```
akash1810: we should remove this option once pluto has gone out

mbarton: Composer will need updating as well

akash1810:
can we ship this?

mbarton:
Yep. I was waiting for https://github.com/guardian/pluto/pull/64 to go through first

PR #230, opened 2017-01-30
Publishing button visibility

Reettaphant:
Request the published atom from the server to be able to look for unpublished changes.

It looks a bit strange at the moment because unsaved changed also appear as unpublished changes but once we start saving all edits immediately, the problem will go away. 

@akash1810 @mbarton

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -68,6 +68,13 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
     }
   }
 
+  def getPublishedMediaAtom(id: String) = APIHMACAuthAction {
+    publishedDataStore.getAtom(id) match {
+      case Some(atom) => Ok(Json.toJson(MediaAtom.fromThrift(atom)))
+      case None => Ok(Json.obj())
+    }
+  }
+
   def publishMediaAtom(id: String) = APIHMACAuthAction { implicit req =>
     implicit val user = req.user
 
--- a/conf/routes
+++ b/conf/routes
@@ -16,6 +16,7 @@ PUT     /api/atom/:id/publish           controllers.Api.publishAtom(id)
 
 POST    /api2/atoms                     controllers.Api2.createMediaAtom
 GET     /api2/atoms                     controllers.Api2.getMediaAtoms()
+GET     /api2/atoms/published/:id       controllers.Api2.getPublishedMediaAtom(id)
 GET     /api2/atoms/:id                 controllers.Api2.getMediaAtom(id)
 PUT     /api2/atoms/:id                 controllers.Api2.putMediaAtom(id)
 POST    /api2/atoms/:id/assets          controllers.Api2.addAsset(id)
--- a/conf/routes
+++ b/conf/routes
@@ -16,7 +16,7 @@ PUT     /api/atom/:id/publish           controllers.Api.publishAtom(id)
 
 POST    /api2/atoms                     controllers.Api2.createMediaAtom
 GET     /api2/atoms                     controllers.Api2.getMediaAtoms()
-GET     /api2/atoms/published/:id       controllers.Api2.getPublishedMediaAtom(id)
+GET     /api2/atoms/:id/published       controllers.Api2.getPublishedMediaAtom(id)
 GET     /api2/atoms/:id                 controllers.Api2.getMediaAtom(id)
 PUT     /api2/atoms/:id                 controllers.Api2.putMediaAtom(id)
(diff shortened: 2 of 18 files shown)


conf/routes
```diff
@@ -16,6 +16,7 @@ PUT     /api/atom/:id/publish           controllers.Api.publishAtom(id)
 
 POST    /api2/atoms                     controllers.Api2.createMediaAtom
 GET     /api2/atoms                     controllers.Api2.getMediaAtoms()
+GET     /api2/atoms/published/:id       controllers.Api2.getPublishedMediaAtom(id)
```
mbarton: `/api2/atoms/:id/published` feels more natural to me?

public/video-ui/src/components/ReactApp.js
```diff
@@ -14,6 +14,7 @@ class ReactApp extends React.Component {
   componentWillMount() {
     if (this.props.params.id) {
       this.props.appActions.getVideo(this.props.params.id);
+      this.props.appActions.getPublishedVideo(this.props.params.id);
```
mbarton: Do we need to have two method calls (and two redux actions) for this? Could we combine them into a single one, even if it ends up making two requests to the server?

public/video-ui/src/util/hasUnpublishedChanges.js
```diff
@@ -0,0 +1,35 @@
...
+  if (!publishedVideo) {
+    return true;
+  }
+
+  const previewPoster = previewVideo.posterImage;
+  const publishedPoster = publishedVideo.posterImage;
+  const previewAssets = previewVideo.assets;
+  const publishedAssets = publishedVideo.assets;
+
+  if (
```
mbarton: If we're doing stuff like this, might it be worth just doing a JSON.stringify and comparing the results? I'm slightly worried that we might add a field in the future and forget to change this code

app/controllers/Api2.scala
```diff
@@ -68,6 +68,13 @@ class Api2 @Inject() (implicit val previewDataStore: PreviewDataStore,
     }
   }
 
+  def getPublishedMediaAtom(id: String) = APIHMACAuthAction {
+    publishedDataStore.getAtom(id) match {
+      case Some(atom) => Ok(Json.toJson(MediaAtom.fromThrift(atom)))
+      case None => Ok(Json.obj())
```
akash1810: is `Ok` the right response code here?

public/video-ui/src/reducers/publishedVideoReducer.js
```diff
@@ -0,0 +1,13 @@
+export default function video(state = null, action) {
```
akash1810: why `null` over `undefined`?

public/video-ui/src/util/hasUnpublishedChanges.js
```diff
@@ -0,0 +1,35 @@
...
+    previewVideo.title === publishedVideo.title &&
```
akash1810: oh wow! would object comparison be cleaner/more generic?

public/video-ui/src/components/ReactApp.js
Reettaphant: We don't, I tried combining them into a single one originally, but in the end I thought that this would be clearer.

app/controllers/Api2.scala
Reettaphant: I think so, there is nothing wrong if nothing gets returned, it just means that the atom has not been published. Happy to hear if you have a better suggestion though.

public/video-ui/src/reducers/publishedVideoReducer.js
Reettaphant: I think it makes sense because it is something that has been explicitly set to null rather than something that has not been defined. Its also what the other reducers use.

Reettaphant:
@mbarton @akash1810 The preview and the publish atoms are not entirely similar so I'm comparing separate properties (in a shorter way!). I guess I could specify a list of properties not to check instead of a list of properties to check to avoid the problem @mbarton referred to but this way is more explicit and we can also end up adding properties we don't need to check in the future. I'm now comparing active assets instead of assets, I think that this detects all unpublished changes (new assets in preview that are not active do not have to published).

public/video-ui/src/reducers/publishedVideoReducer.js
ShaunYearStrong: I think https://stackoverflow.com/questions/5101948/javascript-checking-for-null-vs-undefined-and-difference-between-and explains it well. `null` is a bit more explicit, it's an empty object.

PR #231, opened 2017-01-30
add eslint

akash1810:
rules added:

```js
  "rules": {
    "react/prop-types": "warn",
    "jsx-quotes": ["error", "prefer-double"],
    "prefer-const": "error",
    "semi": ["error", "always"]
  }
```

I'd prefer `react/prop-types` to be `error` but there are so many!

### DIFF
--- a/public/video-ui/src/components/ReactApp.js
+++ b/public/video-ui/src/components/ReactApp.js
@@ -73,8 +73,8 @@ function mapStateToProps(state) {
 
 function mapDispatchToProps(dispatch) {
   return {
-    appActions: bindActionCreators(Object.assign({}, updateSearchTerm, getVideo, publishVideo), dispatch)
+    appActions: bindActionCreators(Object.assign({}, updateSearchTerm, getVideo, publishVideo, saveVideo), dispatch)
   };
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(ReactApp)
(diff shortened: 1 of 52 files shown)


public/video-ui/src/components/ReactApp.js
```diff
@@ -73,8 +73,8 @@ function mapStateToProps(state) {
 
 function mapDispatchToProps(dispatch) {
   return {
-    appActions: bindActionCreators(Object.assign({}, updateSearchTerm, getVideo, publishVideo), dispatch)
+    appActions: bindActionCreators(Object.assign({}, updateSearchTerm, getVideo, publishVideo, saveVideo), dispatch)
```
mbarton: Just checking that this was intended? It doesn't seem like a linting thing

akash1810:
> Will the eslint warnings appear in a report anywhere, or just locally when you build?

js linting is now part of the circle build, via `npm test`, which in turn runs `npm run lint`.

public/video-ui/src/components/ReactApp.js
akash1810: the linter was complaining about an unused import here - https://github.com/guardian/media-atom-maker/pull/231/files#diff-1fd58d0340aed2bb63a38c5f92e9eb65R63

I'll release to CODE before merging to make sure I've not broken anything.

PR #237, opened 2017-02-01
edit video titles and categories

Reettaphant:
Updating videos and categories was broken, this fixes it. @mbarton @akash1810


Reettaphant:
@mbarton Now this always passes in updateVideo function that may update or save and update the video, depending on whether we are creating/editing. Both editing and creating videos works!

PR #238, opened 2017-02-01
fix image selection on create form

akash1810:
bleugh, react!


akash1810:
tested on CODE and works!

PR #240, opened 2017-02-02
Add a description field

Reettaphant:
Add a description field to MAM and allow for creating composer pages with descriptions
@akash1810 @mbarton

### DIFF
--- a/public/video-ui/src/actions/VideoActions/videoPageCreate.js
+++ b/public/video-ui/src/actions/VideoActions/videoPageCreate.js
@@ -24,12 +24,12 @@ function errorReceivingVideoPageCreate(error) {
   };
 }
 
-export function createVideoPage(id, title, composerUrl, data) {
+export function createVideoPage(id, metadata, composerUrl, data) {
   return dispatch => {
 
     dispatch(requestVideoPageCreate());
 
-    return VideosApi.createComposerPage(id, title, composerUrl)
+    return VideosApi.createComposerPage(id, metadata, composerUrl)
     .then(res => {
 
       const pageId = res.data.id;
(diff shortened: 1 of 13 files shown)


public/video-ui/src/actions/VideoActions/videoPageCreate.js
```diff
@@ -24,12 +24,12 @@ function errorReceivingVideoPageCreate(error) {
   };
 }
 
-export function createVideoPage(id, title, composerUrl, data) {
+export function createVideoPage(id, metadata, composerUrl, data) {
```
akash1810: what's the difference between `metadata` and `data`? could they be merged?

public/video-ui/src/services/VideosApi.js
```diff
@@ -85,9 +85,20 @@ export default {
...
+  createComposerPage(id, metadata, composerUrlBase) {
+
+    const initialComposerUrl = composerUrlBase + '/api/content?atomPoweredVideo=true&originatingSystem=composer&type=video';
+    const propertiesToSend= ['title', 'description'];
+
+    const composerUrl = propertiesToSend.reduce((url, property) => {
+      if (metadata[property]) {
+        url += '&initial' + property.charAt(0).toUpperCase() + property.slice(1) + '=' + metadata[property];
+      }
+      return url;
```
akash1810: could we avoid the `string +=` mutation?

maybe something like:

```js
const x = propertiesToSend.reduce((queryString, prop) => {
  queryString.push('foo=bar');
}, []);

const composerUrl = initialComposerUrl + x.join('&');
```

public/video-ui/src/actions/VideoActions/videoPageCreate.js
Reettaphant: I've renamed the data to videoBlock because it is misleading. It gets posted to flex when video is added to the page and it gets parsed from the metadata before createVideoPage action is called

jennysivapalan:
👍

PR #247, opened 2017-02-03
Use latest AMI on a deploy

jennysivapalan:
🎉


jennysivapalan:
I need to add tags to cloudformation stack before merging this. I'll do this tomorrow.

PR #248, opened 2017-02-06
Atom status

Reettaphant:
This PR aligns the process of setting the atom status with the publishing workflow. 

- All newly created atoms are created with status set to private
- They cannot be published until the status is changed. 
- Introduces the concept of preflight checks to media atom maker. As a part of this work, the publish button messaging is also streamlined (no notifications next to the button, the text on the button is either Published, Publishing or Publish). 
- After some thought I decided that at the moment, only the atom status should be in the preflight checks because it is the only thing with different validation in preview and published state. Other errors should be displayed as error messages.

screen shot 2017-02-06 at 14 50 25
screen shot 2017-02-06 at 14 50 10

@mbarton @akash1810 @mr-mr

### DIFF
--- a/public/video-ui/src/util/isPublishable.js
+++ b/public/video-ui/src/util/isPublishable.js
@@ -1,0 +1,15 @@
+import { previewPrivacyState } from '../constants/privacyStates';
+
+export function isPublishable(video) {
+
+  let publishState = {
+    errors: []
+  };
+
+  if (previewPrivacyState === video.privacyStatus) {
+    publishState.errors.push('privacyStatus');
+  }
+
+  return publishState;
+}
+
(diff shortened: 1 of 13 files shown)


public/video-ui/src/util/isPublishable.js
```diff
@@ -0,0 +1,15 @@
+import { previewPrivacyState } from '../constants/privacyStates';
+
+export function isPublishable(video) {
```
akash1810: Bit of an odd name - I'd expect this to return a `bool` but it returns an `Object`?

Reettaphant: Yes it is! It was returning a boolean at first I think, I'm changing the name.

PR #249, opened 2017-02-06
icon component

akash1810:
DRYing out the code a little

no functional change, just simplifying the code a little, and removing some classes

### DIFF
--- a/public/video-ui/styles/layout/_icons.scss
+++ b/public/video-ui/styles/layout/_icons.scss
@@ -20,14 +20,11 @@
     -moz-osx-font-smoothing: grayscale;
 
     vertical-align: middle;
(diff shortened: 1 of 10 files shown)


public/video-ui/styles/layout/_icons.scss
```diff
@@ -20,24 +20,22 @@
...
 
 .icon__done,
 .icon__edit {
   background-color: $color600Grey;
   width: 25px;
   height: 25px;
-  padding: 4px 5px;
   cursor: pointer;
-  margin-right: 10px;
+  display:flex;
```
akash1810: `icon__done/edit` has a custom height/width, align the icon in the centre

PR #255, opened 2017-02-07
don't display default properties

Reettaphant:
The select boxes in both atom edit and create allow for choosing empty values for required atom fields, causing errors on the server. This PR prevents this by hiding the empty values from the select boxes when the atom already has a property chosen. @mbarton @akash1810


public/video-ui/src/components/VideoEdit/formComponents/VideoCategory.js
```diff
@@ -4,7 +4,14 @@ import { videoCategories } from '../../../constants/videoCategories';
 
 export default class VideoCategorySelect extends React.Component {
 
+  defaultOPtion = "Select a category";
```
akash1810: typo?

PR #256, opened 2017-02-07
populate with empty video

Reettaphant:
The `populateEmptyVideo` action that is responsible for uploading the create form with an empty video and for resetting the form to empty state did not exist. This PR adds this functionality by updating the video with `blankVideoData`. 
- For this to work, optional description field is represented with null and passed as an empty string to the text field.

@akash1810 @mbarton

### DIFF
--- a/public/video-ui/src/components/VideoEdit/formComponents/PrivacyStatus.js
+++ b/public/video-ui/src/components/VideoEdit/formComponents/PrivacyStatus.js
@@ -17,7 +17,7 @@ export default class PrivacyStatusSelect extends React.Component {
     return (
       <SelectBox
         fieldName="Privacy Status"
-        fieldValue={this.props.video.privacyStatus}
+        fieldValue={this.props.video ? this.props.video.privacyStatus : ''}
         selectValues={privacyStates || []}
         onUpdateField={this.updatePrivacyStatus}
         video={this.props.video}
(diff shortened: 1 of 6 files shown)


public/video-ui/src/components/VideoEdit/formComponents/PrivacyStatus.js
```diff
@@ -19,7 +19,7 @@ export default class PrivacyStatusSelect extends React.Component {
     return (
       <SelectBox
         fieldName="Privacy Status"
-        fieldValue={this.props.video.privacyStatus}
+        fieldValue={this.props.video ? this.props.video.privacyStatus : ''}
```
akash1810: why `''`?

Reettaphant: React generates an error if a field that is undefined or null gets assigned a value, it needs to be an empty string instead: https://github.com/facebook/react/issues/6222

akash1810: sorry, I meant why `''` rather than `'private'`

Reettaphant: Its only there for when the component is still loading, I don't know if it makes sense to say that a video that hasn't loaded yet is private but I don't think it would make much of a difference in any case.

PR #257, opened 2017-02-07
Fix sorting on front-page by removing comparator

mbarton:
Fixes a nasty error that was caused by a dodgy comparator on `MediaAtom`:

```
IllegalArgumentException: Comparison method violates its general contract!
```

The fix replaces the ordering on `MediaAtom` itself in favour of sorting in the list atoms route.
Atoms that do not have a created time appear after those that do, in any order.


Reettaphant:
👍

PR #261, opened 2017-02-07
more explicit User creation

akash1810:
`User` is defined as

```scala
case class User(email: String, firstName: Option[String], lastName: Option[String])
```

So we're setting `firstName` as `email`, `lastName` as `firstName` and `email` as `lastName` 😂. Lets be more explicit in object creation!


akash1810:
related to https://github.com/guardian/media-atom-maker/pull/260

PR #263, opened 2017-02-13
Upload file to S3

mbarton:
This is the first in a series of PRs that will add direct video uploading functionality to media-atom-maker.

This PR lays the groundwork:

- S3 bucket that will receive the uploads 
- server end-point that uses AWS STS to generate access credentials for the bucket
- simple client-side form that uploads the entire file in one go

simple_upload

**Notes to reviewers**

The STS credentials are generated by assuming the `LimitedUploadRole` which only has access to a prefixed part of the S3 bucket. An additional policy is also specified limiting the access to a new key generated for the upload. As such, the credentials passed to the client should only be able to upload to the bucket and live for 15 minutes (the minimum allowed by STS).

Unfortunately, federated users are not allowed to make assume role requests. In order to test in `DEV`, I have updated the config files to use the credentials of `MediaAtomUser` directly. When deployed, the instance credentials are used as normal.

If anyone knows a better way of importing the `aws-sdk` in `UploadsApi.js`, please let me know.


build_config/webpack.dev.conf.js
```diff
@@ -26,6 +26,10 @@ module.exports = {
         test: /\.(ttf|eot|svg|gif)(\?v=[0-9].[0-9].[0-9])?$/,
         loader: "file-loader?name=[name].[ext]"
       }
+    ],
+    // http://andrewhfarmer.com/aws-sdk-with-webpack/
+    noParse: [
```
clloyd: ewww

public/video-ui/src/components/VideoAssets/VideoAssets.js
```diff
@@ -110,14 +111,28 @@ class VideoAssets extends React.Component {
     return false;
   }
 
+  renderHeader() {
+    const atomId = this.props.video.id;
+
+    function goToUploadPage() {
+      browserHistory.push(`/videos/${atomId}/upload`);
```
clloyd: Any reason you're using this rather than `<Link />` from react-router... also I wouldn't define a function within the renderHeader function

public/video-ui/src/components/VideoUpload/Upload.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import {browserHistory} from 'react-router';
+import Icon from '../Icon';
+
+class VideoUpload extends React.Component {
+  getUploadPolicy = (event) => {
+    if(!this.props.video)
```
clloyd: Curly braces please (and a space after if)

public/video-ui/src/components/VideoUpload/Upload.js
```diff
@@ -0,0 +1,117 @@
...
+
+    if(event.target.files.length == 0) {
+      this.props.uploadActions.clearUploadPolicy();
+    } else {
+      this.props.uploadActions.getUploadPolicy(this.props.video.id, event.target.files[0]);
+    }
+  };
+
+  startUpload = () => {
+    if(!this.props.upload)
```
clloyd: Curly Braces please

public/video-ui/src/components/VideoUpload/Upload.js
```diff
@@ -0,0 +1,117 @@
...
+  goToAtomPage = () => {
+    const atomId = this.props.video ? this.props.video.id : null;
+
+    if(atomId) {
+      this.props.uploadActions.clearUploadPolicy();
+      browserHistory.push(`/videos/${atomId}`);
+    }
+  };
+
+  header() {
```
clloyd: We have a general pattern of prefixing all rendering methods with `render` so this would be `renderHeader`

clloyd: Also I think having all these small render functions actually makes it harder to read? I'd incorporate this into the main render method

public/video-ui/src/reducers/uploadsReducer.js
```diff
@@ -0,0 +1,21 @@
+export default function upload(state = null, action) {
```
clloyd: This component seems to have two "resting" states... it's `null` before an upload, but `{completed: true}` after an upload. Is that correct? Thinking about state when you upload a file... go out of screen then back in to upload another without refreshing

clloyd: Don't be afraid to have a more complex default state

public/video-ui/src/components/VideoUpload/Upload.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import {browserHistory} from 'react-router';
+import Icon from '../Icon';
+
+class VideoUpload extends React.Component {
```
clloyd: Filename and component name is different, would it make more sense for this to be `VideoUpload.js`

public/video-ui/src/actions/UploadActions/upload.js
```diff
@@ -0,0 +1,38 @@
...
+    type: 'UPLOAD_COMPLETE',
+    receivedAt: Date.now()
+  };
+}
+
+export function startUpload(policy, file) {
+  return dispatch => {
+    dispatch(startUploadAction());
+
+    function progress(completed, total) {
```
clloyd: This is a bit ewww, but not sure at first glance how to do it better...

build_config/webpack.dev.conf.js
mbarton: Yeah nasty. Unfortunately I don't know enough webpack to suggest something better

app/controllers/UploadController.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import com.amazonaws.services.securitytoken.model.AssumeRoleRequest
+import com.gu.pandahmac.HMACAuthActions
+import model.{UploadCredentials, UploadPolicy}
+import play.api.libs.json.{JsArray, JsObject, JsString, Json}
+import play.api.mvc.Results.Ok
+import util.{AWSConfig, Logging}
+
+class UploadController @Inject ()(implicit val authActions: HMACAuthActions, val awsConfig: AWSConfig) extends Logging {
+  val folder = "uploads"
```
akash1810: can this be pushed into the config?

app/controllers/UploadController.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  import authActions.APIHMACAuthAction
+
+  def create(atomId: String) = APIHMACAuthAction {
+    log.info(s"Request for upload credentials for atom $atomId")
+
+    val uploadId = UUID.randomUUID().toString
```
akash1810: rather than `randomUUID`, could we use the MD5 of the file to prevent the same file being uploaded?

akash1810: Talked IRL with the conclusion of: this is tricky!

app/model/Upload.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json._
+
+case class UploadPolicy(bucket: String, key: String, region: String, credentials: UploadCredentials)
+case class UploadCredentials(temporaryAccessKey: String, temporarySecretKey: String, sessionToken: String)
```
akash1810: minor: could it be `temporaryAccessId` to keep inline with the way AWS names it?

app/util/AWS.scala
```diff
@@ -87,6 +100,22 @@ class AWSConfig @Inject() (config: Configuration) {
     null
   )
 
+  private def createUploadSTSClient() = {
+    val provider = stage match {
+      case "DEV" =>
+        // Only required in dev. Instance profile credentials are sufficient when deployed
+        val accessKey = config.getString("aws.upload.accessKey").get
```
akash1810: why aren't janus creds enough?

public/video-ui/src/components/VideoUpload/Upload.js
akash1810: @clloyd we should add this to the linter

public/video-ui/src/services/UploadsApi.js
```diff
@@ -0,0 +1,41 @@
+import {pandaReqwest} from './pandaReqwest';
+
+// See http://andrewhfarmer.com/aws-sdk-with-webpack/ for why this is strange
+import 'aws-sdk/dist/aws-sdk';
+const AWS = window.AWS;
```
akash1810: 😷

app/controllers/UploadController.scala
mbarton: Yep it's tricky but preventing multiple uploads of the same thing and resuming interrupted uploads are definitely goals I'd like to address

public/video-ui/src/services/UploadsApi.js
mbarton: 💩 💩 💩 💩 💩 💩 

Any suggestions greatly appreciated? My Webpack knowledge is not sufficient

app/util/AWS.scala
mbarton: Federated users can't make STS assume role requests unfortunately. I believe this is on the roadmap

public/video-ui/src/components/VideoUpload/Upload.js
mbarton: I'm happy to add it ([keyword-spacing](http://eslint.org/docs/rules/keyword-spacing)) but would do so under a different PR since enabling it would add 27 errors to fix (not in changes I've made).

public/video-ui/src/actions/UploadActions/upload.js
mbarton: Without a full blown stream handling/rx library I think it's ok

public/video-ui/src/components/VideoAssets/VideoAssets.js
mbarton: Nope. I've updated them to use `Link`

public/video-ui/src/components/VideoUpload/Upload.js
mbarton: Renamed

public/video-ui/src/components/VideoUpload/Upload.js
mbarton: I agree. I've collapsed all the little functions into a bigger one

public/video-ui/src/reducers/uploadsReducer.js
mbarton: Ah this was a case of copy-paste ignorance on my part. I've changed the default state to:

```javascript
{ complete: false }
```

app/controllers/UploadController.scala
adamnfish: Yeah, depends how you define "same file". If you're just using filename then md5 is ok, but I think that's unlikely to be what you want.

app/controllers/UploadController.scala
mbarton: Sure

cloudformation/media-atom-maker.json
```diff
@@ -236,6 +241,47 @@
...
+            }
+        },
+        "LimitedUploadRole": {
+            "Type": "AWS::IAM::Role",
+            "Properties": {
+                "AssumeRolePolicyDocument": {
+                    "Statement": [{
+                        "Effect": "Allow",
+                        "Principal": {
+                            "AWS": { "Fn::Join": ["", [ "arn:aws:iam::", { "Ref": "AWS::AccountId" }, ":root"]] }
```
adamnfish: No harm in putting this here, put you only need to do principal delegation when you're going across AWS accounts. Within a single account, normal IAM permissions are enough.

app/util/AWS.scala
```diff
@@ -87,6 +100,27 @@ class AWSConfig @Inject() (config: Configuration) {
     null
   )
 
+  private def createUploadSTSClient() = {
+    val provider = stage match {
+      case "DEV" =>
+        // Only required in dev. Instance profile credentials are sufficient when deployed
+        val accessKey = config.getString("aws.upload.accessKey").getOrElse {
+          throw new IllegalArgumentException("Missing aws.upload.accessKey. This is the AwsId output of the dev cloudformation")
```
adamnfish: Much better with these explicit error messages

app/util/AWS.scala
```diff
@@ -87,6 +100,27 @@ class AWSConfig @Inject() (config: Configuration) {
     null
   )
 
+  private def createUploadSTSClient() = {
+    val provider = stage match {
+      case "DEV" =>
```
adamnfish: Watch this space, AWS have promised they'll be fixing the problems that prevent Janus from issuing credentials that can call assumeRole. It'll definitely be worth changing this when that's possible.

cloudformation/media-atom-maker.json
```diff
@@ -162,6 +162,11 @@
                                     "Effect": "Allow",
                                     "Action": ["sts:AssumeRole"],
                                     "Resource": { "Ref": "ContentAtomArn" }
+                                },
+                                {
+                                    "Effect": "Allow",
```
adamnfish: You could merge this into the other sts:AssumeRole directive by adding this as just another resource in the list of resources. You might consider converting this to yaml in a future PR so you could add comments to describe any pieces that aren't immediately obvious.

Great to keep it tightly scoped like this.

cloudformation/media-atom-maker-dev.yml
```diff
@@ -30,6 +34,7 @@ Resources:
     Type: "AWS::IAM::AccessKey"
     Properties:
       UserName: {"Ref": "MediaAtomUser"}
+      Serial: 2
```
adamnfish: I'm not crazy about putting the access and secret keys into cloudformation output, where they're visible to all. Is this something all the devs share? Be good to have a quick chat IRL about this.

It looks like this only grants very limited access to a dev bucket, so not the end of the world and I understand this is because Janus credentials can't be used in this example.

adamnfish: Good choice to rotate the keys when their permissions change, by the way.

mbarton:
@adamnfish wrt the upload taking longer than 15 minutes, I haven't been able to test because the request times out client-side way before then.

In the next PR I'll be moving to multipart uploading and will test it then

cloudformation/media-atom-maker-dev.yml
mbarton: @adamnfish I agree it's not great and certainly not a pattern to copy anywhere else. I guess we'd need to put the secret key as an output at least otherwise we'd be unable to ever see it?

Happy to chat IRL

mbarton:
@akash1810 I've added a warning to the page

9a395d39-f894-4143-bd47-1faaa91912a6

mbarton:
I ended up including multipart uploading into this PR and, as expected, once the credentials expire the client is unable to upload the parts.

public/video-ui/src/services/UploadsApi.js
clloyd: Do you get the error mentioned in the article if you just import `aws-sdk/clients/s3` and `aws-sdk/libs/credentials`

mbarton: Importing S3 directly:

```javascript
import {AWS} from 'aws-sdk/clients/s3';
```

```
bootstrap 89f1160…:19 Uncaught ReferenceError: require is not defined(…)
```

app/model/Upload.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json._
+
+case class UploadPolicy(bucket: String, key: String, region: String, credentials: UploadCredentials)
+case class UploadCredentials(temporaryAccessId: String, temporarySecretKey: String, sessionToken: String)
+
+object UploadPolicy {
+  implicit val format: Format[UploadPolicy] = Jsonx.formatCaseClass[UploadPolicy]
```
jennysivapalan: Why Jsonx? I googled the import (https://github.com/xdotai/play-json-extensions) and it looks to be only needed for case classes with 22+ parameters which this doesn't have?

cloudformation/media-atom-maker.json
```diff
@@ -236,6 +236,48 @@
...
+                    }]
+                },
+                "Path": "/",
+                "Policies": [
+                    {
+                        "PolicyName": "LimitedUploadPolicy",
+                        "PolicyDocument": {
+                            "Statement": [{
+                                "Effect": "Allow",
+                                "Action": ["s3:PutObject", "s3:PutObjectAcl", "s3:ListBucketMultipartUploads", "s3: ListMultipartUploadParts", "s3:AbortMultipartUpload"],
```
jennysivapalan: Small one but could we put the Action list on new lines to make it easier to read?

build_config/webpack.dev.conf.js
jennysivapalan: I don't know webpack much either. We could do some Tools training if there are enough of us that are a bit unsure of best practises?

clloyd: I'd definitely be up for running that

adamnfish: If you extend it beyond tools I know someone else that'd be interested (spoilers, it's me)

app/model/Upload.scala
mbarton: I copied the existing pattern. Looking at the usages in the code, we could definitely just use `Json.format` built into Play here.

We do however use the `formatCaseClassWithDefaults` method from the library elsewhere. So I guess we then use the library everywhere for consistency?

mbarton:
@adamnfish @clloyd @akash1810 @jennysivapalan 

I think I've addressed all the review comments now apart from the webpack crappiness.

Assuming there's nothing outstanding, anyone good for a +1?

build_config/webpack.dev.conf.js
jennysivapalan: @adamnfish I asked @NataliaLKB about if client side classes will cover this which is planned for later this year. Think Tools team needs something a bit earlier for our use-case. Welcome to join if you like.

app/model/Upload.scala
jennysivapalan: sure for consistency let's keep that.

PR #265, opened 2017-02-13
Publishing expired atoms

Reettaphant:
Expired atoms have a 'private' status and should be published so we should remove the server side check for atom status prior to publishing them. @mbarton @akash1810


Reettaphant:
@akash1810 I think it might be better to just check what is doing the publishing - if an atom has expired, we'd only want to allow the expiry poller to publish the atom (and make the video private), but the client should not be able to publish. I've updated the pr accordingly.

PR #267, opened 2017-02-15
refactor AddAssetCommand

akash1810:
restricting add assets to being youtube videos only and setting version server side

this is part 1 of fixing the add asset form and goes some way to fixing https://github.com/guardian/media-atom-maker/issues/41


akash1810:
This could be done as a for comprehension but it looked horrible as there's an Either...

PR #268, opened 2017-02-16
fix add asset form

akash1810:
Previously, you had to refresh the page after adding an asset because we weren't updating state.

This change also shows the asset list iff there isn't a current asset.

add-asset


akash1810:
cc @blishen

public/video-ui/src/actions/VideoActions/createAsset.js
```diff
@@ -2,14 +2,14 @@ import { browserHistory } from 'react-router';
...
 const BLANK_ASSET = {
-  uri: '',
-  mimetype: '',
-  version: 0
+  uri: ''
 };
 
-function requestAssetCreate() {
+function requestAssetCreate(video) {
+  debugger;
```
mbarton: 😱

public/video-ui/src/components/VideoAssets/VideoAssets.js
```diff
@@ -121,7 +117,7 @@ class VideoAssets extends React.Component {
           {this.renderAssetEdit()}
           {this.renderList()}
           {!this.props.video.assets.length ? <span>No assets found</span> : false}
-          {!this.state.showAssetList && this.props.video.assets.length ? <button className="video-assets__show-btn" type="button" onClick={this.showAssetList}>Show all assets</button> : false}
+          {!this.state.showAssetList && this.props.video.assets.length && !!this.props.video.activeVersion ? <button className="video-assets__show-btn" type="button" onClick={this.showAssetList}>Show all assets</button> : false}
```
mbarton: I think this would be clearer refactored out of a one-liner?

akash1810:
@mbarton good to go?

PR #269, opened 2017-02-17
use a filter to redirect

akash1810:
We were previously using a Google Auth error as a means to redirect which is flaky. Now, we intercept the request early in a filter.


jennysivapalan:
👍 🗃

PR #271, opened 2017-02-20
Remove over-chatty alarm

mbarton:
The 4XX alarm was not useful as it was constantly firing due to vulnerability scanning.

This PR removes it.


mbarton:
@akash1810 true. I tried that a few times and each time the security scanner started sending more requests!

PR #272, opened 2017-02-20
Add support for AWS lambda functions

mbarton:
This PR splits the code into multiple sub-projects, allowing us to write AWS lambda functions that share code with the main app. (see the `common` project).

The example `UploaderLambda` uses shared code to download and parse configuration, make a YouTube client and echo back some results.

NB reviewers:

- Logging in the lambda is done through cloudwatch. This is because the kinesis appender we use elsewhere was keeping the application alive even after it terminated, which *may* end up holding a lambda instance until it times out. I'll investigate and fix in subsequent PRs.

- There is a funky line in the `build.sbt` which works around a lack of support for the root project not being in the root folder. As a result, we get three logging lines setting the project when you start SBT rather than one.

lambda


app/util/LogShipping.scala
```diff
@@ -2,54 +2,30 @@ package util
...
-import net.logstash.logback.layout.LogstashLayout
+import com.amazonaws.regions.Region
+import com.gu.media.CrossAccountAccess
+import com.gu.media.logging.KinesisLogging
+import com.typesafe.config.Config
 import org.slf4j.{LoggerFactory, Logger => SLFLogger}
-import play.api.Logger
-
 
 trait LogShipping {
```
jennysivapalan: Trait can be removed?

build.sbt
```diff
@@ -70,13 +41,39 @@ lazy val root = (project in file("."))
...
+  .aggregate(common, app, uploader)
+  .enablePlugins(RiffRaffArtifact)
+  .settings(
+    riffRaffBuildIdentifier := Option(System.getenv("CIRCLE_BUILD_NUM")).getOrElse("dev"),
+    riffRaffManifestBranch := Option(System.getenv("CIRCLE_BRANCH")).getOrElse("dev"),
+    riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
+    riffRaffUploadManifestBucket := Option("riffraff-builds"),
+    riffRaffManifestProjectName := "media-service:media-atom-maker",
+    riffRaffArtifactResources := Seq(
+      (packageBin in Universal in app).value -> s"${(name in app).value}/${(packageBin in Universal in app).value.getName}",
```
jennysivapalan: The `app` is a deb file so think it should be 
```
(packageBin in Debian in app).value -> s"${(name in app).value}/${(packageBin in Debian in app).value.getName}"
```

Though it might be easier to  create variables for `(packageBin in Universal in app)` and `(packageBin in Universal in uploader)` and then use them the couple times they are referred to?

uploader/src/main/scala/com/gu/media/UploaderLambda.scala
```diff
@@ -0,0 +1,31 @@
...
+  case class Input(string: String, integer: Int)
+  case class Output(result: String)
+
+  implicit val inputFormat: Format[Input] = Jsonx.formatCaseClass[Input]
+  implicit val outputFormat: Format[Output] = Jsonx.formatCaseClass[Output]
+}
+
+object UploaderLambdaApp extends App {
+  val lambda = new UploaderLambda
+  println(lambda.handleRequest(Input("hello world!", 4)))
```
jennysivapalan: Is this request to get it to trigger something on a new lambda update or was needed for dev testing so can be removed?

uploader/src/main/scala/com/gu/media/UploaderLambda.scala
```diff
@@ -0,0 +1,31 @@
...
+import com.gu.media.UploaderLambda.{Input, Output}
+import com.gu.media.lambda.LambdaWithParams
+import com.gu.media.logging.Logging
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json.Format
+
+class UploaderLambda extends LambdaWithParams[Input, Output] with Logging {
+  override def handleRequest(input: Input): Output = {
+    val youTube = YouTubeClient(config)
+    val output = s"allowedChannels: ${youTube.allowedChannels.mkString(",")}. result: ${input.string * input.integer}"
```
jennysivapalan: What's `input` and `input.integer` for?

mbarton: Just for testing the skeleton. They will be replaced with the case classes for input/output params (as JSON)

app/util/LogShipping.scala
mbarton: Done

build.sbt
mbarton: Ooops yep I've changed it. I can't work out a way to refactor out the code here without triggering a "dynamic reference error" from SBT (horrible macro error)

uploader/src/main/scala/com/gu/media/UploaderLambda.scala
mbarton: Removed

cloudformation/media-atom-maker.json
```diff
@@ -78,9 +86,20 @@
                 "DesiredCapacity": 1,
                 "InstanceType": "t2.small"
             }
+        },
+        "LambdaBuilds": {
+            "Uploader": {
+                "CODE": "media-service/CODE/media-atom-uploader/media-atom-uploader.zip",
```
akash1810: minor: we could do an inline `Fn::Join` and avoid a mapping as its not providing much benefit here

common/src/main/scala/com/gu/media/logging/LambdaElkLoggingFormat.scala
```diff
@@ -0,0 +1,39 @@
...
+import ch.qos.logback.classic.{Logger => LogbackLogger}
+import ch.qos.logback.core.ConsoleAppender
+import ch.qos.logback.core.encoder.LayoutWrappingEncoder
+import org.slf4j.{LoggerFactory, Logger => SLFLogger}
+
+trait LambdaElkLoggingFormat {
+  private val rootLogger = LoggerFactory.getLogger(SLFLogger.ROOT_LOGGER_NAME).asInstanceOf[LogbackLogger]
+
+  (sys.env.get("STACK"), sys.env.get("APP"), sys.env.get("STAGE")) match {
+    case (Some(stack), Some(app), Some(stage)) =>
```
akash1810: nice!

PR #274, opened 2017-02-20
Add PROD lambda to riff raff config

mbarton:
Update riff raff config with the prod instance of the uploader lambda


jennysivapalan:
👍

PR #279, opened 2017-02-21
refactor usage

akash1810:
As Composer is the only place you can embed atoms, we can get away with making a single capi call to get usages and then use the `find-by-path` endpoint of Composer to link to the article.

-104 +29 😄

### DIFF
--- a/public/video-ui/src/actions/VideoActions/videoUsages.js
+++ b/public/video-ui/src/actions/VideoActions/videoUsages.js
@@ -7,13 +7,10 @@ function requestVideoUsages() {
   };
 }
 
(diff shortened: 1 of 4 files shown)


mbarton:
Good stuff. I'm happy with this providing the UI updates correctly: when you press "Create Video Page" it appears in the list afterwards

akash1810:
> providing the UI updates correctly: when you press "Create Video Page" it appears in the list afterwards

Yep it does - https://github.com/guardian/media-atom-maker/pull/279/files#diff-d2d4cfd580face7d138b5d162c26103eR7

PR #280, opened 2017-02-21
Switch to manual compile time dependency injection

mbarton:
Last of the massive PRs I promise!

- Switch to manual compile time dependency injection (see `di.scala`)
- Move out AWS configuration and access to separate traits (see `com.gu.media.aws` in common)
- Move out YouTube access to separate traits (see `com.gu.media.youtube` in common)
- Add `DataStores` class to reduce number of parameters passed into controllers


akash1810:
This is a little hard to constructively review due to its length 😢. Can we split it up at all? Maybe into the four points listed in the description?

mbarton:
Yeah it's big, sorry about that. Unfortunately, it's a balance now between splitting out the commit (which takes development time) and reviewing it (also takes dev time, although it's not mine of course!).

I would prefer not to split it out. What do you guys think @jennysivapalan @Reettaphant?

jennysivapalan:
I agree with @akash1810. I think if this PR could be four separate commits so that each commit can be read individually or 4 separate PRs that would be easier to review.

mbarton:
Ok I'll split it up

PR #281, opened 2017-02-21
update usage to have links to Composer and Frontend

akash1810:
Previously, we were just linking to Composer. Now we have icons linking to Frontend and Composer.

Also, show creation date relative to now.

image

Before Deploying
- [x] update configuration with url to Viewer

### DIFF
--- a/public/video-ui/src/actions/VideoActions/getVideos.js
+++ b/public/video-ui/src/actions/VideoActions/getVideos.js
@@ -1,5 +1,5 @@
 import VideosApi from '../../services/VideosApi';
-import {searchText} from '../../services/capi';
+import ContentApi from '../../services/capi';
 
 function requestVideos() {
   return {
@@ -66,12 +66,10 @@ function adaptCapiAtom(atom) {
 }
 
 export function searchVideosWithQuery(query) {
-  const encodedQuery = encodeURIComponent(query);
-
   return dispatch => {
     dispatch(requestSearchVideos());
 
-    return searchText(encodedQuery)
+    return ContentApi.search(query)
       .then(res => {
         const capiAtoms = res.response.results;
         const atoms = capiAtoms.map(adaptCapiAtom);
--- a/public/video-ui/src/components/Icon.js
+++ b/public/video-ui/src/components/Icon.js
@@ -24,6 +24,21 @@ export class ComposerIcon extends React.Component {
   }
 }
 
+export class ViewerIcon extends React.Component {
+  render () {
+    return (
+      <span className="icon svg-icon">
+        <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32">
+          <g fill="#FFF">
+            <path d="M16 5.2C5.7 5.2 0 16 0 16s5.7 10.8 16 10.8S32 16 32 16 26.3 5.2 16 5.2zm0 18.4c-4.4 0-8-3.6-8-8s3.6-8 8-8 8 3.6 8 8-3.6 8-8 8z"/>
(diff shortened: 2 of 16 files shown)


akash1810:
cc @mr-mr

public/video-ui/src/components/Icon.js
```diff
@@ -1,5 +1,29 @@
 import React from 'react';
 
+export class FrontendIcon extends React.Component {
+  render () {
+    return (
+      <span className="icon svg-icon">
+        <svg xmlns="http://www.w3.org/2000/svg" width="128" height="128" viewBox="0 0 128 128">
```
akash1810: stolen from https://github.com/guardian/grid/tree/master/kahuna/public/js/components/gr-icon/icons 😉

public/video-ui/src/actions/VideoActions/getVideos.js
```diff
@@ -66,12 +66,10 @@ function adaptCapiAtom(atom) {
 }
 
 export function searchVideosWithQuery(query) {
-  const encodedQuery = encodeURIComponent(query);
```
akash1810: this has been moved into `capi.js`, closer to where its needed

jennysivapalan:
Should we link to viewer (preview) too? Then the user can check mobile and apps too if they wish. Though

akash1810:
@jennysivapalan great idea! Will add it.

akash1810:
> Should we link to viewer (preview) too? 

Added.

image

public/video-ui/src/components/Icon.js
```diff
@@ -1,5 +1,44 @@
...
+      </span>
+    );
+  }
+}
+
+export class ViewerIcon extends React.Component {
+  render () {
+    return (
+      <span className="icon svg-icon">
+        <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32">
```
akash1810: copied from https://github.com/guardian/editorial-viewer/blob/master/public/images/eye-white-32.svg

akash1810:
This breaks the `create new page` button... blocking merging for now.

akash1810:
> This breaks the create new page button

Fixed by polling CAPI up to 10 seconds because it takes a little time for new Composer pages to reach it.

polling

public/video-ui/src/components/VideoUsages/VideoUsages.js
```diff
@@ -71,11 +73,28 @@ export default class VideoUsages extends React.Component {
   };
 
   renderUsage = (usage) => {
-    const composerLink = `${this.getComposerUrl()}/find-by-path/${usage}`;
+    const composerLink = `${this.getComposerUrl()}/content/${usage.fields.internalComposerCode}`;
+    const viewerLink = `${this.getViewerUrl()}/preview/${usage.id}`;
+    const websiteLink = `https://gu.com/${usage.id}`;
```
jennysivapalan: I would use `https://www.guardian.com` and save a redirect!

akash1810: ah, yep fair point. I was being lazy 😄 . Updated.

public/video-ui/styles/abstracts/_variables.scss
```diff
@@ -89,3 +89,4 @@ $anchorColor: $cWhite;
 $toolbarHeight: 50px;
 $videoBlockWidth: 230px;
 $standard-padding: 10px;
+$iconSize: 16px; /* Preferred icon size */
```
mbarton: 👍 I was looking for this in my branch

akash1810:
> I'm on the fence about polling CAPI rather than updating state locally immediately then merging it with the real state once updated.
I guess in reality if the page doesn't appear in CAPI, that will appear as a problem for the user (won't render in preview) so it's worth the poll.

`VideoUsage` renders properties from a capi response (e.g. `usage.fields.internalComposerCode`). As the capi object is complex, It looked a bit messy to have a create and use a fake capi object then replace it with the real capi object.

Polling seemed like a lesser evil.

PR #282, opened 2017-02-21
Compile-time DI && common AWS and YouTube access

### LINKED ISSUE #24: Use static dependency injection, e.g. MacWire
As it is really annoying to have it all compile and then barf during injection at run time.

https://github.com/adamw/macwire

mbarton:
Re-run of #280 in easier (sort of) to digest chunks. It's recommended to review each commit of this PR separately.

* **Manual compile-time dependency injection**
    * Remove `@Inject` attributes and convert all implicit parameters to real ones
    * RequestLogging has not been converted as it is loaded by Play itself
    * Pull up `ExpiryPoller.start` to run in `di.scala`, not `Api2`
* **Add DataStores class to reduce number of parameters**
    * `HasDataStores` trait means we don’t have to rename references everywhere
* **Split AWSConfig into traits in common**
    * Relies on pure Typesafe Config rather than Play Configuration
    * DynamoAccess has been moved now even though it will actually be used in later PRs for the upload API
* **Split YouTube into common traits**
    * Expiring an atom stays in the main app since it relies on our `MediaAtom` domain model
    * Add a `YouTubeMetadataUpdate` class to allows us to update the metadata in common code without bringing across the whole data model
    * Bump `play-json-extensions` version to fix library eviction problem when moving the `YouTubeChannel` reads and writes to common

Fixes #24


app/controllers/Api2.scala
```diff
@@ -1,49 +1,37 @@
 package controllers
 
-import javax.inject.Inject
-
-import _root_.util.{AWSConfig, ExpiryPoller, YouTubeConfig, YouTubeVideoUpdateApi}
-import akka.actor.ActorSystem
+import _root_.util.{AWSConfig, YouTubeConfig, YouTubeVideoUpdateApi}
```
akash1810: the `_root_` of all evil

app/util/ExpiryPoller.scala
```diff
@@ -43,5 +49,31 @@ case class ExpiryPoller(override val stores: DataStores, youtubeConfig: YouTubeC
       })
     })
   }
+
+  def updateStatusIfExpired(thriftAtom: Atom): Option[MediaAtom] = {
```
jennysivapalan: Not sure if it is overkill but this could be a unit tested?

app/util/Youtube.scala
```diff
@@ -1,233 +0,0 @@
...
-      .setRegionCode("GB")
-
-    request.execute.getItems.asScala.toList
-      .filter(_.getSnippet.getAssignable)
-      .map(YouTubeVideoCategory.build)
-      .sortBy(_.title)
-  }
-}
-
-case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends Logging {
```
jennysivapalan: this is quite a big diff change - I'm going to assume everything has been moved to one of the traits!

app/controllers/Api2.scala
mbarton: The true evil is the root package

app/util/ExpiryPoller.scala
mbarton: Sure. I'll do that when I move the expiry poller to run in a lambda

app/data/DataStores.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  private def createPreview() = {
+    new PreviewDynamoDataStore[MediaAtom](aws.dynamoDB, aws.dynamoTableName) {
+      def fromAtomData = { case AtomData.Media(data) => data }
+      def toAtomData(data: MediaAtom) = AtomData.Media(data)
+    }
+  }
+}
+
+trait HasDataStores {
```
akash1810: slightly odd naming

common/src/main/scala/com/gu/media/Settings.scala
```diff
@@ -0,0 +1,16 @@
...
+import scala.collection.JavaConverters._
+
+trait Settings {
+  val config: Config
+
+  def getString(name: String): Option[String] = if(config.hasPath(name)) { Some(config.getString(name)) } else { None }
+  def getStringList(name: String): List[String] = if(config.hasPath(name)) { config.getStringList(name).asScala.toList } else { List.empty}
+  def getBoolean(name: String): Option[Boolean] = if(config.hasPath(name)) { Some(config.getBoolean(name)) } else { None }
+
+  def getMandatoryString(name: String, hint: String = ""): String = Option(config.getString(name)).getOrElse {
```
akash1810: could probably get away with removing `hint`. Its only used to point towards where the dev config keys come from, but as its a shared config, one would hope new keys would be added to the shared file before merging anyway.

common/src/main/scala/com/gu/media/aws/CrossAccountAccess.scala
```diff
@@ -0,0 +1,22 @@
...
+trait CrossAccountAccess { this: AwsAccess =>
+  def getCrossAccountCredentials(sessionId: String): AWSCredentialsProvider = {
+    new AWSCredentialsProviderChain(
+      new ProfileCredentialsProvider("composer"),
+      new STSAssumeRoleSessionCredentialsProvider(credsProvider, getStsRole(config), sessionId)
+    )
+  }
+
+  private def getStsRole(config: Config): String = {
+    if(config.hasPath("aws.kinesis.stsRoleToAssume")) {
```
akash1810: `config` already does this... simpler to do `config.getString("aws.kinesis.stsRoleToAssume").getOrElse("")`?

mbarton: This is a raw typesafe config rather than a play config. If I don't do the check before it throws an exception

common/src/main/scala/com/gu/media/Settings.scala
mbarton: I know what you mean, but it was added in response to the review from original uploading PR

app/data/DataStores.scala
mbarton: Yeah it's crap. I'll change it

PR #285, opened 2017-02-21
Only respect disallowedVideos and allowedChannels if non-empty

mbarton:
A production outage (inability to publish) was caused by an error in reading the `youtube.allowedChannels` setting.

Prior to the changes in #272, the setting was only respected if it was present in the config file. That PR changed the code to read the setting, meaning that if the setting was missing the result was an empty list.

Before: `Option[List[String]]`
After: `List[String]`

As such, we need to treat the empty list as "no instruction" and not refuse any operations.

### DIFF
--- a/app/util/Youtube.scala
+++ b/app/util/Youtube.scala
@@ -39,7 +39,7 @@ case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends Logging {
   private def protectAgainstMistakesInDev(video: Video) = {
     val videoChannelId = video.getSnippet.getChannelId
 
-    if (config.disallowedVideos.nonEmpty && config.disallowedVideos.contains(video.getId)) {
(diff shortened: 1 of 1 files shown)


app/util/Youtube.scala
```diff
@@ -39,13 +39,13 @@ case class YouTubeVideoUpdateApi(config: YouTubeConfig) extends Logging {
   private def protectAgainstMistakesInDev(video: Video) = {
     val videoChannelId = video.getSnippet.getChannelId
 
-    if (config.disallowedVideos.contains(video.getId)) {
+    if (config.disallowedVideos.nonEmpty && config.disallowedVideos.contains(video.getId)) {
```
akash1810: is the `nonEmpty` check needed here?

mbarton: Nope. I've removed it

PR #286, opened 2017-02-22
Saving workflow

Reettaphant:
- Saves the metadata and youtube metadata forms only when the form is closed so that we can validate form fields. 

- Using the form validate function in `VideoDisplay.js` to determine when the buttons that close the forms are disabled is not ideal but checking the form validity to determine this would require writing multiple validate functions instead of having one that checks all video fields. If we have time, we should probably spend some time thinking about how we do form validation and perhaps get rid of redux forms.

- The 'create button is disabled in the create form if the atom is not valid

- Only one form can be edited at a time. If edit metadata or edit youtube forms are open, other atom data, including the poster image, cannot be updated: 
screen shot 2017-02-24 at 09 45 21

@mbarton @akash1810 @jennysivapalan

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -32,24 +32,23 @@ class VideoDisplay extends React.Component {
     this.props.videoActions.debouncedSaveVideo(video);
   };
 
+  updateVideo = (video) => {
+    this.props.videoActions.updateVideo(video);
+  };
+
   selectVideo = () => {
     window.parent.postMessage({atomId: this.props.video.id}, '*');
   };
 
-  manageEditingState = (value, property) => {
+  manageEditingState = (property) => {
 
-    if (property === 'metadata') {
+    if (this.props.editState[property]) {
 
-      this.setState({
-        metadataEditable: value
-      });
-
-    } else if (property === 'youtube') {
-
-      this.setState({
-        youtubeEditable: value
-      });
+      this.saveAndUpdateVideo(this.props.video);
     }
+
+    const newEditableState = Object.assign(this.props.editState, {[property]: !this.props.editState[property]});
+    this.props.videoActions.updateVideoEditState(newEditableState);
   };
 
   disableEditing = () => {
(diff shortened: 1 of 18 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -62,15 +62,41 @@ class VideoDisplay extends React.Component {
...
+      formFields = ['title', 'category'];
+    } else if (property === 'youtubeEditable') {
+      formFields = ['youtubeCategory', 'youtubeChannel', 'privacyStatus'];
+    }
+    const errors  = validate(Object.assign(this.props.video, {
+      youtubeCategory: this.props.video.youtubeCategoryId,
+      youtubeChannel: this.props.video.channelId
+    }));
+
+    if (formFields.some(field => {
```
akash1810: if `property` != `'metadataEditable'` or  `'youtubeEditable'` it'll be `undefined` and this line will fail.

Could we instantiate `formFields` as `[]`? It also has the side-effect of becoming a `const` too.

public/video-ui/src/components/Video/Display.js
```diff
@@ -62,15 +62,41 @@ class VideoDisplay extends React.Component {
...
+
+  };
+
+  renderEditButton = (property) => {
 
-    if (editable) {
+    if (this.props && this.props.editState[property]) {
       return (
-        <Icon className="icon__done" icon="done" onClick={this.manageEditingState.bind(this, false, property)}/>
+        <button disabled={this.isButtonDisabled(property)} onClick={this.manageEditingState.bind(this, property)}>
```
akash1810: `.bind` is a little scary...

public/video-ui/src/components/VideoMetaData/VideoMetaData.js
```diff
@@ -8,41 +8,40 @@ import VideoCategorySelect from '../VideoEdit/formComponents/VideoCategory';
 import { Field, reduxForm } from 'redux-form';
 import {validate} from '../../constants/videoEditValidation';
 
-
 const VideoMetaData = (props) => {
 
     return (
-        <div className="form__group">
+        <form className="form__group">
```
akash1810: why was this change needed?

Reettaphant: It wasn't in the end! I'll change it back.

Reettaphant:
@akash1810 I've now updated the styles of the disabled buttons after a chat with @mr-mr , and I also added validation to the create form to disable the create form submission when the atom is invalid to avoid errors there.

public/video-ui/src/components/Video/Display.js
```diff
@@ -8,69 +8,89 @@ import YoutubeMetaData from '../YoutubeMetaData/YoutubeMetaData';
...
+
+    return false;
+  }
+
+
+  cannotCloseEditForm = (property) => {
+
+    const formFields = [];
+    if (property === 'metadataEditable') {
+      formFields.splice(0, 0, 'title', 'category');
```
akash1810: could use `.concat`

akash1810: `.concat` was bad advice as it returns a new array. The spread operator is the right way to go `formFields.push(...['title', 'category'])`.

PR #287, opened 2017-02-22
render picker only if video loaded

Reettaphant:
@jennysivapalan @akash1810

### DIFF
--- a/public/video-ui/src/components/VideoEdit/formComponents/VideoPoster.js
+++ b/public/video-ui/src/components/VideoEdit/formComponents/VideoPoster.js
@@ -19,7 +19,7 @@ class VideoPosterImageEdit extends React.Component {
 
   render () {
 
-    if (!this.props.editMode) {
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/VideoEdit/formComponents/VideoPoster.js
```diff
@@ -19,7 +19,7 @@ class VideoPosterImageEdit extends React.Component {
 
   render () {
 
-    if (!this.props.editMode) {
+    if (this.props.video && !this.props.editMode) {
```
akash1810: [PropTypes](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) will help here as we can enforce `this.props.video` is required.

Reettaphant: I think it will always be null when the page is loading so its not really invalid data?

PR #288, opened 2017-02-23
Transcodes a master video to mp4

jennysivapalan:
There are two parts in this PR. If it's too long I can separate them into two PRs.
* Lambda (and supporting infrastructure / deployment) - when given a pipeline ID and a master file name will transcode a master file to an mp4. The pipeline is aware of the input and output buckets.
* An API endpoint in the app that for a pipeline ID (set in config) will render the status of each job in a pipeline. 

Assumptions
* One pipeline per stage environment. The lambda can accept any pipeline ID but the app assumes just one pipeline ID is set for now. This can be extended later if need multiple pipelines.
* The API endpoint just gets one page of results from AWS and is ordered by most recent. Hopefully we won't need to transcode and host videos often.
* I've guessed the format to transcode to mp4s, this may need to change once we test this in production.

Before merging
[ ] Update AWS config in S3

Future PRs that need to be raised to complete the no youtube option:
* Trigger a transcode from the app (currently this is just done by using the Test option on the Lambda)
* Add Fastly config on the mp4 S3 bucket.


app/controllers/Transcoder.scala
```diff
@@ -0,0 +1,18 @@
+package controllers
+
+import javax.inject._
+
+import com.gu.pandahmac.HMACAuthActions
+import play.api.libs.json.Json
+import play.api.mvc.Controller
+
+class Transcoder @Inject()(val authActions: HMACAuthActions, transcoder: util.Transcoder) extends Controller {
```
mbarton: We can remove the `@Inject` now

app/util/Transcoder.scala
```diff
@@ -0,0 +1,40 @@
...
+import akka.agent.Agent
+import com.amazonaws.services.elastictranscoder.model.ListJobsByPipelineRequest
+import model.transcoder.JobStatus
+
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+class Transcoder @Inject()(awsConfig: AWSConfig) {
```
mbarton: As above

cloudformation/media-atom-maker-dev.yml
```diff
@@ -198,15 +208,58 @@ Resources:
...
+            Statement:
+              - Effect: "Allow"
+                Action: ["elastictranscoder:CreateJob"]
+                Resource: ["arn:aws:elastictranscoder:*"]
+
+  TranscoderLambda:
+    Type: "AWS::Lambda::Function"
+    Properties:
+      Description: "Transcodes masters to mp4s"
+      Handler: "com.gu.media.Lambda::handleRequest"
```
mbarton: Maybe give this class a less generic name?

cloudformation/media-atom-maker.json
```diff
@@ -381,6 +388,73 @@
...
+                }]
+          }
+        },
+
+        "TranscoderLambda": {
+          "Type": "AWS::Lambda::Function",
+          "Properties": {
+            "Description": "Transcodes masters to mp4s",
+            "FunctionName":  {"Fn::Join": ["-", ["media-atom-maker-transcoder", {"Ref": "Stage"} ] ] },
+            "Handler": "com.gu.media.Lambda::handleRequest",
```
mbarton: As above

app/util/Transcoder.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  val agent: Agent[List[JobStatus]] = Agent(List.empty)
```
mbarton: Could we just request the pipeline state on-demand, rather than caching the results?

jennysivapalan:
As chatted with @mbarton I will change the Akka agent to using Play's CacheApi.

app/util/Transcoder.scala
jennysivapalan: As chatted I think having a small cache is useful for not hitting AWS so much. I'll reduce the cache time to 20 seconds.

jennysivapalan:
@mbarton have addressed your suggestions if you're okay to 👍 ?

app/util/Transcoder.scala
```diff
@@ -0,0 +1,28 @@
...
+import play.api.cache.CacheApi
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
+class Transcoder(awsConfig: AWSConfig, cache: CacheApi) {
+
+  def getJobsStatus = cache.getOrElse[List[JobStatus]]("transcoderJobs", 20.seconds){updateJobsStatus}
+
+  private def updateJobsStatus = {
```
mbarton: I'd probably give this function brackets in the definition since it does a whole request to a back-end. Not a dealbreaker

project/Dependencies.scala
```diff
@@ -30,6 +30,7 @@ object Dependencies {
   val awsS3 = "com.amazonaws" % "aws-java-sdk-s3" % awsVersion
   val awsDynamo = "com.amazonaws" % "aws-java-sdk-dynamodb" % awsVersion
   val awsLambdaCore = "com.amazonaws" % "aws-lambda-java-core" % "1.1.0"
+  val awsTranscoder = "com.amazonaws" % "aws-java-sdk-elastictranscoder" % awsVersion
```
akash1810: not for this PR, but why is `awsLambdaCore` frozen at `1.1.0` rather than `awsVersion`?

transcoder/src/main/scala/com/gu/media/TranscoderLambda.scala
```diff
@@ -0,0 +1,56 @@
...
+    if (jobInfo.masterFileName.endsWith("mxf")) {
+      val transcoderClient = new AmazonElasticTranscoderClient()
+      transcoderClient.setRegion(Region.getRegion(Regions.EU_WEST_1))
+
+      val jobInput = new JobInput().withKey(jobInfo.masterFileName)
+
+      val mp4FileName = jobInfo.masterFileName.replace(".mxf", ".mp4")
+
+      val jobOutput = new CreateJobOutput()
+        .withKey(s"$mp4FileName")
```
akash1810: why are we string interpolating here?

jennysivapalan: Good point - a hangover from a refactor I made. I'll remove it.

project/Dependencies.scala
mbarton: I believe the SDK for building lambdas operates on a different schedule to the AWS clients

PR #291, opened 2017-02-24
Grid tweaks and Usages

mr-mr:
Mostly the zero states of each of the blocks, also includes a clean up of the usages header and button

Zero state on preview, usages and assets
image

Usage button updated to secondary button and lives in the block header
image

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -7,6 +7,8 @@ import VideoMetaData from '../VideoMetaData/VideoMetaData';
 import YoutubeMetaData from '../YoutubeMetaData/YoutubeMetaData';
 import VideoPoster from '../VideoPoster/VideoPoster';
 import GridImageSelect from '../utils/GridImageSelect';
+import {getVideoBlock} from '../../util/getVideoBlock';
+import {getStore} from '../../util/storeAccessor';
 import Icon from '../Icon';
 
 class VideoDisplay extends React.Component {
@@ -18,6 +20,7 @@ class VideoDisplay extends React.Component {
 
   componentWillMount() {
     this.props.videoActions.getVideo(this.props.params.id);
+    this.props.videoActions.getUsages(this.props.params.id);
   }
 
   saveVideo = () => {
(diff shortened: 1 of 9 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -62,6 +66,37 @@ class VideoDisplay extends React.Component {
...
+
+    return this.props.videoActions.createVideoPage(this.props.video.id, metadata, this.getComposerUrl(), videoBlock);
+  }
+
+  getComposerUrl = () => {
+    return getStore().getState().config.composerUrl;
+  }
+
+  renderCreateButton = () => {
+      const filterUsageType = this.props.usages.filter(value => value.type === 'video');
```
akash1810: would `videoPageUsages` be a more descriptive variable name?

public/video-ui/src/components/VideoUsages/VideoUsages.js
```diff
@@ -56,21 +40,21 @@ export default class VideoUsages extends React.Component {
...
-          disabled={this.state.pageCreated}
-          onClick={this.pageCreate}>
-          Create video page
-        </button>
-      );
-    }
-
-    return (<div>Publish this atom to enable the creation of composer pages</div>);
-  };
+  // renderCreateButton = () => {
```
akash1810: can remove commented out code

public/video-ui/src/components/VideoAssets/VideoAssets.js
```diff
@@ -110,14 +110,14 @@ class VideoAssets extends React.Component {
   renderHeader() {
     const buttons = <div className="video-assets__buttons">
       <a className="button" onClick={this.showAssetForm}>
-        <Icon className="icon__edit" icon="add"/>
+        <Icon className="icon__edit icon__spacing" icon="add"/>
```
akash1810: `icon__spacing` is adding a `margin-right` but there's no content to the right of this `Icon`. Would be better to add the spacing to the `a` above instead.

mr-mr: The other content is the upload button.

akash1810: yep, but the upload button is a sibling of the `a` rather than the `Icon`

it is a fairly minor point tbh, but makes the css simpler imo

public/video-ui/src/components/Video/Display.js
akash1810: meep, I meant for `videoPageUsages` would be more descriptive for `filterUsageType` rather than `renderCreateButton`

PR #293, opened 2017-02-27
remove redirect on google auth error

akash1810:
This was put in place for graceful fallback when users went to `media-atom-maker.` rather than `video.`.
I tried using a Filter to do this redirection earlier in #269 but it broke the healthcheck.

Let's just delete the redirection entirely and remove the old domain from DNS as no-one is using it anymore anyway.

Before Deploy
- [ ] Ask CP to send comms


mbarton:
👍 👍

akash1810:
delaying merging until Wednesday

PR #297, opened 2017-02-27
Use latest atom-publisher-lib

mbarton:
Upgrade to use the latest `atom-publisher-lib`. This is so we can also use the latest `scanamo` which supports consistent reads/writes.

Sorry for the diff spam. The actual code in each action should not have changed, only the access to the datastores. The exception is `AddAssetCommand`, where I have refactored the main body of the code into a method.

- ~~When updating the `Api` controller, I have swallowed all errors as `AtomNotFound`. Is there a better way?~~
- I've added helper methods to avoid repeating `this atom does not exist` code in controllers and commands

### DIFF
--- a/app/controllers/Api.scala
+++ b/app/controllers/Api.scala
@@ -34,15 +34,15 @@ class Api (override val stores: DataStores,
 
   def getMediaAtom(id: String) = APIAuthAction { implicit req =>
     previewDataStore.getAtom(id) match {
-      case Some(atom) => Ok(Json.toJson(atom))
-      case None => NotFound(jsonError(s"no atom with id $id found"))
+      case Right(atom) => Ok(Json.toJson(atom))
+      case _ => NotFound(jsonError(s"no atom with id $id found"))
     }
   }
 
   def getPublishedMediaAtom(id: String) = APIAuthAction { implicit req =>
     publishedDataStore.getAtom(id) match {
-      case Some(atom) => Ok(Json.toJson(atom))
-      case None => Ok("not published")
+      case Right(atom) => Ok(Json.toJson(atom))
+      case _ => Ok("not published")
     }
   }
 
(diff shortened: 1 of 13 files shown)


app/controllers/Api.scala
```diff
@@ -33,15 +33,15 @@ class Api (override val stores: DataStores,
 
   def getMediaAtom(id: String) = APIAuthAction { implicit req =>
     previewDataStore.getAtom(id) match {
-      case Some(atom) => Ok(Json.toJson(atom))
-      case None => NotFound(jsonError(s"no atom with id $id found"))
+      case Right(atom) => Ok(Json.toJson(atom))
+      case _ => NotFound(jsonError(s"no atom with id $id found"))
```
akash1810: The only other case is `Left`... shall we be more explicit?

app/data/DataStores.scala
```diff
@@ -25,17 +35,14 @@ class DataStores(aws: AWSConfig) extends MediaAtomImplicits {
...
-      def fromAtomData = { case AtomData.Media(data) => data }
-      def toAtomData(data: MediaAtom) = AtomData.Media(data)
-    }
-  }
-
-  private def createPreview() = {
-    new PreviewDynamoDataStore[MediaAtom](aws.dynamoDB, aws.dynamoTableName) {
-      def fromAtomData = { case AtomData.Media(data) => data }
-      def toAtomData(data: MediaAtom) = AtomData.Media(data)
+  def getDataStores[T: ClassTag: DynamoFormat](dynamoFormats: AtomDynamoFormats[T]): (PreviewDynamoDataStore[T], PublishedDynamoDataStore[T]) = {
```
akash1810: any particular reason to return a tuple?

app/controllers/Api.scala
mbarton: Definitely, I was unsure what to return. Probably an InternalServerError like Api2 I guess?

app/data/DataStores.scala
mbarton: Nope. It could easily be two methods

mbarton:
@akash1810 yes absolutely

PR #298, opened 2017-02-28
S3 and DynamoDB triggered lambdas

mbarton:
In preparation for adding the API and logic for tracking video uploads, this PR updates the cloudformation:

- Adds the Dynamo table that will back the API and process
- Split the uploader Lambda into two:
  - triggered by S3 events
  - triggered by Dynamo events

The lambdas themselves do not currently do anything other than parrot back the events they receive. Further PRs will implement the actual functionality.


build.sbt
```diff
@@ -85,7 +85,9 @@ lazy val root = (project in file("root"))
     riffRaffManifestProjectName := "media-service:media-atom-maker",
     riffRaffArtifactResources := Seq(
       (packageBin in Debian in app).value -> s"${(name in app).value}/${(name in app).value}.deb",
-      (packageBin in Universal in uploader).value -> s"${(name in uploader).value}/${(packageBin in Universal in uploader).value.getName}",
+      // we have an entry here for each lambda that uses this code (see the cloud formation)
+      (packageBin in Universal in uploader).value -> s"media-atom-uploader-s3-events/${(packageBin in Universal in uploader).value.getName}",
```
mbarton: Clunky but allows us to share a single build between the lambdas. 99% of the code will be shared, with only the entry points differing

cloudformation/media-atom-maker-dev.yml
```diff
@@ -115,9 +115,30 @@ Resources:
...
+      ProvisionedThroughput:
+        ReadCapacityUnits: "5"
+        WriteCapacityUnits: "5"
+      StreamSpecification:
+        StreamViewType: "NEW_IMAGE"
+
   UserUploadBucket:
     Type: "AWS::S3::Bucket"
     Properties:
+      BucketName: "media-atom-maker-upload-dev"
```
mbarton: A fixed bucket name is required to avoid a circular dependency with the lambda trigger

cloudformation/media-atom-maker-dev.yml
```diff
@@ -218,6 +256,43 @@ Resources:
...
       Timeout: 300
 
+  UploaderS3EventsPermission:
+    Type: "AWS::Lambda::Permission"
+    Properties:
+      Action: "lambda:InvokeFunction"
+      FunctionName: !Ref UploaderS3EventsLambda
+      Principal: "s3.amazonaws.com"
+      # use hard-coded name rather than ref to avoid circular dependency
+      SourceArn: "arn:aws:s3:::media-atom-maker-upload-dev"
```
akash1810: could we pass the bucket name as a param or have it as a mapping?

cloudformation/media-atoms-dynamo-code.json
```diff
@@ -15,8 +15,12 @@
             "Description": "Code audit dynamo table",
             "Type": "String",
             "Default": "audit-media-atom-maker-CODE"
+        },
+        "UploadTrackingAtomMakerCodeTable": {
```
akash1810: do we want to bite the bullet now and put new dynamo tables in the main cfn template? cc @jennysivapalan

uploader/src/main/scala/com/gu/media/DynamoEventsLambda.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.media
+
+import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
+import com.amazonaws.services.lambda.runtime.events.DynamodbEvent
+import scala.collection.JavaConverters._
+
+class DynamoEventsLambda extends RequestHandler[DynamodbEvent, Unit] {
+  override def handleRequest(input: DynamodbEvent, context: Context): Unit = {
+    val records = input.getRecords.asScala
+    val log = context.getLogger
```
akash1810: would `logger` be a clearer name?

cloudformation/media-atoms-dynamo-code.json
mbarton: I'd vote to do this as part of a wider refactor of the templates into stacks

jennysivapalan: It looks like merging them  into the main one and keeping the same dynamo tables isn't possible just yet. I'm fairly sure it will create new tables if we merge them into the main one. We should double check but I vaguely remember this conversation with @philwills a couple weeks ago. 
So it's up to you - we could put all the dynamo cloudformation in this one file and it be the one thing outside of the main one?

mbarton: @jennysivapalan if we intend to move them in the future, would it make sense to do that now while the tool is lightly used? (so we could possibly copy data without downtime)

jennysivapalan: Chatted IRL and we've agreed to leave it as-is but document the use of two cloudformations for the stack

cloudformation/media-atom-maker.json
```diff
@@ -396,6 +433,52 @@
...
+              }
+            },
+            "MemorySize": 256,
+            "Role":  { "Fn::GetAtt": [ "UploaderRole", "Arn" ] },
+            "Runtime": "java8",
+            "Timeout": 300
+          }
+        },
+
+        "UploaderDynamoEventsMapping": {
```
jennysivapalan: What does this do?

cloudformation/media-atoms-dynamo-prod.json
```diff
@@ -118,7 +123,47 @@
...
+              },
+              {
+                "AttributeName": "id",
+                "KeyType": "RANGE"
+              }
+            ],
+            "ProvisionedThroughput": {
+              "ReadCapacityUnits": "5",
+              "WriteCapacityUnits": "5"
+            },
```
jennysivapalan: Is 5 enough for Prod?

mbarton: I don't know so I copied what the other tables are using. I will be throttling the updates to the table to happen only when a part has finished uploading or every X seconds to update the progress of an upload.

cloudformation/media-atom-maker.json
mbarton: Connects the lambda to a stream of updates to the tracking table. It will then use that to make decisions (upload a part to YouTube etc)

PR #299, opened 2017-02-28
Fix name of lambda function in riff-raff

mbarton:
Quick fix to the failing deploys on CODE due to the old CloudFormation name for the uploader lambda being hard-coded in riff-raff.yaml.

NB: a full fix for this is in #298, which sets the names directly.


Reettaphant:
👍

PR #300, opened 2017-03-01
Move expiring atoms to a lambda

### LINKED ISSUE #251: Run expiry poller in a single place
Currently the expiry poller runs on all instances which is a recipe for some nice 'n tasty DynamoDB conflicts. Instead it should run in a single location.

One obvious solution is to run in a Lambda function:

- Write in JS to avoid slow JVM start-up time
- Download secret config from S3 bucket
- List DynamoDB table to find atoms that need to be expired
- **EXPIRE THEM** 
  - Set YouTube videos to private
  - Set atom to expired
  - Expire related video pages (#214)

mbarton:
Expiring atoms was previously performed by the `ExpiryPoller`, which ran periodically on all instances of the app. This PR replaces it with a lambda scheduled to run with the same period.

The only action performed when expiring an atom is to make all the YouTube videos private, so the lambda doesn't need access to the dynamo tables at all. Instead it makes a CAPI query to get all the atoms, rolls through them finding expired ones and makes the videos private.

Fixes #251


app/controllers/Api2.scala
```diff
@@ -77,13 +75,7 @@ class Api2 (override val stores: DataStores, conf: Configuration, val authAction
 
   def putMediaAtom(id: String) = APIHMACAuthAction { implicit req =>
     parse(req) { atom: MediaAtom =>
-      val thriftAtom = atom.asThrift
-      val atomWithExpiryChecked = expiryPoller.updateStatusIfExpired(thriftAtom) match {
```
mbarton: Expiring a newly created atom is still possible and is now covered by a special case when publishing (below)

common/src/main/scala/com/gu/media/lambda/LambdaBase.scala
```diff
@@ -10,32 +10,40 @@ import com.gu.media.aws.AwsAccess
 import com.typesafe.config.{Config, ConfigFactory}
 
 trait LambdaBase extends AwsAccess {
-  private val regionName = envSetting("REGION")
```
mbarton: These changes allow us to run the lambda locally without downloading the config from S3

common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala
```diff
@@ -81,7 +81,7 @@ trait YouTubeVideos { this: YouTubeAccess with Logging =>
       case _ => None
     }
 
-  def setStatusToPrivate(id: String, atomId: String): Unit = {
+  def setStatusToPrivate(id: String): Unit = {
```
akash1810: `Unit` 😷

expirer/src/main/scala/com/gu/media/expirer/ExpirerLambda.scala
```diff
@@ -0,0 +1,71 @@
...
+        setStatusToPrivate(video)
+      } catch {
+        case NonFatal(err) =>
+          log.error(s"Unable to expire $video", err)
+      }
+    }
+  }
+
+  @tailrec
+  private def getExpiredAssets(pageSize: Int, page: Int, now: Long, before: Set[String]): Set[String] = {
```
akash1810: `getExpiredAssets` or `getExpiredAtoms`?

expirer/src/test/scala/com/gu/media/CapiResponses.scala
```diff
@@ -0,0 +1,60 @@
+package com.gu.media
+
+import java.time.Instant
+
+trait CapiResponses {
+  def capiResult(atoms: List[String], page: Int = 1, pages: Int = 1) =
+    s"""
```
akash1810: would be better to put test data in an external file(s)

common/src/main/scala/com/gu/media/Settings.scala
```diff
@@ -6,9 +6,9 @@ import scala.collection.JavaConverters._
 trait Settings {
   val config: Config
 
-  def getString(name: String): Option[String] = if(config.hasPath(name)) { Some(config.getString(name)) } else { None }
-  def getStringList(name: String): List[String] = if(config.hasPath(name)) { config.getStringList(name).asScala.toList } else { List.empty}
-  def getBoolean(name: String): Option[Boolean] = if(config.hasPath(name)) { Some(config.getBoolean(name)) } else { None }
+  def getString(name: String): Option[String] = if(config != null && config.hasPath(name)) { Some(config.getString(name)) } else { None }
```
mbarton: The trait stacking and overriding meant that it was null in test. I'll take another look at how to structure it, it's not a great fix and I ran into the same problem writing tests for the uploader lambda

expirer/src/main/scala/com/gu/media/expirer/ExpirerLambda.scala
mbarton: It accumulates a set of all the video ids referenced by expired atoms. Perhaps `getExpiredVideos` is clearer?

common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala
mbarton: I'm not sure what you mean here?

PR #301, opened 2017-03-01
Show Expired in status bar

mbarton:
Changes the indicator at the top of an atom page to show if it has expired.

e109a2eb-0575-4d02-936b-100d8d06b357

Colours selected in honour of:

clifford_325


mr-mr:
I'm disappointed there's no skull icon in the material library. Would be a nice little addition

PR #302, opened 2017-03-01
mandatory description

Reettaphant:
Make the description field a mandatory field. 

Old atoms that do not have description field will require the addition of one if their metadata gets edited. 

@akash1810


Reettaphant:
Description should only be mandatory if the video is public. For the ui to make sense, we need to merge the metadata and youtube forms. This will be done in another PR.

PR #304, opened 2017-03-01
Frontpage grid updates

mr-mr:
Small design tweaks to the frontpage, preparing for shortcut links (edit and select video).

Once an atom has a video added to it's asset list, the 'no video' lozenge is removed. There might be cases to add different lozenges in (unpublished changes, but for now it's cleaner to show nothing if everything is green with the atom in question.

image

image


akash1810:
This looks really good!! Nice work!

PR #307, opened 2017-03-02
Fix name of lambda function in riff-raff conf

mbarton:
Quick fix to the failing deploys on CODE due to the old CloudFormation name for the uploader 
lambda being hard-coded in riff-raff.yaml... again (See #299)

NB: a full fix for this is in #298, which sets the names directly.

### DIFF
--- a/conf/riff-raff.yaml
+++ b/conf/riff-raff.yaml
@@ -14,7 +14,7 @@ deployments:
       fileName: media-atom-uploader.zip
       functions:
         CODE:
-          name: media-atom-maker-CODE-UploaderLambda-QFNJ4Y06QLZ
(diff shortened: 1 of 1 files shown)


akash1810:
might want to consider setting the `FunctionName` value in cloudformation - http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-functionname

mbarton:
Agreed, although changing to use the `functionNames` parameter riff-raff also solves the problem: https://github.com/guardian/media-atom-maker/blob/6cc6f6adadc912f0b21f29f0c4a003891b25566c/conf/riff-raff.yaml#L14

PR #311, opened 2017-03-03
blackbox integration tests

jharewinton:
Initial setup to run a short integration test against a deployed build of the app

This also breaks the unit tests out into a separate suite, so tests need to be run more specifically. See readme for details


test/Suites.scala
```diff
@@ -0,0 +1,7 @@
+import integration.IntegrationTests
+import org.scalatest.Suites
+import test.ThriftUtilSpec
+
+class UnitTestsSuite extends Suites(new ThriftUtilSpec)
```
mbarton: Perhaps we could tag the integration tests instead? http://www.scalatest.org/user_guide/tagging_your_tests

That way we don't have to remember to add new unit tests to the suite

test/integration/IntegrationTests.scala
```diff
@@ -0,0 +1,82 @@
...
+    /* Create the atom */
+
+    val body = RequestBody.create(JSON, json)
+    val response = gutoolsPost(s"$targetBaseUrl/api2/atoms", Some(body))
+
+    response.code() should be (201)
+
+    val atomId = (Json.parse(response.body().string()) \ "id").get.as[String]
+
+    Json.parse(gutoolsGet(s"$targetBaseUrl/api2/atoms/$atomId/published").body().string()).toString() should be ("{}")
```
mbarton: I think this should be after the eventually block testing the main API response, otherwise it may not catch my mistake before

test/Suites.scala
jharewinton: Solid idea, will see if I can get regular 'sbt test' to ignore the integration tests by default to make local testing easier too

mbarton: I've done that before by putting integration tests in `src/it/scala` and adding that as a test root but it always felt smelly and non-standard

test/integration/IntegrationTests.scala
```diff
@@ -0,0 +1,82 @@
...
+  "Creating a new atom, adding an asset and making it the current asset" should "be represented in the atom maker API" in {
+    val asset = Config.asset
+    val assetId = Config.assetId
+    val json = generateJson(
+      title = s"test-atom-$timestamp",
+      description = "test atom",
+      category = "News",
+      channelId = Config.channelId,
+      youtubeCategoryId = Config.youtubeCategoryId,
+      expiryDate = DateTime.now().getMillis + (100 * 60 * 60 * 24)
```
akash1810: minor: could use a variable to remove the (semi) magic number

jharewinton:
This has fundamentally changed over the last coupe of days, so going to close this PR for now

PR #313, opened 2017-03-06
Initial upload tracking API

mbarton:
This PR adds an API for tracking running uploads against an atom. The UI has not been modified, other than to hit the new end-points.

When the user selects a file, they post to the API with the atom ID and size of the file. The response contains a number of parts (1GB). The client then uploads each parts in turn, requesting S3 bucket access credentials for each part individually. This chunking will allow us to start uploading to YouTube before we have finished uploading to S3.

Once a part has been uploaded, a lambda is triggered to update the dynamo table backing the API. Future PRs will react to that update to begin YouTube uploads etc.

*Notes for reviewers*:

- The `atomId` and `part` are passed as query string parameters because the route names were very unwieldy otherwise. It still feels hacky so I'm very open to better suggestions!


mbarton:
This will now need to change sufficiently to require a new PR

PR #315, opened 2017-03-06
add rel="noopener noreferrer" attribute because target=_blank is insecure

akash1810:
see https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


SiAdcock:
👍  awesome!

PR #318, opened 2017-03-08
Chunked uploads

mbarton:
Modifies the work in progress upload UI to upload files in 100MB chunks. These chunks will be our unit of failure recovery if we fail to upload to YouTube (or if we fail to upload to S3, in the future).

The client begins by sending the file size to the server, which determines how the file will be split. This information is recorded in a dynamo table to support progress tracking by other users (in future PRs).

The client then proceeds to perform S3 multipart uploads for each chunk, one after the other. STS upload credentials are requested for each chunk. The progress bar shown to users is in aggregate across all chunks.

[Dodgy architecture diagram](https://docs.google.com/drawings/d/1ATMKjbI2OAyenEgNnEuinWj61-5cDOl-2drsFUw2rtc/edit): 

direct video upload

This PR also contains supporting cloud-formation changes:

- Removal of dynamo triggered lambda (this will not be used going forward)
- Fix typo in media-atom-expirer build key
- Remove atomId from schema of dynamo uploads table

Notes to reviewers:
- The API route names are a bit smelly (`_credentials` etc). Can we think of ways to improve them?


app/controllers/UploadController.scala
```diff
@@ -2,64 +2,91 @@ package controllers
...
-    val credentials = generateCredentials(uploadId, keyPolicy)
-    log.info(s"Received STS credentials. uploadKey=$key. atomId=$atomId")
+      table.put(upload)
+      Ok(Json.toJson(upload))
+    }
+  }
 
-    val policy = UploadPolicy(awsConfig.userUploadBucket, key, awsConfig.region.toString, credentials)
-    Ok(Json.toJson(policy))
+  def stop(id: String) = APIHMACAuthAction {
```
jennysivapalan: I think the method should be called `delete`.

app/controllers/UploadController.scala
```diff
@@ -2,64 +2,91 @@ package controllers
...
-    ))
+object UploadController {
+  case class CreateRequest(atomId: String, filename: String, size: Long)
+  case class CreateResponse(id: String, region: String, bucket: String, parts: List[UploadPart])
 
-    Json.stringify(json)
-  }
+  implicit val createRequestFormat: Format[CreateRequest] = Jsonx.formatCaseClass[CreateRequest]
+  implicit val createResponseFormat: Format[CreateResponse] = Jsonx.formatCaseClass[CreateResponse]
 }
```
jennysivapalan: I like the use of case classes here, makes it easier to follow. 👍

cloudformation/media-atom-maker-dev.yml
```diff
@@ -18,6 +18,10 @@ Parameters:
     Description: "Stage name"
     Type: "String"
     Default: "DEV"
+  TranscodeOutputBucketName:
```
jennysivapalan: oops, sorry for forgetting to add this.

common/src/main/scala/com/gu/media/upload/Upload.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json.Format
+
+case class Upload(id: String, atomId: String, user: String, bucket: String, region: String, parts: List[UploadPart])
+
+object Upload {
+  implicit val format: Format[Upload] = Jsonx.formatCaseClass[Upload]
+
+  def calculateChunks(size: Long): List[(Long, Long)] = {
```
jennysivapalan: A unit test would be useful here.

conf/routes
```diff
@@ -26,7 +26,10 @@ PUT     /api2/atom/:id/asset-active     controllers.Api2.setActiveAsset(id)
 PUT     /api2/atom/:id/pluto-id         controllers.Api2.setPlutoId(id)
 
 # user uploaded videos
-POST    /api2/atom/:atomId/uploads      controllers.UploadController.create(atomId)
+GET     /api2/uploads                   controllers.UploadController.list(atomId)
+POST    /api2/uploads                   controllers.UploadController.create
+DELETE  /api2/uploads/:id               controllers.UploadController.stop(id)
+POST    /api2/uploads/:id/_credentials  controllers.UploadController.credentials(id, key)
```
jennysivapalan: Why with `_` could we just have `/api2/uploads/:id/credentials/:key` or could the key be sent as a header?

common/src/main/scala/com/gu/media/upload/Upload.scala
mbarton: I've gone one better and added a wizzy ScalaCheck properties-based test. I also refactored the code to hopefully be more readable and less `while` loopy

PR #322, opened 2017-03-14
add delete endpoint & include in integration tests

jharewinton:
Adds a delete endpoint to api2 to allow for atom deletion. This is used in the integration tests to allow the tests to cleanup after themselves. This also means we can start running tests on a schedule, a la prodmon, so I have included a runner shell script that will notify to slack on failure


integration-tests/src/test/scala/IntegrationTests.scala
```diff
@@ -2,20 +2,31 @@ package integration
...
 
-class IntegrationTests extends FlatSpec with Matchers with Eventually with IntegrationPatience with GuHttp with TestAtomJsonGenerator {
+class IntegrationTests extends FlatSpec with Matchers with Eventually with IntegrationPatience with GuHttp with TestAtomJsonGenerator with BeforeAndAfterAll {
 
   val targetBaseUrl: String = Config.targetBaseUrl
   val JSON = MediaType.parse("application/json; charset=utf-8")
 
   def apiUri(atomId: String): String = s"$targetBaseUrl/api/atom/$atomId"
 
+  var createdAtoms: List[String] = List() /* Add all created atoms IDs to this list as first action after atom created. This allows for test cleanup outside the test flow  */
```
jharewinton: I appreciate that the use of a var here is very naughty, and I promise not to do it again

akash1810: could use a `ListBuffer`?

PR #323, opened 2017-03-14
Update media-atom-maker-dev.yml

akash1810:
Bucket names are globally unique, setting the name in the template means we can only run one DEV stack. Lets get Amazon to generate the bucket name.


mbarton:
@akash1810 I think I fixed this in #318 by making the transcoder bucket name a parameter

PR #320, opened 2017-03-14
Kinesis stream of upload operations

mbarton:
NB: this PR depends on #318 

This PR adds a kinesis stream of instructions to upload a chunk of a video from S3 to YouTube. The actual uploading itself will be in a subsequent PR.

We now also use S3 multipart copy to stitch together all the chunks into a single key that can be used to sync back to Pluto/send for transcoding.

PR #324, opened 2017-03-14
Upload video parts to YouTube

mbarton:
**NB: This PR is dependent on #320**

This PR reacts to events on the kinesis stream introduced in #320 by uploading each part in turn to YouTube. Once all parts have been uploaded, the resulting YouTube video is added as an asset to the atom.

Other changes:

- Use Dynamo consistent reads to fix problem where part uploads would not see the YouTube resumable upload end-point in the DB
- Only use plain S3 PUT from the client rather than multipart uploads to avoid a really annoying race condition where the last part (small and therefore a PUT) would trigger the lambda before the preceding multipart upload
- Include `plutoProjectId` in the upload metadata (required for integration with Pluto sync)


mbarton:
Temporarily closing until #320 is re-opened

PR #325, opened 2017-03-15
Kinesis stream of upload actions

mbarton:
This PR adds a kinesis stream of actions to be performed as parts of a video are uploaded.

The original design (#320) used a lambda triggered by S3 PutObject events. Unfortunately, the ordering of the events received was not always correct, leading to problems when hooked up to YouTube which does not like receiving part 2 before part 2 etc.

This PR removes the S3 lambda and sends the upload complete message on Kinesis once the client has 'completed' the part, in a separate POST request. Once all parts are complete, we trigger an S3 multipart copy to create the `full` part which can be ingested by Pluto.

`UploadAction` subclasses express the operations sent on the stream:

- `UploadPartToYouTube`: currently just logs, subsequent PR will add the functionality
- `DeleteParts`: removes parts, leaving just the `full` part

Potentially unhelpful architecture diagram:

manual complete uploading


common/src/main/scala/com/gu/media/upload/UploadsTable.scala
```diff
@@ -4,17 +4,10 @@ import com.gu.media.aws.DynamoAccess
 import com.gu.scanamo.syntax._
```
Reettaphant: Is there some reason why this is not called a datastore like the atom and audit datastores?

app/controllers/UploadController.scala
```diff
@@ -46,41 +51,54 @@ class UploadController(val authActions: HMACAuthActions, awsConfig: AWSConfig, o
...
+      NoContent
     }
   }
 
-  private def buildUpload(atom: MediaAtom, user: User, size: Long) = {
-    val id = UUID.randomUUID().toString
-
-    Upload(
-      id = id,
+  private def buildUpload(atom: MediaAtom, channelId: String, user: User, size: Long) = {
```
akash1810: could we get the `channelId` from the `atom`?

cloudformation/media-atom-maker-dev.yml
```diff
@@ -240,17 +240,26 @@ Resources:
                     - "arn:aws:dynamodb:${Region}:${Account}:table/${Table}"
                     - { Region: !Ref "AWS::Region", Account: !Ref "AWS::AccountId", Table: !Ref "UploadTrackingDynamoTable" }
               - Effect: "Allow"
-                Action: ["s3:GetObject"]
+                Action: ["s3:*"]
```
akash1810: why this change?

common/src/main/scala/com/gu/media/aws/UploadAccess.scala
```diff
@@ -12,6 +12,10 @@ trait UploadAccess { this: Settings with AwsAccess =>
   lazy val uploadSTSClient = createUploadSTSClient()
 
   private def createUploadSTSClient() = {
+    if(!userUploadRole.startsWith("arn:")) {
```
akash1810: why do we need to do this?

mbarton: If you use the non-arn format you only get an "Unauthorized" error from Amazon which is quite confusing. After the second time it happened to me, I added a check to give a better exception 😄

app/controllers/UploadController.scala
mbarton: we do, I just pattern match on it at the level above since we need to issue an error if it is missing

cloudformation/media-atom-maker-dev.yml
mbarton: To allow the deletion of entries once uploads are complete. I'll make the permissions more specific to be clearer

app/controllers/UploadController.scala
mbarton: I'll change it so we pattern match inside the method and throw an exception (caught in the `parse` wrapper method) if it isn't there

mbarton: Accidentally fixed in #326 instead of this PR

common/src/main/scala/com/gu/media/upload/UploadLifecycle.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  private def chunk(uploadId: String, size: Long): List[UploadPart] = {
+    val boundaries = Upload.calculateChunks(size)
+
+    boundaries.zipWithIndex.map { case ((start, end), id) =>
+      UploadPart(UploadPartKey(aws.userUploadFolder, uploadId, id).toString, start, end)
+    }
+  }
+
+  private def createFullObject(upload: Upload): UploadFullKey = {
```
jennysivapalan: I'm not sure what `createFullObject` could it be renamed to something more descriptive?

mbarton: I renamed it to `createCompleteObject` and the corresponding URI class to `CompleteUploadKey`

PR #326, opened 2017-03-15
Upload video parts to YouTube

mbarton:
**NB: This PR is dependent on #325**

This PR reacts to events on the kinesis stream introduced in #325 by uploading each part in turn to YouTube. Once all parts have been uploaded, the resulting YouTube video is added as an asset to the atom.

### DIFF
--- a/cloudformation/media-atom-maker.json
+++ b/cloudformation/media-atom-maker.json
@@ -488,7 +488,8 @@
                 "APP": "media-atom-uploader",
                 "STAGE": { "Ref": "Stage" },
                 "CONFIG_BUCKET": { "Ref": "ConfigBucket" },
-                "CONFIG_KEY": { "Fn::FindInMap": [ "LambdaConfig", "Uploader", { "Ref": "Stage" }] }
+                "CONFIG_KEY": { "Fn::FindInMap": [ "LambdaConfig", "Uploader", { "Ref": "Stage" }] },
+                "DOMAIN": { "Fn::Join": ["", [ "https://", { "Ref": "DomainToMonitor" } ]] }
               }
             },
             "MemorySize": 256,
--- a/cloudformation/media-atom-maker.json
+++ b/cloudformation/media-atom-maker.json
@@ -488,8 +488,7 @@
                 "APP": "media-atom-uploader",
                 "STAGE": { "Ref": "Stage" },
                 "CONFIG_BUCKET": { "Ref": "ConfigBucket" },
-                "CONFIG_KEY": { "Fn::FindInMap": [ "LambdaConfig", "Uploader", { "Ref": "Stage" }] },
-                "DOMAIN": { "Fn::Join": ["", [ "https://", { "Ref": "DomainToMonitor" } ]] }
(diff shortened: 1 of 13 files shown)


uploader/src/main/scala/com/gu/media/InputStreamRequestBody.scala
```diff
@@ -0,0 +1,53 @@
...
+      // terminate
+
+      case remaining =>
+        val amount = if(remaining > SEGMENT_SIZE) { SEGMENT_SIZE } else { remaining }
+        val written = source.read(sink.buffer(), amount)
+
+        // important otherwise we just make a massive buffer!
+        sink.flush()
+
+        if(written == -1) {
```
akash1810: could this be `if(written != -1)` to remove the empty block of code?

uploader/src/main/scala/com/gu/media/UploadActionsLambda.scala
```diff
@@ -2,31 +2,101 @@ package com.gu.media
...
+
+    val url = startUpload(upload.metadata.title, upload.youTube.channel, upload.id, upload.parts.last.end)
+    val updated = upload.copy(youTube = upload.youTube.copy(upload = Some(url)))
+
+    table.put(updated)
+    url
+  }
+
+  private def addAsset(atomId: String, videoId: String): Unit = {
+    val actuallyPerformRequest = stage != "DEV" && domain.nonEmpty
```
akash1810: why can't we do this in `DEV`?

mbarton: I'd like to make it work in DEV, but since it's performing an HMAC request from a lambda running in AWS, I couldn't think of a way of making it hit the end-point (since I'm running the server locally).

mbarton: As discussed IRL, I'll do a follow-up PR that calls the upload code directly from the app in DEV, bypassing Kinesis. CODE and PROD will still use the kinesis route so uploads survive a re-deploy and we get retry on error

cloudformation/media-atom-maker.json
```diff
@@ -488,7 +488,8 @@
                 "APP": "media-atom-uploader",
                 "STAGE": { "Ref": "Stage" },
                 "CONFIG_BUCKET": { "Ref": "ConfigBucket" },
-                "CONFIG_KEY": { "Fn::FindInMap": [ "LambdaConfig", "Uploader", { "Ref": "Stage" }] }
+                "CONFIG_KEY": { "Fn::FindInMap": [ "LambdaConfig", "Uploader", { "Ref": "Stage" }] },
+                "DOMAIN": { "Fn::Join": ["", [ "https://", { "Ref": "DomainToMonitor" } ]] }
```
akash1810: what's this env var used for?

mbarton: The HMAC add asset request

akash1810: discussed IRL - can we use the value in the S3 config file?

mbarton: As discussed IRL, I will remove this and replace it with the `host` setting from `media-atom-maker.private.conf`

PR #329, opened 2017-03-16
add publish step to integration tests

jharewinton:
also splits the formerly large integration tests into a series of tests + some simple splitting up of the test framework to make it easier to add more tests


integration-tests/src/test/scala/AtomCreationTestBase.scala
```diff
@@ -0,0 +1,80 @@
...
+    assetResponse.code() should be(200)
+
+    eventually {
+      val assets = (Json.parse(gutoolsGet(apiEndpoint).body().string()) \ "data" \ "assets") (0)
+      (assets \ "id").get.as[String] should be(Config.assetId)
+    }
+  }
+
+  test("Make an asset current for an existing atom") {
+    val currentAssetResponse = gutoolsPut(s"${targetBaseUrl}/api2/atom/${atomId}/asset-active", Some(jsonBody(s"""{"youtubeId":"${Config.assetId}"}""")))
```
jennysivapalan: small one but `targetBaseUrl` and `atomId` don't need the braces as they are strings.

integration-tests/src/test/scala/IntegrationTestBase.scala
```diff
@@ -0,0 +1,34 @@
...
+class IntegrationTestBase extends FunSuite with Matchers with Eventually with IntegrationPatience with GuHttp with TestAtomJsonGenerator with BeforeAndAfterAll {
+
+  val targetBaseUrl: String = Config.targetBaseUrl
+
+  def apiUri(atomId: String): String = s"$targetBaseUrl/api/atom/$atomId"
+
+  var atomStore = new ListBuffer[String]() /* Add all created atoms IDs to this list as first action after atom created. This allows for test cleanup outside the test flow  */
+
+  def deleteAtom(id: String) = {
+    println(s"Deleting atom $id")
```
jennysivapalan: Swap the println for log line? There is one below too.

integration-tests/src/test/scala/services/GuHttp.scala
```diff
@@ -27,7 +30,10 @@ trait GuHttp {
 
     body match {
       case Some(e) => req.post(e)
-      case None => req.addHeader("Content-Length", "0")
+      case None => {
+        req.addHeader("Content-Length", "0")
+        req.post(emptyBody)
```
jennysivapalan: Why do we do this?

PR #334, opened 2017-03-21
Direct uploading support in DEV

mbarton:
Previously we were adding assets via a HMAC authenticated request to media-atom-maker, which naturally did not work in DEV.

The fix is to bypass the lambda when running in DEV, running the upload actions in a separate thread in the main application.

We now have two interfaces:

- `UploadActionSender`: with Kinesis and separate-thread backed implementations
- `UploadActionHandler`: with sub-classes for adding the asset using an HMAC request (CODE, PROD) and adding it directly using the existing `AddAssetCommand` (DEV)

Additional stuff:

- Add an `uploadedToYouTube` boolean to each `UploadPart` (to display progress to the user in subsequent PRs)
- Move the S3 multipart copy to be an `UploadAction` (and thus happen asynchronously)
- Include asset version when syncing with Pluto
- Start YouTube multipart upload before sending first upload action to the queue
  - This avoids a race condition that was removing the YouTube upload URI from the database
- Split `UploadLifecycle` into `UploadController` and `UploadActionHandler`
- Remove unused progress tracking function from `InputStreamRequestBody`


Reettaphant:
👍

PR #335, opened 2017-03-22
Make upload actions self describing

mbarton:
This PR removes the reliance on the uploads table for making decisions on how to process an action (upload to YouTube, copy parts etc). This removes a class of race conditions that were removing upload progress inadvertently, making uploads more reliable.

The `Upload` class has been restructured to be completely immutable, apart from an `UploadProgress` field recording the number of bytes uploaded to S3 and YouTube. This is updated on the `complete` callback and is now commutative, in that completing part 4 cannot accidentally mark part 3 as incomplete.

The other piece of state previously in the table was the YouTube multipart upload URI. This is now send back to the client on the first `complete` and parroted back for each subsequent request.

NB: this means we send a YouTube URL used internally to the client. I don't believe this to be a risk since the user is panda authed and would need access to our YouTube credentials to use it.


akash1810:
> The other piece of state previously in the table was the YouTube multipart upload URI. This is now send back to the client on the first complete and parroted back for each subsequent request.
NB: this means we send a YouTube URL used internally to the client. I don't believe this to be a risk since the user is panda authed and would need access to our YouTube credentials to use it.

Not sure why we need to do this?

mbarton:
@akash1810 it means we never have to look-up the state from Dynamo in order to generate the next action to go to the lambda. Otherwise we have to store the state and use consistent reads to access it

PR #340, opened 2017-03-23
Send uploaded videos to pluto

Reettaphant:
Sends uploaded videos to pluto

- If the associated atom has pluto project id, upload goes to a kinesis stream for pluto to read from
- If pluto id is missing, upload goes to a dynamo table where the project id has to be manually added. It goes to the same kinesis stream for pluto to read from afterwards
- Once pluto has finished ingesting, delete s3key from s3 and remove the associated entry from the dynamo table

@akash1810 @mbarton


conf/routes
```diff
@@ -25,6 +25,10 @@ PUT     /api2/atom/:id/publish          controllers.Api2.publishMediaAtom(id)
 PUT     /api2/atom/:id/asset-active     controllers.Api2.setActiveAsset(id)
 PUT     /api2/atom/:id/pluto-id         controllers.Api2.setPlutoId(id)
 
+PUT     /api/pluto/:id/send              controllers.Api2.sendToPluto(id)
+GET     /api2/pluto-list                controllers.Api2.getPlutoAtoms()
```
mbarton: Maybe rationalise these to something like:

```
GET /api2/pluto
PUT /api2/pluto/:id
```

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.duration._
+import play.api.libs.json._
+import com.amazonaws.services.sqs.model.{DeleteMessageRequest, ReceiveMessageRequest, Message}
+import scala.concurrent.ExecutionContext
+import scala.collection.JavaConversions._
+import com.gu.media.logging.Logging
+import data.DataStores
+
+case class PlutoMessageConsumer(val stores: DataStores, awsConfig: AWSConfig)
```
mbarton: Ah didn't realise it was SQS not Kinesis. Either way, since the call blocks we should consider running this in its own Akka dispatcher to avoid blocking the threads on the main app

app/data/DataStores.scala
```diff
@@ -61,6 +63,7 @@ trait UnpackedDataStores {
   val previewDataStore: PreviewDataStore = stores.preview
   val publishedDataStore: PublishedDataStore = stores.published
   val auditDataStore: AuditDataStore = stores.audit
+  val plutoDataStore: PlutoDataStore = stores.pluto
```
mbarton: These guys are only here for compatibility with older code. I'd just call `stores.pluto` instead

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,63 @@
...
+    ).getMessages.toList
+
+  private def handleMessage(msg: Message): Unit = {
+
+    val body = Json.parse(msg.getBody)
+
+    (body \ "Message") match {
+      case JsDefined(message) => {
+        val messageString = message.toString
+        val s3Key = messageString.substring(1, messageString.length - 1)
```
mbarton: Why do we need to do this? Maybe make the JSON parsing return a case class?

project/Dependencies.scala
```diff
@@ -34,6 +35,10 @@ object Dependencies {
   val awsLambdaCore = "com.amazonaws" % "aws-lambda-java-core" % "1.1.0"
   val awsLambdaEvents = "com.amazonaws" % "aws-lambda-java-events" % "1.3.0"
   val awsTranscoder = "com.amazonaws" % "aws-java-sdk-elastictranscoder" % awsVersion
+  val awsSNS = "com.amazonaws" % "aws-java-sdk-sns" % awsVersion
+  val awsSQS = "com.amazonaws" % "aws-java-sdk-sqs" % awsVersion
+  val awsKinesisClient = "com.amazonaws" % "amazon-kinesis-client" % kinesisClientVersion
```
mbarton: Not required any more?

public/video-ui/src/actions/PlutoIdActions/addProject.js
```diff
@@ -0,0 +1,36 @@
+import PlutoVideosApi from '../../services/PlutoVideosApi';
+
+function requestAddProject() {
+  return {
+    type:       'ADD_PROJECT_GET_REQUEST',
```
mbarton: This might be style nitpicking but I don't think these action names convey intent very well.

Maybe instead:

```
ADD_PROJECT_REQUEST
PROJECT_ADDED
```

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,70 @@
...
+  private def getMessages(waitTime: Int, maxMessages: Int): Seq[Message] =
+    awsConfig.sqsClient.receiveMessage(
+      new ReceiveMessageRequest(awsConfig.plutoQueueUrl)
+        .withWaitTimeSeconds(waitTime)
+        .withMaxNumberOfMessages(maxMessages)
+    ).getMessages.asScala.toList
+
+  private def handleMessage(msg: Message): Unit = {
+
+    val body = Json.parse(msg.getBody)
```
mbarton: I'm not sure this allows for extension of `PlutoMessage`, since you're passing in the string.

Maybe something like:

```scala
val json = Json.parse(msg.getBody)

(json \ "Messsage").validate[PlutoMessage] match {
  case JsSuccess(message, _) =>
     // do stuff

  case JsError(errors) =>
    log.error("Could not extract a message body blah blah")
```

cloudformation/media-atom-maker.json
```diff
@@ -55,6 +55,10 @@
             "Description": "Name of the upload tracking dynamo table",
             "Type": "String"
         },
+        "ManualPlutoTable": {
```
Reettaphant: I'm not sure if this is something we should be doing instead of just adding to the s3 secrets file. @akash1810 do you know a good reason for doing this?

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  private def handleMessage(msg: Message): Unit = {
+
+    val bodyJson = Json.parse(msg.getBody)
+
+    val message: JsLookupResult = (bodyJson \ "Message")
+
+    val validated: JsResult[PlutoMessage] = (bodyJson \ "Message").validate[PlutoMessage]
+
+    (bodyJson \ "Message").validate[PlutoMessage] match {
```
mbarton: We already do this above

cloudformation/media-atom-maker.json
akash1810: all our dynamo tables exist in a separate stack (not sure why). For this reason, we pass them in to the main template as params to craft the [policy](https://github.com/guardian/media-atom-maker/blob/master/cloudformation/media-atom-maker.json#L838). As we have to do this, it makes it simpler to `cat` to the conf file in the launch config so as not to duplicate table names...

I'd love to just merge the two stacks into one template...

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,88 @@
...
+import play.api.libs.json._
+
+case class PlutoMessageConsumer(val stores: DataStores, awsConfig: AWSConfig)
+  extends UnpackedDataStores with Logging {
+
+  val actorSystem = ActorSystem("PlutoMessageConsumer")
+
+
+  def start(scheduler: Scheduler)(implicit ec: ExecutionContext): Unit = {
+    log.info("Starting uploads sqs queue reader")
```
jennysivapalan: Might be worth being clearer as to which sqs we are using, this might be in the `awsConfig`?

project/Dependencies.scala
```diff
@@ -10,6 +10,7 @@ object Dependencies {
   val slf4jVersion = "1.7.21"
   val typesafeConfigVersion = "1.3.0" // to match what we get from Play transitively
   val scanamoVersion = "0.9.1" // to match what we get from atom-publisher-lib transitively
+  val kinesisClientVersion = "1.7.4"
```
jennysivapalan: This isn't used so can be removed.

app/util/PlutoMessageConsumer.scala
```diff
@@ -0,0 +1,88 @@
...
+    actorSystem.scheduler.scheduleOnce(0.seconds)(processMessages())
```
jennysivapalan: Why do we only do this once? Should we be regularly polling the queue?

jennysivapalan:
@Reettaphant As the PR is quite long I think it's worth adding a couple more points to the desciption so we can go back and read it easily if needed. The things I think we can add are that we use email to notify that a Pluto id is missing so needs to be manually added as that's seems an important part of this PR.  On the third point that we get notified that an atom is in Pluto via an SQS queue.

cloudformation/media-atom-maker.json
jennysivapalan: I tried to merge them once but I think it we'd have to migrate the data to a new table.

app/util/PlutoMessageConsumer.scala
Reettaphant: It just starts the queue reader, we only need to start it once.

PR #341, opened 2017-03-24
Set duration on publish

### LINKED ISSUE #336: Set duration on atom at publish
Duration has been missing on some atoms. Setting it on publish guarantees it will be there.

mbarton:
Fixes #336.

Previously we only set the duration when an asset was activated. Now we set it again when the atom is published as a double check.

I've also taken the opportunity to refactor PublishAtomCommand to hopefully be easier to read. As such, it may be easier to look at the file itself rather than the diff when reviewing.

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -35,11 +35,11 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
 
     getActiveAsset(atom) match {
       case Some(asset) if asset.platform == Youtube =>
-        val withDuration = atom.copy(duration = youTube.getDuration(asset.id))
+        val atomWithDuration = atom.copy(duration = youTube.getDuration(asset.id))
 
-        updateThumbnail(withDuration, asset)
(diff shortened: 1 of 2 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -26,75 +26,50 @@ case class PublishAtomCommand(id: String, fromExpiryPoller: Boolean, override va
...
+    if(atom.privacyStatus.contains(PrivacyStatus.Private)) {
+      log.error(s"Unable to publish atom ${atom.id}, privacy status is set to private")
+      AtomPublishFailed("Atom status set to private")
+    }
 
-        atom.activeVersion match {
-          case Some(atomVersion) => {
+    getActiveAsset(atom) match {
+      case Some(asset) if asset.platform == Youtube =>
+        val withDuration = atom.copy(duration = youTube.getDuration(asset.id))
```
akash1810: `atomWithDuration` may be more descriptive

mbarton:
We don't need `fromExpiryPoller` any more since the expirer lambda uses totally different code

PR #339, opened 2017-03-23
UI for direct uploads

mbarton:
Adds a UI for direct uploads. Once the video has been added, we display YouTube processing status.

Processing status is looked up for all assets and stored as state in `VideoTrail`. If any of the assets are still processing, we poll for the progress every 5 seconds.

There are three phases:

- `YouTube processing`: we are waiting for YouTube to pick up the video after it has been uploaded
- `YouTube processing (12s remaining)`: YouTube is transcoding and processing
- `YouTube processing`: we are waiting for YouTube to say everything is finished

Unfortunately only the second phase has progress information available.

There is now also a progress bar within the header that indicates the S3 upload progress.

image

Reviewer notes:
- There are two distinct stages: uploading to S3 and uploading to YouTube
- The former is powered off the `localUpload` redux state key, the latter off `uploads`
- We track all uploads server side but only show the ones uploading to YouTube
   - This avoids the confusion of seeing the in-progress upload from someone else
- The progress bar for YouTube uploads moves in 100MB chunks
   - We use polling to update it

PR #342, opened 2017-03-27
Assets Page UI changes

mr-mr:
NB: depends on #339 

Adds YouTube processing status to the new video assets page introduced in #339.

Processing status is looked up for all assets and stored as state in `VideoTrail`. If any of the assets are still processing, we poll for the progress every 5 seconds.

There are three phases:

- `YouTube processing`: we are waiting for YouTube to pick up the video after it has been uploaded
- `YouTube processing (12s remaining)`: YouTube is transcoding and processing
- `YouTube processing`: we are waiting for YouTube to say everything is finished

Unfortunately only the second phase has progress information available.

There is now also a progress bar within the header that indicates the S3 upload progress.

image


mbarton:
Closed in favour of #339

PR #343, opened 2017-03-28
Add simple "load more" paging to the front page

### LINKED ISSUE #233: Pagination on the front and search pages
A big 'ol list of atoms is not going to scale as we add more, especially since we load an image for each one.

Things to consider:
  - Search results are doing via the CAPI proxy
    - Add a MAM end-point with unified search paging that delegates to CAPI internally?
  - Dynamo secondary index support to order results by last updated time
    - The initial front-page should show the latest atoms
    - Co-ordinate with atom-workshop work since this is likely to be a common requirement

mbarton:
The front page now only shows the latest 30 atoms, with a "Load More" button at the bottom that shows 30 more every time it is clicked.

As a result, this PR unifies listing atoms and searching into a single server-side end-point. In CODE and PROD the results are from CAPI, in DEV they are filtered server-side after a Dynamo scan.

Fixes #233

### DIFF
--- a/public/video-ui/src/components/Videos/Videos.js
+++ b/public/video-ui/src/components/Videos/Videos.js
@@ -8,7 +8,7 @@ class Videos extends React.Component {
   };
 
   componentDidMount() {
-    this.props.videoActions.getVideos();
+    this.props.videoActions.getVideos(this.props.limit);
   }
 
   componentWillReceiveProps(newProps) {
@@ -18,7 +18,7 @@ class Videos extends React.Component {
     if (oldSearch !== newSearch && newSearch !== "") {
       this.props.videoActions.searchVideosWithQuery(newSearch);
     } else if (newSearch === "" && oldSearch !== "") {
-      this.props.videoActions.getVideos();
+      this.props.videoActions.getVideos(this.props.limit);
     }
   }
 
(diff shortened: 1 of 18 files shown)


public/video-ui/src/components/Videos/Videos.js
```diff
@@ -37,12 +36,27 @@ class Videos extends React.Component {
     return (this.props.videos.map((video) => <VideoItem key={video.id} video={video} />));
   }
 
+  renderMoreLink() {
+    if(this.props.limit === -1) {
+      return false;
+    }
```
Reettaphant: Does this mean we don't paginate search results?

public/video-ui/src/reducers/videosReducer.js
```diff
@@ -1,10 +1,15 @@
-export default function videos(state = [], action) {
-  switch (action.type) {
+import {frontPageSize} from '../constants/frontPageSize';
 
+export default function videos(state = { entries: [], limit: frontPageSize }, action) {
+  switch (action.type) {
+    case 'VIDEOS_GET_REQUEST':
+      return { entries: state.entries, limit: action.limit };
+    case 'VIDEOS_SEARCH_REQUEST':
```
Reettaphant: Where is this action type coming from? I can't find it.

public/video-ui/src/components/Videos/Videos.js
Reettaphant: When is this limit set to -1? Should we also render something else when there is nothing more to show?

PR #344, opened 2017-03-28
Remove redux forms

Reettaphant:
Removes redux forms to decrease complexity. After this PR, we will be able to add warnings to fields.

Add asset form remains untouched because it will be removed in the video upload UI work.

@akash1810 @mbarton


public/video-ui/src/components/FormFields/ImageSelector.js
```diff
@@ -19,16 +19,18 @@ class VideoPosterImageEdit extends React.Component {
...
 
-    if (this.props.video && !this.props.editMode) {
+    if (!this.props.editMode) {
       return (
           <div className="form__row">
-            <label className="form__label">Poster image</label>
+            <label className="form__label">{this.props.fieldName}</label>
             <div className="form__imageselect">
               <GridImageSelect
                 video={this.props.video}
```
mbarton: Is this property still required?

public/video-ui/src/components/Video/Display.js
```diff
@@ -166,18 +181,22 @@ class VideoDisplay extends React.Component {
...
+                  updateFormErrors={this.updateYoutubeFormErrors}
                 />
               </div>
               <div className="video__detailbox">
                 <div className="video__detailbox__header__container">
                   <header className="video__detailbox__header">Poster Image</header>
-                  <GridImageSelect editState={this.props.editState} video={this.props.video || {}} updateVideo={this.saveAndUpdateVideo} gridUrl={this.props.config.gridUrl} createMode={false}/>
+                  <GridImageSelect
+                    editState={this.props.editState}
+                    video={this.props.video || {}}
```
mbarton: as above, do we need this property still (instead of setting the value as video.posterImage?)

public/video-ui/src/components/VideoEdit/VideoEdit.js
```diff
@@ -1,87 +1,116 @@
...
-            updateVideo={props.updateVideo} />
+    return (
+      <div className="form__group">
+        <ManagedForm
+          data={this.props.video}
+          updateData={this.props.updateVideo}
+          editable={this.props.editable}
+          updateFormErrors={this.props.updateFormErrors}
+        >
+          <ManagedField
```
mbarton: For me it would be cleaner to have single elements for each field type rather than nesting them, since they're all managed. So you could say something like:

```html
<TextInput field="title" name="Title" maxLength={fieldLengths.title} isRequired />
```

but not a deal breaker

public/video-ui/src/components/VideoEdit/formComponents/ContentFlags.js
```diff
@@ -1,37 +0,0 @@
-import React from 'react';
```
mbarton: Start of my favourite section of the PR...

public/video-ui/src/components/VideoEdit/formComponents/YoutubeKeywords.js
```diff
@@ -1,23 +0,0 @@
...
-      <div className="form__row">
-        <label className="form__label">Keywords</label>
-        <KeywordPicker keywords={this.props.video.tags} updateKeywords={this.updateKeywords} editable={this.props.editable}/>
-      </div>
-    );
-  }
-
-}
-
-export default YoutubeKeywordsSelect;
```
mbarton: End of my favourite section 💥 💥 💥

public/video-ui/src/util/validateField.js
```diff
@@ -0,0 +1,23 @@
+import FieldNotification from '../constants/FieldNotification';
+
+const validateField = (fieldValue, isRequired: false, isDesired: false) => {
```
mbarton: Nice. Is it possible to also add validation rules for a specific form?

public/video-ui/src/components/FormFields/ItemPicker.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import KeywordPicker from '../utils/KeywordPicker';
+
+class ItemPicker extends React.Component {
```
mbarton: I think it would be clearer to make `ItemPicker` and `KeywordPicker` one component, unless it is re-used elsewhere

public/video-ui/src/util/validateField.js
Reettaphant: Yes. I think there are examples in atom workshop for doing this.

public/video-ui/src/components/VideoEdit/VideoEdit.js
Reettaphant: The field component contains functionality for validating specific fields and updating the field values, reducing repetition in the input components.

public/video-ui/src/components/FormFields/ItemPicker.js
Reettaphant: Yes, this makes sense. I guess its quite specific form component so its unlikely to be reused.

public/video-ui/src/components/ManagedForm/ManagedField.js
```diff
@@ -0,0 +1,99 @@
...
+        fieldErrors: notifications.errors,
+        fieldWarnings: notifications.warnings
+      });
+
+      this.props.updateFormErrors(notifications.errors, this.props.name);
+    }
+
+  }
+
+  static propTypes = {
```
akash1810: 😍

public/video-ui/src/util/validateField.js
shaundillon: Yep, we pass validation functions to the specific field as an array and this func loops through them.

shaundillon:
Just dropping in to say this is awesome! I'm amazed it's actually a net removal too. Well done

Reettaphant:
I've made a small modification and moved the error checking functions to the managed form component so I think its all done now.

PR #346, opened 2017-03-29
make publishing atom integration test more robust

jharewinton:


integration-tests/src/test/scala/AtomCreationTests.scala
```diff
@@ -67,9 +68,13 @@ class AtomCreationTests extends IntegrationTestBase with CancelAfterFailure {
...
 
   test("Publishing an existing atom") {
-    val publishResponse = gutoolsPut(s"$targetBaseUrl/api2/atom/$atomId/publish")
+    val initialPublishResponse = gutoolsPut(s"$targetBaseUrl/api2/atom/$atomId/publish")
 
-    publishResponse.code() should be (200)
+    initialPublishResponse.code() match {
+      case 200 =>
+      case 400 => throw new TestFailedException(s"Publishing atom returned 400: ${initialPublishResponse.body().string()}", 5)
+      case _ => { eventually { gutoolsPut(s"$targetBaseUrl/api2/atom/$atomId/publish").code() should be (200) } }
```
jharewinton: thought about making this tail recursive, but seemed like overkill, as would essentially have to duplicate Eventually to do it, for the sake of saving 1 http call that we should only hit in exceptional circumstances

PR #349, opened 2017-03-29
endpoints to get, create and update pluto projects

akash1810:
This PR is also a start of removing the extra dynamo stacks. The new table is defined in the main template and not `cat`ed to the config file in LaunchConfig. Instead, we look up instance tags and [craft the table name in code](https://github.com/guardian/media-atom-maker/pull/349/files#diff-939dbba1b8854def696be19ce74e7717R12) as its deterministic.

### DIFF
--- a/common/src/main/scala/com/gu/media/pluto/PlutoProject.scala
+++ b/common/src/main/scala/com/gu/media/pluto/PlutoProject.scala
@@ -3,7 +3,6 @@ package com.gu.media.pluto
 import org.cvogt.play.json.Jsonx
 import org.joda.time.DateTime
 import play.api.libs.json._
-import play.api.libs.functional.syntax._
 import com.gu.media.util.JsonDate._
 
 case class PlutoProject (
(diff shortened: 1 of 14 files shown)


common/src/main/scala/com/gu/media/pluto/PlutoProject.scala
```diff
@@ -0,0 +1,39 @@
...
+  id: String,
+  collectionId: String,
+  headline: String,
+  productionOffice: String,
+  status: String,
+  created: DateTime
+)
+
+object PlutoProject {
+  implicit val plutoProjectReads: Reads[PlutoProject] = (
```
mbarton: Could we just use `Jsonx.formatCaseClass` and define a `Format` for `DateTime`?

conf/routes
```diff
@@ -28,9 +28,13 @@ PUT     /api2/atom/:id/pluto-id         controllers.Api2.setPlutoId(id)
 PUT     /api2/pluto/:id                 controllers.Api2.sendToPluto(id)
 GET     /api2/pluto                     controllers.Api2.getPlutoAtoms()
 
-
 DELETE  /api2/atom/:id                  controllers.Api2.deleteAtom(id)
 
+GET     /api2/pluto/projects            controllers.PlutoProjectController.listProjects()
+GET     /api2/pluto/projects:id         controllers.PlutoProjectController.getProject(id)
```
mbarton: Missing `/`?

akash1810: hmm, good point! I've been curling it locally with no problems... not sure how though!

common/src/main/scala/com/gu/media/pluto/PlutoProject.scala
akash1810: yep yep! Done.

PR #355, opened 2017-03-31
New navigation states

mr-mr:
PR now rebased with https://github.com/guardian/media-atom-maker/pull/339

So with the addition of the Upload assets page I wanted to address the state of the navigation, to make where the editor is clearer and avoid adding too much extra guff to the page.

I've split it into 3 layers:
Layer one is the homepage - 
image

Layer two is the edit page
image

Layer three is the upload asset to the edit page
image

Plus a gif 'cause I wanted to do one.
vam_navigation

Have a look and see what you think about the choices of what's visible. Not everything is visible on every state and that is by design.

### DIFF
--- a/public/video-ui/src/components/Header.js
+++ b/public/video-ui/src/components/Header.js
@@ -43,6 +43,14 @@ export default class Header extends React.Component {
     );
   }
 
+  renderVideoTitle() {
+    return (
+      <div className="flex-container">
+        <span className="header__video__title">{this.props.video.title}</span>
+      </div>
+    )
+  }
+
   renderFeedback() {
     return (
       <nav className="topbar__nav-link">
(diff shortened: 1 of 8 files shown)


public/video-ui/src/components/Header.js
```diff
@@ -18,15 +18,39 @@ export default class Header extends React.Component {
...
+
+  renderSearch() {
+    return (
+      <div className="flex-container topbar__global">
         <VideoSearch {...this.props}/>
       </div>
     );
   }
 
+  renderHeaderBack() {
```
akash1810: would `renderBack` be explicit enough as we're in the `Header` component already?

jennysivapalan:
👍 Looks good to me @mr-mr, you'll need to update and merge the conflicts with master.

PR #356, opened 2017-04-02
pluto project ingestion lambda

akash1810:
depends on https://github.com/guardian/node-riffraff-artefact/pull/21

A Node lambda that reacts to kinesis messages and `POST`/`PUT`s to the MAM API and sends logs on the ELK kinesis stream.

These messages will be sent from Pluto and will relate to a new Project being created or an existing Project being updated.

Pluto will also send a message after it has created a master for an Atom, which this lambda will process too.


pluto-message-ingestion/file-config.js
```diff
@@ -0,0 +1,53 @@
...
+const parseHocon = require('hoconjs/build/hoconjs');
+
+const EnvironmentConfig = require('./environment-config');
+
+const FILE_NAME = 'media-atom-maker.private.conf';
+
+class FileConfig {
+  static read() {
+    if (EnvironmentConfig.isDev) {
+      return FileConfig._readFromDisk();
```
akash1810: Yep, could do. Only reason not to was to avoid a network call as we already have the file on disk... happy to remove though as its simpler to do the same thing in all environments.

akash1810:
(finally) tested in CODE! 🎉

teamcity.sh
```diff
@@ -7,4 +7,9 @@
 # So, lets run this file as a build step in every branch. That way, master just gets a friendly message, and other branches,
 # namely `aa-pluto-project-ingestion` can customise it.
 
-echo 'howdy partner'
+set -e
+
+pushd pluto-message-ingestion
+npm install
```
akash1810: these can be added to TC build steps once in master

cloudformation/media-atom-maker.json
```diff
@@ -365,7 +370,7 @@
                                 {
                                     "Effect": "Allow",
                                     "Action": ["sts:AssumeRole"],
-                                    "Resource": [{ "Ref": "ContentAtomArn" }, { "Fn::GetAtt" : ["LimitedUploadRole", "Arn"] }]
```
jennysivapalan: Is `"Ref": "ContentAtomArn"` not being used here?

akash1810: it is, but I've extracted into a new `CrossAccountAccessPolicy` resource that then gets applies to the relevant roles as its a little DRYer

akash1810:
> I think we should add a small readme for each lambda project. Just so it's easy to go in a remember what each lambda is aiming to achieve?

Definitely! Will add one.

pluto-message-ingestion/pluto-message-processor.js
```diff
@@ -0,0 +1,95 @@
...
+              resolve(resp);
+            })
+            .catch(err => {
+              const logDetail = {
+                status: err.status,
+                response: err.response,
+                project: project
+              };
+
+              if (err.status === 404) {
```
akash1810: A little painful...

This will occur when a pre-existing project in Pluto gets updated. Obviously, MAM doesn't know of pre-existing projects, but will still receive a message. Discarding the message effectively means MAM can only Direct Upload to projects created **after** these changes go out.

Need to chat with @fredex42 and Editorial to confirm the lifespan of a project - if a project can span a few months we maybe want to relax this restraint entirely and ingest all messages into Dynamo otherwise Editorial would have to create a new Pluto project just to Direct Upload.

fredex42: @akash1810 yes I am afraid projects can span many months.  I would suggest failing back to the "project-created" behaviour if an update message is received for a project that does not exist yet; the only alternative i can think of is a one-time hit to ingest all current projects into the system before sending it live.

akash1810: Thinking about it some more, as we're launching with Sport and News, whose projects span a fortnight at max, I'm tempted to leave this as is as and address it in the future should it become a problem when more desks adopt the Tool, as its the cheapest option.

akash1810:
merging as battle tested pretty heavily now! 😄

PR #363, opened 2017-04-04
Adding embedmode to the header

mr-mr:
In addition to the header changes, merging the large select video banner into the header when the VAM is in embed mode in composer.

image

image


public/video-ui/src/app.js
```diff
@@ -33,7 +33,9 @@ setStore(store);
 store.dispatch({
   type:       'CONFIG_RECEIVED',
   config:     Object.assign({}, extractConfigFromPage(), {
-    embeddedMode: qs.parse(location.search).embeddedMode
+    embeddedMode: qs.parse(location.search).embeddedMode,
+    isPreviewMode: qs.parse(location.search).embeddedMode === 'preview',
```
mbarton: Do these guys get used below? Seems like we're still doing logic against the embeddedMode field

PR #364, opened 2017-04-05
Select video button fix

mr-mr:
Button only appears after all published changes are published.


mbarton:
:crossed_fingers: if it works :crossed_fingers:

PR #373, opened 2017-04-07
Prompt user before reloading the page when uploading to S3

mbarton:
If the user reloads the page whilst uploading a video to S3 they will have to try again from scratch. This PR adds the classic "do you really want to reload the page" dialog as a safeguard.

### DIFF
--- a/public/video-ui/src/actions/UploadActions/s3Upload.js
+++ b/public/video-ui/src/actions/UploadActions/s3Upload.js
@@ -26,9 +26,15 @@ function uploadComplete() {
 
 export function startUpload(id, file, completeFn) {
   return dispatch => {
+    // Start prompting the user about reloading the page
+    window.onbeforeunload = () => { return false; };
+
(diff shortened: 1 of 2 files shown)


public/video-ui/src/actions/UploadActions/s3Upload.js
```diff
@@ -26,9 +26,15 @@ function uploadComplete() {
 
 export function startUpload(id, file, completeFn) {
   return dispatch => {
+    // Start prompting the user about reloading the page
+    window.onbeforeunload = () => { return false; };
```
akash1810: should we add a message?

mbarton: Browsers remove support for that. See "Custom text support removed" here: https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Browser_compatibility

PR #376, opened 2017-04-07
add field warnings to video description

Reettaphant:
- Adds a warning to the description field if a description is missing but does not prevent saving the atom. 

@akash1810 @mbarton @mr-mr 
screen shot 2017-04-07 at 16 31 00

### DIFF
--- a/public/video-ui/src/components/FormFields/TextArea.js
+++ b/public/video-ui/src/components/FormFields/TextArea.js
@@ -16,8 +16,8 @@ export default class TextArea extends React.Component {
       );
     }
 
-    const hasError = this.props.touched && this.props.errors.length > 0;
-    const hasWarning = this.props.warnings.length > 0;
+    const hasError = this.props.touched && this.props.error;
+    const hasWarning = this.props.warning;
 
     function getTextAreaClassName() {
       let postFix;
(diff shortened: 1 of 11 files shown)


public/video-ui/src/components/FormFields/TextArea.js
```diff
@@ -17,18 +17,35 @@ export default class TextArea extends React.Component {
...
+      if (hasError) {
+        postFix = "form__field--error";
+      } else if (hasWarning) {
+        postFix = "form__field--warning";
+      }
+      else {
+        postFix = "";
+      }
+
+      return "form__field " + postFix;
```
akash1810: minor: could return early to avoid the `let` and string concatenation

public/video-ui/src/components/Video/Display.js
```diff
@@ -119,11 +120,26 @@ class VideoDisplay extends React.Component {
...
+    if (notification.type === 'error') {
+      this.props.formErrorActions.updateFormErrors(notification);
+    } else {
+      this.props.formErrorActions.updateFormErrors(null);
+    }
+  }
+
+  validateDescription = (description) => {
+    if (!description) {
+      return new FieldNotification('required', 'It is recommeded you fill in this field for seo', 'warning');
```
akash1810: minor: could have [static getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static) on the `FieldNotification` class for `'required'` and `'warning'`, making this line (and others below) read something like:

```js
return new FieldNotification(FieldNotification.required, 'It is recommeded you fill in this field for seo', FieldNotification.warning);
```

PR #377, opened 2017-04-10
collect editable form field names to check for unpublished changes

Reettaphant:
Instead of hardcoding a list of all atom fields we need to compare to check for unpublished changes we compile a list of editable form fields by looking at which form fields get validated. This covers all the atom fields we check except for the assetVersion which is updated by the app when a new asset is selected. This means that we can go from hardcoding all the fields to hardcoding one. 

### DIFF
--- a/public/video-ui/src/components/ReactApp.js
+++ b/public/video-ui/src/components/ReactApp.js
@@ -37,6 +37,15 @@ class ReactApp extends React.Component {
     this.props.appActions.updateSearchTerm(searchTerm);
   };
 
+  getEditableFields = () => {
+    const allFields = this.props.checkedFormFields;
+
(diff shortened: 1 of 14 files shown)


public/video-ui/src/components/ReactApp.js
```diff
@@ -37,6 +37,18 @@ class ReactApp extends React.Component {
     this.props.appActions.updateSearchTerm(searchTerm);
   };
 
+  getEditableFields = () => {
+    const allFields = this.props.checkedFormFields;
+    if (!allFields) {
+      return [];
+    }
```
jennysivapalan: Why do we have to do this check?

Reettaphant: We don't actually so I'll remove it. I thought it might be undefined when the page loads but it is not.

PR #380, opened 2017-04-12
Route for triggering a video to be self-hosted

jennysivapalan:
Given the permission `add_self_hosted_asset` is granted users can choose to for a video to be self-hosted rather than go to YouTube. We now have to separate routes (upload to YouTube) or (self-host) which each have their own permission. So much of this PR is checking the logic that the intent of upload matches the user permissions.

screen shot 2017-04-12 at 17 04 52

I've put the trigger of the elastic transcoder into the uploader lambda as this seems to be best placed on where to make decisions on uploads. Providing people are happy with this approach I will remove the transcoder lambda in a separate PR (as it is no longer needed).

This PR does not cover showing the progress of a transcoded video. The PR is getting quite big so I will do this a separate PR. No one will have the permission to use the self-host route till we've got this and a Fastly config in place.

Before merge
- [ ]  clear out the uploads table for each environment
- [ ] apply the cloudformation changes


public/video-ui/src/components/VideoUpload/VideoUpload.js
```diff
@@ -63,30 +63,45 @@ class VideoUpload extends React.Component {
...
     }
+    else if (addAssetPermission && addSelfHostedAssetPermission) {
+      return <div> {this.renderStartUpload(false, 'Upload')} {this.renderStartUpload(true, 'Upload avoiding YouTube')} </div>;
+    }
+    else if (addAssetPermission) {
+      return this.renderStartUpload(false, 'Upload');
+    }
+    else if (addSelfHostedAssetPermission) {
+      return this.renderStartUpload(true, 'Upload avoiding YouTube');
+    }
```
jennysivapalan: Is there a nicer way to do these checks?

Reettaphant: Could you write separate functions for rendering buttons for adding youtube assets and self hosted assets that return false if correct permissions do not exist or call `this.renderStartUpload` with the correct arguments if they do exist? Then you could just do something like
`<div>{this.renderAddAsset()} {this.renderAddHostedAsset()}</div>`

app/controllers/UploadController.scala
```diff
@@ -42,45 +41,66 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
-        Ok(Json.toJson(CompleteResponse(uploadUri)))
+  def complete(id: String) = CanUploadAsset { implicit req =>
+    partRequest(id, req) { (upload, part, optionalUri) =>
+      val selfHost = upload.metadata.selfHost
+      if(!PermissionsUploadHelper.permissionAndIntentMatch(req.permissions, selfHost))
+        Unauthorized(s"User is not authorised with permissions to complete upload, self-hosted value: ${selfHost}")
+      else if (selfHost)
+        startUploadToSelfHost(upload, part, req.permissions)
+      else
+        startUpload(upload, part, optionalUri, req.permissions)
```
mbarton: Maybe call this `startUploadToYouTube` for consistency?

common/src/main/scala/com/gu/media/Permissions.scala
```diff
@@ -41,3 +51,11 @@ class MediaAtomMakerPermissionsProvider(stage: String, credsProvider: AWSCredent
     }
   }
 }
+
+object PermissionsUploadHelper {
+  def permissionAndIntentMatch(permission: Permissions, selfHosted: Boolean) : Boolean = {
```
mbarton: I'd move this into UploadController directly or rename it to be a bit clearer (something like `canPerformUpload`).

Thinking about it, I wonder if we can refactor the code in UploadController to have two clear routes from the entry point, one for YouTube and another for self hosted. Ideally we wouldn't need this method which handles both routes.

common/src/main/scala/com/gu/media/upload/actions/UploadActionHandler.scala
```diff
@@ -140,6 +148,25 @@ abstract class UploadActionHandler(store: UploadsDataStore, plutoStore: PlutoDat
     }
   }
 
+  private def transcodeToS3(fileName: String, pipelineId: String) {
+
+    val jobInput = new JobInput().withKey(fileName)
+
+    val mp4FileName = fileName + ".mp4"
```
mbarton: Do we need m3u8 as well for mobile @akash1810?

common/src/main/scala/com/gu/media/Permissions.scala
jennysivapalan: I had it in the controller but then wanted to put a test around so that we catch ourselves if we make changes to it later. So pulled it out to unit test it easier. Happy to rename it.

app/controllers/UploadController.scala
```diff
@@ -42,45 +41,66 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
-
-      Ok(Json.toJson(upload))
+      if (PermissionsUploadHelper.permissionAndIntentMatch(raw.permissions, req.selfHost)) {
+        log.info(s"Request for upload under atom ${req.atomId}. filename=${req.filename}. size=${req.size}, selfHosted=${req.selfHost}")
+        val atom = MediaAtom.fromThrift(getPreviewAtom(req.atomId))
+        val upload = buildUpload(atom, raw.user, req.size, req.selfHost)
+        table.put(upload)
+        Ok(Json.toJson(upload))
+      }
+      else Unauthorized(s"User is not authorised with permissions to upload asset, self-hosted value: ${req.selfHost}")
```
akash1810: worth also adding `${req.user.email}` similar to logs above?

app/controllers/UploadController.scala
```diff
@@ -42,45 +41,66 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
-  def complete(id: String) = CanAddAsset { implicit req =>
-    partRequest(id, req) {
-      case (upload, part, Some(uploadUri)) =>
-        partComplete(upload, part, uploadUri)
-        Ok(Json.toJson(CompleteResponse(uploadUri)))
+  def complete(id: String) = CanUploadAsset { implicit req =>
+    partRequest(id, req) { (upload, part, optionalUri) =>
+      val selfHost = upload.metadata.selfHost
+      if(!PermissionsUploadHelper.permissionAndIntentMatch(req.permissions, selfHost))
+        Unauthorized(s"User is not authorised with permissions to complete upload, self-hosted value: ${selfHost}")
```
akash1810: worth also adding `${req.user.email}` similar to logs above?

common/src/main/scala/com/gu/media/upload/actions/UploadActionHandler.scala
```diff
@@ -17,6 +19,9 @@ abstract class UploadActionHandler(store: UploadsDataStore, plutoStore: PlutoDat
 
   private val s3 = uploaderAccess.s3Client
   private val bucket = uploaderAccess.userUploadBucket
+  private val transcoderClient = uploaderAccess.transcoderClient
+  transcoderClient.setRegion(Region.getRegion(Regions.EU_WEST_1))
+  private val TRANSCODER_PRESET_ID = "1351620000001-000001" //System preset: Generic 1080p
```
akash1810: minor: should this be read from config?

common/src/main/scala/com/gu/media/upload/actions/UploadActionHandler.scala
akash1810: I think mp4 is ok initially, especially as self hosting will be a rare occurrence.

I've got a secret desire to produce a single [DASH encoding](https://en.wikipedia.org/wiki/Dynamic_Adaptive_Streaming_over_HTTP) so all clients get adaptive playback... but that's a long way away!

Although m3u8s are adaptive too, support on Android isn't great.

public/video-ui/src/components/VideoUpload/VideoUpload.js
```diff
@@ -63,30 +63,45 @@ class VideoUpload extends React.Component {
...
-      });
+      }, selfHost);
     }
   };
 
+
+
   renderButtons(uploading) {
+    const addAssetPermission = getStore().getState().config.permissions.addAsset
+    const addSelfHostedAssetPermission = getStore().getState().config.permissions.addSelfHostedAsset
```
akash1810: could be a little DRYer by:

```js
const permissions = getStore().getState().config.permissions;
const addAssetPermission = permissions.addAsset;
const addSelfHostedAssetPermission = permissions. addSelfHostedAsset;
```

common/src/main/scala/com/gu/media/upload/actions/UploadActionHandler.scala
jennysivapalan: Yeah likely - I was thinking of waiting till we see this end to end because it might be that we need to transcode it in a few formats so this might need to change to a list.

public/video-ui/src/components/VideoUpload/VideoUpload.js
jennysivapalan: Just tried it and it's much nicer, thanks!

common/src/main/scala/com/gu/media/Permissions.scala
jennysivapalan: @mbarton yes I suspect we could have two routes but I wasn't sure of UI logic. Worth re-visiting when I do the work on showing the progress?

jennysivapalan:
Thanks for the feedback @Reettaphant @mbarton @akash1810. I've made changes based on suggestions.

PR #383, opened 2017-04-13
convert dynamo templates to yaml using cfn-flip

akash1810:
Because YAML!

Also used the [`-c` option](https://github.com/awslabs/aws-cfn-template-flip)


akash1810:
Tried updating on CODE and got:

`
Error creating change set: The submitted information didn't contain changes. Submit different information to create a change set.
`

So 🎉 !

PR #384, opened 2017-04-13
bump atom maker

Reettaphant:
Because we need to release a new version of it for usage policies and the new version allows us delete code as well. 

### DIFF
--- a/app/data/AtomListStore.scala
+++ b/app/data/AtomListStore.scala
@@ -3,7 +3,6 @@ package data
 import java.time.Instant
 
 import com.gu.atom.data.PreviewDynamoDataStore
-import com.gu.contentatom.thrift.atom.media.{MediaAtom => ThriftMediaAtom}
 import com.gu.media.CapiPreviewAccess
 import model.Category.Hosted
 import model.commands.CommandExceptions.AtomDataStoreError
@@ -62,7 +61,7 @@ class CapiBackedAtomListStore(capi: CapiPreviewAccess) extends AtomListStore {
   }
 }
 
-class DynamoBackedAtomListStore(store: AtomListStore.PreviewStore) extends AtomListStore {
+class DynamoBackedAtomListStore(store: PreviewDynamoDataStore) extends AtomListStore {
   override def getAtoms(search: Option[String], limit: Option[Int]): MediaAtomList = {
     // We must filter the entire list of atoms rather than use Dynamo limit to ensure stable iteration order.
     // Without it, the front page will shuffle around when clicking the Load More button.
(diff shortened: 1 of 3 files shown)


app/data/DataStores.scala
```diff
@@ -2,29 +2,20 @@ package data
...
-import com.gu.scanamo.DynamoFormat
 import util.atom.MediaAtomImplicits
-import com.gu.scanamo.scrooge.ScroogeDynamoFormat._
 import model.commands.CommandExceptions._
 import util.AWSConfig
 
-import scala.reflect.ClassTag
-
 class DataStores(aws: AWSConfig, capi: CapiPreviewAccess) extends MediaAtomImplicits {
   import cats.syntax.either._ // appears unused but is required to make the data stores compile
```
marialivia16: This is probably not used anymore.

app/data/DataStores.scala
```diff
@@ -2,29 +2,20 @@ package data
...
 import com.gu.atom.data._
 import com.gu.atom.publish._
-import com.gu.contentatom.thrift.{Atom, AtomData}
-import com.gu.contentatom.thrift.atom.media.MediaAtom
+import com.gu.contentatom.thrift.Atom
 import com.gu.media.pluto.PlutoProjectDataStore
 import com.gu.media.upload.UploadsDataStore
 import com.gu.media.{CapiPreviewAccess, PlutoDataStore}
-import com.gu.scanamo.DynamoFormat
 import util.atom.MediaAtomImplicits
```
marialivia16: Also, not sure about this one. It seems to be unnecessary.

I've definitely had lots of issues in the past related to imports, whenever possible try to remove the unused ones (but don't count on IntelliJ to help).

app/data/AtomListStore.scala
```diff
@@ -103,7 +102,7 @@ class DynamoBackedAtomListStore(store: AtomListStore.PreviewStore) extends AtomL
 }
 
 object AtomListStore {
-  type PreviewStore = PreviewDynamoDataStore[ThriftMediaAtom]
+  type PreviewStore = PreviewDynamoDataStore
```
marialivia16: Do you actually use this in many places? It might not be worth defining a new type for something this basic now.

app/data/DataStores.scala
marialivia16: Actually, as an alternative, I can also suggest importing cats everywhere 😆 That also fixed lots of issues...

app/data/DataStores.scala
Reettaphant: It is unnecessary, but I don't think its related to this PR, I can remove it in another PR.

PR #393, opened 2017-04-19
disable add asset link if edits open

Reettaphant:
To force the user to save edits before adding assets to prevent the edits from disappearing. 

@akash1810 @jennysivapalan @mbarton

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -132,6 +132,12 @@ class VideoDisplay extends React.Component {
     return null;
   }
 
+  handleAssetClick = (e) => {
+    if (this.cannotMakeEdits()) {
+      e.preventDefault();
+    }
+  }
+
   renderEditButton = (property) => {
 
     if (this.props && this.props.editState[property]) {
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -145,7 +151,8 @@ class VideoDisplay extends React.Component {
     return <div className="video__detailbox">
       <div className="video__detailbox__header__container">
         <header className="video__detailbox__header">Video Preview</header>
-        <Link className="button" to={`/videos/${this.props.video.id}/upload`}>
+        <Link className={"button " + (this.cannotMakeEdits() ? "disabled" : "")} to={`/videos/${this.props.video.id}/upload`}
```
akash1810: would it be more semantic to disable the element rather than preventing the default click handler?

Reettaphant: I don't think I can disable because it is a link instead of a button: https://github.com/ReactTraining/react-router/issues/1082

jennysivapalan:
Have you got a screenshot /gif of how this works?

Reettaphant:
When one of the edit forms is open, the Edit Assets link is just disabled like this:
screen shot 2017-04-20 at 09 20 49
The styling is the same as with the disabled icons on the other forms.

public/video-ui/src/components/Video/Display.js
akash1810: ooh! 🤔

akash1810: could use CSS hacks too http://stackoverflow.com/a/10276157

Reettaphant: It doesn't seem to have very good browser support and the solutions that do seem more complicated than preventing default.

PR #394, opened 2017-04-19
remove unused trait

Reettaphant:
This trait seems to be unused. There is a generateHtml method in `app/model/MediaAtom.scala` which is being used and is a more sensible place for it. @akash1810 @jennysivapalan


Reettaphant:
@akash1810 I spoke too soon, we are actually only able to delete two lines of code. Maria and I think cleaning up the atom maker libraries code would allow us to delete/simplify code in mam but since its not essential, we are going look into it at some later point.

PR #395, opened 2017-04-19
Decakeify creating AWS credentials

mbarton:
The app suffered problems last week when one of the 3 PROD machines did not have permission to access the CAPI streams for publishing. This is the second of several PRs (#388) to mitigate such problems in the future.

This changes how we construct AWS credentials to no longer use the 🍰 pattern and replaces it with simple code for the three different routes through the code:

- `AwsCredentials.dev()` (when running the Play app in dev)
- `AwsCredentials.app()` (when running the Play app in PROD)
- `AwsCredentials.lambda()` (when running the lambdas)


Reettaphant:
Should we also update the Readme instructions for getting media atoms to capi?

common/src/main/scala/com/gu/media/aws/AwsCredentials.scala
```diff
@@ -0,0 +1,58 @@
+package com.gu.media.aws
+
+import com.amazonaws.auth._
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder
+import com.gu.media.Settings
+
+case class AwsCredentials(instance: AWSCredentialsProvider, crossAccount: AWSCredentialsProvider,
+                          upload: AWSCredentialsProvider)
```
Reettaphant: Would something like kinesisAccount be a better name than crossAccount since most of the time of dev we are just using media service credentials?

mbarton: I disagree. We also use kinesis streams within media-service for direct upload, so I want to make a clear distinction between what is local and cross-account

PR #396, opened 2017-04-19
hide content created by integration tests from atom list

jharewinton:


jharewinton:
@akash1810 we do already, but this gives us some extra insurance if we start running the tests on prod too. Means that user's won't notice tests while they're running, and if anything goes wrong with the teardown, we're not going to clog up the atom list with loads of test atoms

PR #399, opened 2017-04-20
Usage policy

Reettaphant:
Set a usage policy for the video. If adds are blocked, tracking policy is chosen, otherwise monetisation policy is chosen.  

I looked into putting the jar files into S3, but I don't think its worth the effort, mainly because setting up credentials so that everything works locally and on teamcity looks like too much work for the reward.

Sonatype is added into resolvers here because the atom libraries taken a long time to get to maven for some reason and I got tired of waiting for them. I think it is a good idea to leave it here because we are likely to have similar problems in the future.


build.sbt
```diff
@@ -1,13 +1,16 @@
...
 
 lazy val commonSettings = Seq(
   scalaVersion in ThisBuild := "2.11.8",
   organization in ThisBuild := "com.gu",
 
-  resolvers += "Guardian Bintray" at "https://dl.bintray.com/guardian/editorial-tools",
+  resolvers ++= Seq("Guardian Bintray" at "https://dl.bintray.com/guardian/editorial-tools",
+    "Sonatype OSS Snapshots" at "http://oss.sonatype.org/content/repositories/snapshots/",
+    Resolver.sonatypeRepo("snapshots")
+  ),
```
jennysivapalan: Perhaps we should check the `Sonatype releases` before snapshots. Quite a few Guardian repos do:
```
"Sonatype OSS" at "http://oss.sonatype.org/content/repositories/releases/",
"Sonatype OSS Snapshots" at "http://oss.sonatype.org/content/repositories/snapshots/",
```

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
```diff
@@ -0,0 +1,126 @@
...
+import com.google.api.services.youtubePartner.model._
+import com.gu.media.logging.Logging
+import com.typesafe.config.Config
+import org.joda.time.DateTime
+import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
+
+class YouTubeClaims(override val config: Config) extends YouTubeAccess with Logging {
+
+  private def createAsset(title: String, videoId: String): String = {
```
akash1810: minor: why return `String` over `Asset`?

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
```diff
@@ -0,0 +1,126 @@
...
+      .setPolicy(policy)
+      .setContentType("audiovisual")
+
+    val createdClaim = partnerClient.claims()
+      .insert(claim)
+      .setOnBehalfOfContentOwner(contentOwner)
+      .execute()
+  }
+
+  private def getPolicy(blockAds: Boolean): Policy = {
```
akash1810: minor: would `getNewPolicy` be more explicit?

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
```diff
@@ -0,0 +1,126 @@
...
+        .claimSearch()
+        .list
+        .setVideoId(videoId)
+        .execute()
+
+      if (videoClaims.getPageInfo.getTotalResults() == 0) {
+        createVideoClaim(atomId, userName, blockAds, videoId)
+      } else {
+
+        val claimsList = videoClaims.getItems().asScala
```
akash1810: does this handle the paged response?

app/data/JsonConversions.scala
```diff
@@ -118,7 +118,10 @@ object JsonConversions {
       )
   }
 
-  implicit val flagsWrites = {flags: Flags => (__ \ "legallySensitive").writeNullable[Boolean]}
+  implicit val flagsWrites = {flags: Flags =>
+    (__ \ "legallySensitive").writeNullable[Boolean] and
```
akash1810: minor: is it worth using a case class here, now there's more than one field?

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
Reettaphant: No, but the claims listed here are only the claims for a particular video so I didn't think it would be possible for one video to have so many claims we'd need a paged response - is this correct?

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
Reettaphant: Yes, this is confusing, it should return `Asset` and then get the id string from that.

app/data/JsonConversions.scala
Reettaphant: As we discussed, we should think about how we do json conversions in general when we add the missing composer fields to the atom model.

common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
akash1810: > I didn't think it would be possible for one video to have so many claims

There's no limit to the number of claims a video can have on YouTube. Every copy of a video could get a claim. That is, if random people start ripping our videos and uploading them to their own channel, we would get a claim for each one. Obviously, the more people do this, the more claims. So we're probably making an assumption that our videos don't get ripped and uploaded, which may be ok for now...?

Reettaphant: Since it is not a big thing to fix I can just do it now and then we don't have to worry about it.

PR #400, opened 2017-04-20
New delete workflow

mbarton:
Modify the UI for deleting atoms to be less confusing and inline with how Composer does permanent deletions.

- The delete button (already behind a permissions flag) is not enabled until all Composer usages have been removed
- The delete button itself now prompts you if you really want to delete the atom (rather than entering the atom id into a text box)
- Any associated YouTube videos remain but are made private

### DIFF
--- a/public/video-ui/src/components/Videos/AdvancedActions.js
+++ b/public/video-ui/src/components/Videos/AdvancedActions.js
@@ -6,22 +6,31 @@ class AdvancedActions extends React.Component {
     permissions = getStore().getState().config.permissions;
     showActions = this.permissions.deleteAtom;
 
+    state = { deleteDoubleCheck: false }
+
     renderDelete() {
         if(!this.permissions.deleteAtom) {
             return false;
         }
 
-        const doDelete = () => {
-            const result = prompt("Enter the atom ID to confirm deletion (it can be copied from the URL)");
+        const disabled = this.props.usage.length > 0;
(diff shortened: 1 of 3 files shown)


public/video-ui/src/components/Videos/AdvancedActions.js
```diff
@@ -6,22 +6,31 @@ class AdvancedActions extends React.Component {
...
+                this.setState({ deleteDoubleCheck: true });
             }
         };
 
         return <li className="action-list__item">
-            <button className="btn label__expired action-list__button" onClick={doDelete}>DELETE</button>
-            <span className="right">Usages will not be replaced and the video will remain on YouTube</span>
+            <button className="btn label__expired action-list__button" onClick={doDelete} disabled={disabled}>
+                {deleteMsg}
+            </button>
```
Reettaphant: Do we assume that people know why the button is disabled?

mbarton: Good point. I've updated the message on the left to explain why the button is disabled:

```
All usages of the atom must be removed before deletion
```

akash1810:
could you add a gif please?

akash1810:
> The delete button (already behind a permissions flag) is not enabled until all Composer usages have been removed

Should we make this a server side check too? (in a subsequent PR).

mbarton:
@akash1810 #408 for server-side check

PR #401, opened 2017-04-20
convert cfn template to yaml

akash1810:
also removes the linting of cfn templates as I couldn't find a decent yaml linter. I had considered using the [js sdk](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudFormation.html#validateTemplate-property) but bizarrely it requires aws credentials even to validate a template.


akash1810:
Closing as didn't work and is now out of date. Still worth doing, but will complete as a new PR to avoid having a WIP one open for too long.

PR #403, opened 2017-04-21
run tests on prod

jharewinton:
will require download of new config file from S3, and TeamCity job changes before prod tests start

### DIFF
--- a/integration-tests/int-test-runner.sh
+++ b/integration-tests/int-test-runner.sh
@@ -6,7 +6,7 @@ if [ $STATUS -eq 1 ]
 then
   if [ "$INT_TEST_TARGET" == "PROD" ]
   then
-    curl -X POST --data-urlencode 'payload={"text": "Media Atom Maker integration tests have failed on PRODUCTION"}' ${SLACK_URL}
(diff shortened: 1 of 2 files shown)


integration-tests/int-test-runner.sh
```diff
@@ -4,6 +4,11 @@ export STATUS=$?
 
 if [ $STATUS -eq 1 ]
 then
-  curl -X POST --data-urlencode 'payload={"text": "Media Atom Maker integration tests have failed on CODE"}' ${SLACK_URL}
+  if [ "$INT_TEST_TARGET" == "PROD" ]
+  then
+    curl -X POST --data-urlencode 'payload={"text": "<!here> Media Atom Maker integration tests have failed on PRODUCTION"}' ${SLACK_URL}
```
akash1810: Does TC provide a test result file that we can link to to find which test failed? And/or should we add a link to the slack message to the TC project for ease?

jharewinton: I'll add in link to the project. Could do a test result file, but considering these are 'monitoring' tests, they're running every 15 mins, so could end up filling up team city with artefacts quite quickly. From the project page, can get the failing build log really quickly

PR #413, opened 2017-05-02
Better upload errors client-side

mbarton:
Improves the error handling code for direct uploads to log the status code and response text where available.

Also breaks down 'error uploading video' into errors at each specific part request.

### DIFF
--- a/public/video-ui/src/actions/UploadActions/getUploads.js
+++ b/public/video-ui/src/actions/UploadActions/getUploads.js
@@ -1,5 +1,5 @@
 import {UploadsApi} from '../../services/UploadsApi';
-import Logger from '../../logger';
+import {errorDetails} from '../../util/errorDetails';
 
 function runningUploads(uploads) {
   return {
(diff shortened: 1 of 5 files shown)


public/video-ui/src/actions/UploadActions/getUploads.js
```diff
@@ -10,8 +10,6 @@ function runningUploads(uploads) {
 }
 
 function errorGettingUploads(error) {
-  Logger.error(error);
```
Reettaphant: Why is this line removed?

mbarton: We now print useful info about the error into the big red UI box rather than on the console

Reettaphant: Sounds sensible. Would be useful to do this for other error reporting as well later?

mbarton: Yeah I've held off doing that just to keep the PR small

PR #414, opened 2017-05-02
Space for new fields

Reettaphant:
Makes a number of changes to the UI to make space for new composer fields and to make the workflow for updating composer pages more sensible: 
- Merge youtube metadata and metadata forms into one giant form 
- Move delete button to the header away from its own section 
- Move create composer page button to the header
- Remove update composer page button. Make updating composer pages part of the publish action
- Remove add assets button text to make look the same as other edit buttons and add a tooltip to the button with the add assets text

The big edit atom data form could probably do with some styling to make it look more structured, any ideas are welcomed. 

screen shot 2017-05-02 at 11 25 08

screen shot 2017-05-02 at 12 11 07

@akash1810 @mbarton


public/video-ui/src/reducers/editStateReducer.js
```diff
@@ -1,10 +1,7 @@
-export default function video(state = {
-    metadataEditable: false,
-    youtubeEditable: false
-  }, action) {
+export default function video(state = false, action) {
```
akash1810: this is a curious change - `state` going from an object to a bool 🤔

public/video-ui/src/components/ManagedForm/ManagedForm.js
```diff
@@ -23,6 +24,13 @@ export class ManagedForm extends React.Component {
     }
   }
 
+  getFormClass = () => {
+    if (React.Children.toArray(this.props.children).some(child => child.type.componentType === 'managedField')) {
```
Reettaphant: For some reason I made the type of  `managedSection` `managedField`so I'm going to change that. But I'm not sure if there is a simpler way of doing this, `managedForm` needs to know about the type of its children to know its css class.

public/video-ui/src/components/ManagedForm/ManagedSection.js
```diff
@@ -0,0 +1,34 @@
...
+    children: PropTypes.oneOfType([
+        PropTypes.element,
+        PropTypes.arrayOf(PropTypes.element)
+    ]),
+    editable: PropTypes.bool,
+    updateErrors: PropTypes.func
+  };
+
+  static get componentType () {
+    return "managedField";
```
Reettaphant: Yes but I should probably get rid of them.

public/video-ui/src/reducers/editStateReducer.js
Reettaphant: Because we used to have two forms to track so they were in an object but now we just have to track the edit state of one form so it might as well be a bool.

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -0,0 +1,135 @@
...
+import {videoCategories} from '../../constants/videoCategories';
+import {privacyStates} from '../../constants/privacyStates';
+
+class VideoData extends React.Component {
+
+  hasCategories = () => this.props.youtube.categories.length !== 0;
+  hasChannels = () => this.props.youtube.channels.length !== 0;
+
+  componentWillMount() {
+    if (! this.hasCategories()) {
```
Reettaphant: I think its done because if people navigate around the app without refreshing the page the youtube channels and categories will remain stored in the properties so there is no need to fetch them again.

PR #415, opened 2017-05-02
Fix automatic panda renegotiation

mbarton:
Fixes a couple of bugs that were stopping [automatic panda session renegotiation](https://github.com/guardian/media-atom-maker/blob/bf1e6bf7e1950ec993e37079b2f69176145b6e65/public/video-ui/src/services/pandaReqwest.js#L17) from working:

- [Correct status code check](https://github.com/guardian/media-atom-maker/blob/bf1e6bf7e1950ec993e37079b2f69176145b6e65/public/video-ui/src/services/pandaReqwest.js#L13)
- Upgrade panda-hmac to avoid https://github.com/guardian/panda-hmac/pull/1

I've also taken the opportunity to disallow HMAC access to all end-points by default, enabling it only for those required by [CDS](https://github.com/guardian/content_delivery_system/blob/master/CDS/scripts/js_utils/media-atom/media-atom.js#L10) and the upload lambda.

### DIFF
--- a/public/video-ui/src/services/pandaReqwest.js
+++ b/public/video-ui/src/services/pandaReqwest.js
@@ -9,7 +9,7 @@ function poll(reqwestBody, timeout) {
   function makeRequest(resolve, reject) {
     reqwest(reqwestBody)
       .then(response => resolve(response))
-      .catch(err => {
+      .fail(err => {
         if (err.status === 419) {
           const store = getStore();
           const reauthUrl = store.getState().config.reauthUrl;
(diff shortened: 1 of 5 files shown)


public/video-ui/src/services/pandaReqwest.js
```diff
@@ -9,8 +9,8 @@ function poll(reqwestBody, timeout) {
   function makeRequest(resolve, reject) {
     reqwest(reqwestBody)
       .then(response => resolve(response))
-      .fail(err => {
-        if (err === 419) {
+      .catch(err => {
```
currysoup: https://github.com/ded/reqwest

The reqwest github page seems to suggest `fail` is the correct function to use here?

currysoup: The `Promise` at the bottom is `.catch` because it's a plain JS `Promise`. This is why we used `Q` in tagmanager, because it presents a consistent API for handling promises with `Reqwest`

mbarton: They both work (verified in a debugger). Happy to put it back though

PR #417, opened 2017-05-03
Add react router

Reettaphant:
- Add react router redux to clear video and edit form state when the route changes so we don't need do when components dismount.
- Currently clears the state when the path is different, not when the video ids are different. Because it looks fine as it is, I didn't think it was worth the trouble to add this in although am happy to do so. 

@akash1810 @mbarton @shaundillon


public/video-ui/src/util/storeMiddleware.js
```diff
@@ -0,0 +1,25 @@
+export const storeMiddleware = ({dispatch, getState}) => (next) => (action) => {
+
+  const prevState = getState();
+  let result = next(action);
```
akash1810: do we need this assignment?

package.json
```diff
@@ -59,6 +59,7 @@
     "react-dom": "^15.3.2",
     "react-redux": "^5.0.1",
     "react-router": "^3.0.0",
+    "react-router-redux": "^4.0.8",
```
akash1810: could we also update the [`yarn.lock` file](https://yarnpkg.com/en/docs/yarn-lock)?

Reettaphant: 👍

PR #423, opened 2017-05-04
update yarn.lock

akash1810:
its a really exciting diff adding react-tooltip


akash1810:
> Does this happen if we use npm install rather than yarn install?

Yep. Ideally, new modules get installed with `yarn`. We can also make CI only fail if a module has been installed but `yarn.lock` doesn't show it I think. May be too much though.

PR #428, opened 2017-05-05
pre-commit hook... back with a vengence 😈

akash1810:
Use `lint-staged` and `husky` for linting as [recommended](https://github.com/prettier/prettier#pre-commit-hook-for-changed-files). This change also only runs the hook when js files are changed within `public/video-ui/src/`.

If this doesn't work for all of us, I'm happy to stop pursuing it.

### DIFF
--- a/public/video-ui/src/services/pandaReqwest.js
+++ b/public/video-ui/src/services/pandaReqwest.js
@@ -37,10 +37,11 @@ function poll(reqwestBody, timeout) {
 export function pandaReqwest(reqwestBody, timeout = 0) {
   const payload = Object.assign({ method: 'get' }, reqwestBody);
 
-  if(payload.data) {
+  if (payload.data) {
     payload.contentType = payload.contentType || 'application/json';
 
-    if(payload.contentType === 'application/json' && typeof payload.data === 'object') {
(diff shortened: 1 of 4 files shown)


public/video-ui/src/services/pandaReqwest.js
```diff
@@ -37,10 +37,11 @@ function poll(reqwestBody, timeout) {
 export function pandaReqwest(reqwestBody, timeout = 0) {
   const payload = Object.assign({ method: 'get' }, reqwestBody);
 
-  if(payload.data) {
+  if (payload.data) {
     payload.contentType = payload.contentType || 'application/json';
 
-    if(payload.contentType === 'application/json' && typeof payload.data === 'object') {
+    // prettier-ignore
```
akash1810: otherwise prettier goes crazy and turns this into

```js
if (
    payload.contentType === 'application/json' &&
    typeof payload.data === 'object'
) {
    payload.data = JSON.stringify(payload.data);
}
```

akash1810:
Merging. Per Slack convo, lets give this a week to see how it feels. If its more pain than its worth, we will remove it entirely.

PR #432, opened 2017-05-09
fix riffraff warnings

akash1810:
makes this less scary!

image


sihil:
I've merged https://github.com/guardian/riff-raff/pull/436 so you should be able to revert this now. Otherwise you'll get deprecation warnings that mirror these!

PR #436, opened 2017-05-10
Use consistent reads for uploads

mbarton:
User reported issue with direct upload:

picture 412

The investigation to this is still on-going. This PR kicks things off:

- Uses consistent reads on the uploads table
    - The code does currently expect to be able to read out what was previously written in (by ID)

- Add low level logging for the JSON in and out of the uploads Kinesis stream
    - The lambda appeared to run the `DeleteParts` command rather than the expected `UploadPartToYouTube` command

### DIFF
--- a/uploader/src/main/scala/com/gu/media/UploadActionsLambda.scala
+++ b/uploader/src/main/scala/com/gu/media/UploadActionsLambda.scala
@@ -22,7 +22,7 @@ class UploadActionsLambda extends RequestHandler[KinesisEvent, Unit]
   with SESSettings
   with KinesisAccess
   with ElasticTranscodeAccess {
-  
+
   val store = new UploadsDataStore(this)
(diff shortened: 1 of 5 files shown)


uploader/src/main/scala/com/gu/media/UploadActionsLambda.scala
```diff
@@ -43,8 +43,11 @@ class UploadActionsLambda extends RequestHandler[KinesisEvent, Unit]
     val record = input.getRecords.get(0)
     val data = new String(record.getKinesis.getData.array(), "UTF-8")
 
+    log.info(s"Received JSON from Kinesis: $data")
```
akash1810: minor: could send `$data` as the second arg to make it searchable in ELK

mbarton: oh I didn't know that, cool! I'll remember for next time

PR #440, opened 2017-05-11
client side pluto project selection

akash1810:
Annoyingly I've realised that [CDS](https://github.com/guardian/content_delivery_system/blob/master/CDS/scripts/js_utils/media-atom/media-atom.js#L119-L140) is currently setting the Pluto Master ID as the `plutoProjectId` in the Atom.

I've raised [this PR](https://github.com/guardian/content-atom/pull/86) to add more fields to the thrift definition on the assumption we want to push these links to capi (and because its the cheapest option).

screen shot 2017-05-11 at 14 10 20

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -164,7 +164,7 @@ class Api2 (override val stores: DataStores, conf: Configuration, override val a
         previewDataStore.getAtom(upload.atomId) match {
           case Right(atom) => {
             val mediaAtom = MediaAtom.fromThrift(atom)
-            mediaAtom.plutoProjectId match {
+            mediaAtom.plutoData match {
               case None => acc ++ Map(upload.atomId -> mediaAtom)
               case Some(string) => acc
             }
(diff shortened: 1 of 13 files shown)


Reettaphant:
What happens if Pluto is down and we cannot fetch the ids? Also, should the project ids be added to the `VideoPlutoAdd` component as well?

akash1810:
> What happens if Pluto is down and we cannot fetch the ids?

Good point! The field shouldn't be required.

> Also, should the project ids be added to the VideoPlutoAdd component as well?

Quite possibly, I forgot to look at it..

mbarton:
Does it make more sense to have this field on the direct upload page rather than the main one?

It would make it more likely that people select one if we make it a pre-requisite for uploading. Though it would still need to work if the project didn't exist.

akash1810:
> Does it make more sense to have this field on the direct upload page rather than the main one?

Yes. However I'd like to do this as a separate PR because these fields should also go there:
- YouTube Category
- YouTube Channel
- Privacy Status

Reettaphant:
I think putting youtube specific fields to the uploads page is a very good idea. But if we do it, we'd also need to make sure that we can save atoms without these fields and that there are appropriate checks in place before publishing the atom.

PR #435, opened 2017-05-10
Direct upload step functions skeleton

mbarton:
Adds a skeleton for how AWS step functions could manage video uploads. At the moment it simply triggers on an upload, waits for the video chunks in S3 and then fires off the other lambdas (all of which just do dummy logging). The existing video pipeline continues untouched alongside.

The step functions are deployed by Riff-Raff as a separate cloud-formation stack, alongside the main app. There is also a DEV stack and a script to update it.

The [uploader README](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions/uploader/README.md) goes into a bit more detail.

PR #446, opened 2017-05-12
Better error handling when YouTube uploads 503

mbarton:
The integration tests have failed a few times now with the "infamous Kinesis eternal retry of death" caused by YouTube 503ing at us when we try to upload a video.

This PR modifies the behaviour to catch the error and pass over the chunk. This is not full failure recovery (that will be addressed under #435 and further PRs)

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubeUploader.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubeUploader.scala
@@ -4,8 +4,9 @@ import java.io.InputStream
 
 import com.gu.media.upload.actions.UploaderAccess
 import com.gu.media.util.InputStreamRequestBody
+import com.gu.media.youtube.YouTubeUploader.{MoveToNextChunk, UploadError, VideoFullyUpload}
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/media/youtube/YouTubeUploader.scala
```diff
@@ -65,13 +66,35 @@ class YouTubeUploader(aws: UploaderAccess, youTube: YouTubeAccess) {
...
-      Some((Json.parse(str) \ "id").as[String])
+  private def parseResult(result: String): YouTubeUploader.Result = {
+    if(result.isEmpty) {
+      MoveToNextChunk
     } else {
-      // next part
-      None
+      val json = Json.parse(result)
+
+      ((json \ "id").asOpt[String], (json \ "error").asOpt[JsObject]) match {
```
akash1810: minor: could we use a case class with read/writes for this?

mbarton: could do but feels overkill

PR #447, opened 2017-05-13
update docs

akash1810:
updated docs can be seen here - https://github.com/guardian/media-atom-maker/tree/aa-docs/docs


docs/08-other.md
```diff
@@ -0,0 +1,6 @@
+# Other
+
```
Reettaphant: If someone new to the project is looking for this information, they may not be able to find under this heading. Could we link to this from elsewhere?

docs/03-running.md
```diff
@@ -0,0 +1,29 @@
...
+To run the Angular App:
+
+- Generate javascript files by running `gulp watch`
+- Start sbt by running `./sbt app/run`
+- Open the app by going [here](https://video.local.dev-gutools.co.uk/atoms)
+
+## React App
+There are two ways to run the React App.
+
+Once its running, open the app by going [here](https://video.local.dev-gutools.co.uk/videos)
```
Reettaphant: I'd make this sentence the first of the section, then the instructions for two ways of running the app would come right after the stating that there are two ways of running the app.

README.md
```diff
@@ -1,56 +1,4 @@
...
-0. Modify `AwsCredentials.scala` to set the `crossAccount` field to `composer` profile credentials.
-
-## Testing
-
-To run unit tests `sbt test`
-
-To run the blackbox integration tests against a deployed environment:
-0. Download the config from S3 `aws s3 cp s3://atom-maker-conf/TEST/media-atom-maker-integration-test.private.conf integration-tests/src/test/conf/media-atom-maker-integration-test.private.conf`
-0. Run `sbt integrationTests/test`
+Follow the [docs](./docs) to get setup, run the app, test and deploy changes.
```
Reettaphant: `setup`, 'run`, `test` and `deploy` could be links as well.

docs/08-other.md
akash1810: I've moved this into the `running` file

PR #448, opened 2017-05-15
convert cfn template to yaml

akash1810:
https://github.com/guardian/media-atom-maker/pull/401 take II.

The template of #401 was generated using [cfn-flip](https://github.com/awslabs/aws-cfn-template-flip) with the `--clean` option, which resulted in a big cloudformation diff. 

Omitting using `--clean` produced this, 0 change, template.

image


Reettaphant:
The build seems to failing because the json templates are still referenced in `pluto-message-ingestion\package.json` but otherwise 👍

akash1810:
Ah thanks @Reettaphant! Updated now.

PR #449, opened 2017-05-15
one dynamo template to rule them all

akash1810:
Merge the CODE and PROD templates into a single one, using Mappings.

WIP PR as the AWS console is showing the tables will be deleted and recreated - have raised a support ticket for confirmation of whether we'll encounter data loss.


akash1810:
Amazon have responded and have confirmed that changing Logical Names in the template causes resource deletion and creation, which would result in data loss. This is a bit sad and means we have to continue to maintain a template per stage.

I'd propose adding new tables into the main template. We should also be able to painlessly move a few of the newer tables there too, such as:
- UploadTrackingAtomMakerCodeTable
- ManualPlutoAtomMakerCodeTable

As they have 0 entries at the moment.

PR #450, opened 2017-05-15
only talk to YouTube Partner API when preview != published

akash1810:
to reduce the probability of being rate limited by YouTube

Also changes the template of the YouTube Asset title, removing usernames because 🔒  and setting a CustomID to the Atom ID. This CustomID comes through to various YouTube reports and is useful for grouping videos up to their Atom.

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -27,25 +27,22 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
   def process(): T = {
     log.info(s"Request to publish atom $id")
 
-    val thriftAtom = getPreviewAtom(id)
-    val atom = MediaAtom.fromThrift(thriftAtom)
+    val thriftPreviewAtom = getPreviewAtom(id)
+    val previewAtom = MediaAtom.fromThrift(thriftPreviewAtom)
 
-    if(atom.privacyStatus.contains(PrivacyStatus.Private)) {
-      log.error(s"Unable to publish atom ${atom.id}, privacy status is set to private")
+    if(previewAtom.privacyStatus.contains(PrivacyStatus.Private)) {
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/media/youtube/YouTubeClaims.scala
```diff
@@ -110,7 +111,7 @@ class YouTubeClaims(override val config: Config) extends YouTubeAccess with Logg
     }
   }
 
-  def createOrUpdateClaim(atomId: String, videoId: String, userName: String, blockAds: Boolean): Claim = {
+  def createOrUpdateClaim(atomId: String, videoId: String, blockAds: Boolean): Claim = {
```
akash1810: I wanted to make `atomId` a `MediaAtom` rather than a `String` but that meant moving the models which is probably too big for this PR...

app/model/commands/PublishAtomCommand.scala
```diff
@@ -92,20 +89,68 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
...
 
-  private def updateYouTube(atom: MediaAtom, asset: Asset): Unit = {
+  private def updateYouTube(previewAtom: MediaAtom, asset: Asset): MediaAtom = {
+    updateYoutubeMetadata(previewAtom, asset)
+    updateYoutubeThumbnail(previewAtom, asset)
+    createOrUpdateYoutubeClaim(previewAtom, asset)
+  }
+
+  private def createOrUpdateYoutubeClaim(previewAtom: MediaAtom, asset: Asset): MediaAtom = {
+    if (previewAtom.blockAds.isEmpty) {
```
Reettaphant: When would the blockAds field not be set? If its only for unpublished atoms, don't we check for that underneath?

akash1810: `previewAtom.blockAds.isEmpty` is empty after we initial create an atom. Doing this here saves a database call as we know that it also hasn't been published.

PR #451, opened 2017-05-17
remove script

akash1810:
as its redundant as we use a shared stack


Reettaphant:
We might still want to create temporary separate stacks so that cloudformation changes of work in that is in progress don't override dev properties, but I guess that can be done via the console as well.

akash1810:
@Reettaphant yep! I also forgot there's another reason for removing this script... it won't work with Janus credentials because we [create AccessKeys in the DEV stack](https://github.com/guardian/media-atom-maker/blob/master/cloudformation/media-atom-maker-dev.yml#L99) which Janus credentials don't have permission to.

Reettaphant:
Yes, that is an excellent reason for getting rid of it.

PR #452, opened 2017-05-17
don't send "- video" to YouTube's title

akash1810:
Editorial add "- video" as it helps with SEO on platform. This just isn't needed on a YouTube video title because its a video platform.


akash1810:
@gidsg you may like this! 😄

PR #453, opened 2017-05-17
Add new endpoint to be used by Workflow to create a new atom

marialivia16:
- [x] The current endpoint has too many required fields, we need a more relaxed one.
- [x] Allow CORS.

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -73,6 +73,15 @@ class Api2 (override val stores: DataStores, conf: Configuration, override val a
     }
   }
 
+  def createWorkflowMediaAtom = APIAuthAction { implicit req =>
+    parse(req) { title: String =>
+      val command = CreateWorkflowAtomCommand(title, stores, req.user)
+      val atom = command.process()
+
+      Created(Json.toJson(atom)).withHeaders("Location" -> atomUrl(atom.id))
+    }
+  }
+
   def putMediaAtom(id: String) = APIHMACAuthAction { implicit req =>
     parse(req) { atom: MediaAtom =>
       val command = UpdateAtomCommand(id, atom, stores, req.user)
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -1,6 +1,5 @@
 package controllers
 
-import _root_.util.AWSConfig
 import com.gu.atom.play.AtomAPIActions
 import com.gu.media.MediaAtomMakerPermissionsProvider
 import com.gu.media.logging.Logging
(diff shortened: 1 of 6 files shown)


app/controllers/Api2.scala
```diff
@@ -73,6 +81,16 @@ class Api2 (override val stores: DataStores, conf: Configuration, override val a
     }
   }
 
+  def createWorkflowMediaAtom = CORSable(awsConfig.workflowUrl) {
+      APIAuthAction { implicit req =>
+        parse(req) { wma: WorkflowMediaAtom =>
+          val command = CreateWorkflowAtomCommand(wma.title, stores, req.user)
```
akash1810: should we pass `wma` into `CreateWorkflowAtomCommand` to remain strongly typed?

app/model/commands/CreateWorkflowAtomCommand.scala
```diff
@@ -0,0 +1,89 @@
...
+import scala.util.{Failure, Success}
+
+case class CreateWorkflowAtomCommand(workflowMediaAtom: WorkflowMediaAtom, override val stores: DataStores, user: PandaUser)
+
+  extends Command with Logging {
+
+  type T = MediaAtom
+
+  def process(): MediaAtom = {
+    val atomId = randomUUID().toString
```
mchv: Indirectly related,  but I chatted with @tomrf1 and @shaundillon about this. We don't want the `id` to be random `UUID` but a human readable path registered through the [`path-manager`](https://github.com/guardian/path-manager), for instance for an explainer: `/atom/explainer/what-you-should-know-about-obama`.

I think @tomrf1 has started to work on extending `path-manager` to provide this.

LATaylor-guardian: We don't currently do this for any of our atoms...

tomrf1: It's on my todo list

akash1810: Interesting - is there any particular reason for changing?

mchv: As I just discussed with @LATaylor-guardian about this,  please find below the 4 reasons:

- Content API consistency: currently the atom with `id`:  `acb90e30-85a7-4d80-8dfa-a4d1f3fae642` is accessible on `https://content.guardianapis.com/atom/explainer/acb90e30-85a7-4d80-8dfa-a4d1f3fae642` on content API. This does not follow the rule that any item should be accessed by its `id`. Worse than than that, if you use the `atoms` endpoint of the `API` you can't find how to navigate to this `item` in particular, you have to know how to construct the `path` yourself, which is a leaky abstraction.

- Frontend consistency: Without registering the path to `path-manager`, nothing prevent a page to be registered with the same `id` than an atom. This is very unlikely but as some atoms may be rendered alone (without an article), it is easier to ensure it will never happened.  

- Nice urls: Some atoms could be rendered alone by frontend, and will use their `id` as the path (with a potential prefix such as `embed`). Having nice urls exactly like on article have several benefits. 

- Ease of debugging: Having the `type` and a summary of the content of the atom is very useful when debugging. it is easier to remember.

Note there is no plan of migrating existing atoms, only new atoms created when the code will be ready we have a `better` id.

akash1810: > Nice urls: Some atoms could be rendered alone by frontend, and will use their id as the path (with a potential prefix such as `embed`). Having nice urls exactly like on article have several benefits.

Oooh! I like the sound of this. Currently, `media-atoms` can be associated with a video type Composer page to achieve this. For example here - https://www.theguardian.com/football/video/2017/may/19/new-wembley-arch-lit-up-team-colours-10th-anniversary-stadium-timelapse-video.

Also, does this slightly break the atom concept of being a part of content rather than content in their own right?

PR #459, opened 2017-05-24
Step functions (actual implementation)

mbarton:
Step functions part 2/3! Sorry for this being a bit of a big PR, it breaks down as follows:

The code for each `UploadAction` has been split out into classes for re-use between both the kinesis powered pipeline and the step functions powered pipeline:

- [PlutoUploadActions](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/common/src/main/scala/com/gu/media/upload/PlutoUploadActions.scala)
- [S3UploadActions](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/common/src/main/scala/com/gu/media/upload/S3UploadActions.scala)
- [TranscoderUploadActions](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/common/src/main/scala/com/gu/media/upload/TranscodeUploadActions.scala)
- [YouTubeUploader](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/common/src/main/scala/com/gu/media/youtube/YouTubeUploader.scala#L56)
- [AddAssetActions](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/uploader/src/main/scala/com/gu/media/AddAssetActions.scala)

These actions are called from both implementations:

- Kinesis [UploadActionHandler](https://github.com/guardian/media-atom-maker/blob/mbarton/step-functions-publish/common/src/main/scala/com/gu/media/upload/actions/UploadActionHandler.scala#L20)

- The various [step function lambdas](https://github.com/guardian/media-atom-maker/tree/mbarton/step-functions-publish/uploader/src/main/scala/com/gu/media/upload)

I have also added a checkbox to the assets pages that allows you to select the step functions route rather than the existing kinesis route (which is still the default). You'll only see the checkbox if you have "self hosted" permission to avoid normal users seeing it.

NB: if you chose the step functions route in dev the asset will not be added to the atom (it cannot do the HMAC request). This will be addressed in a follow up PR.

Once this is merged, another PR (part 3/3!) will remove the previous implementation and tidy up the code accordingly.


common/src/main/scala/com/gu/media/upload/S3UploadActions.scala
```diff
@@ -0,0 +1,85 @@
...
+        s3.deleteObject(upload.metadata.bucket, part.key)
+      } catch {
+        case NonFatal(err) =>
+          // if we can't delete it, no problem. the bucket policy will remove it in time
+          log.warn(s"Unable to delete part $part: $err")
+      }
+    }
+  }
+
+  private def copyPart(multipartId: String, bucket: String, part: Int, source: String, destination: String): PartETag = {
```
akash1810: `PartETag` 🎉  🏷

PR #462, opened 2017-05-26
Fix step function uploading

mbarton:
Follow up to #459 after testing on CODE.

- Add permission for pipeline lambdas to access the upload tracking table and manual pluto table
- Bump up lambda timeout to 5 minutes
    - Addresses timeout when doing a multipart copy of large files
    - May also make things faster by giving us more compute power on the lambda
- Remove UI filtering of uploads
    - This was causing failed uploads to disappear from the UI
- `GetChunkFromS3` lambda writes to the progress table
    - As above, this made the upload disappear

### DIFF
--- a/uploader/src/main/resources/cfn-template.yaml
+++ b/uploader/src/main/resources/cfn-template.yaml
@@ -26,8 +26,11 @@ Parameters:
   UploadBucketName:
     Description: "The S3 bucket where videos are uploaded to"
     Type: "String"
-  UploadProgressTable:
+  UploadTrackingTable:
     Description: "The Dynamo table to store progress"
+    Type: "String"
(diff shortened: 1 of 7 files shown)


uploader/src/main/resources/cfn-template.yaml
```diff
@@ -26,9 +26,12 @@ Parameters:
   UploadBucketName:
     Description: "The S3 bucket where videos are uploaded to"
     Type: "String"
-  UploadProgressTable:
+  UploadTrackingTable:
```
akash1810: any reason for this name change?

mbarton: For consistency with the main cloudformation template

PR #463, opened 2017-05-26
make Publish command async

akash1810:
Calls YT APIs in a series of Futures for ⚡️

### DIFF
--- a/app/controllers/Api2.scala
+++ b/app/controllers/Api2.scala
@@ -16,6 +16,9 @@ import util.atom.MediaAtomImplicits
 import play.api.libs.json._
 import play.api.mvc._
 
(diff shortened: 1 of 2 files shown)


app/controllers/Api2.scala
```diff
@@ -16,6 +16,9 @@ import util.atom.MediaAtomImplicits
 import play.api.libs.json._
 import play.api.mvc._
 
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
```
mbarton: Should we use the Play execution context instead? 

https://www.playframework.com/documentation/2.5.x/ThreadPools#Using-the-default-thread-pool

PR #464, opened 2017-05-30
content bundling

akash1810:
Commercial want to build content bundles of YouTube videos based on video tags. They have [supplied a mapping](https://docs.google.com/spreadsheets/d/1t0rVeddEaiYhp4yHyY856j9nJh49eJltoisD9XRtUUc/edit#gid=0) of Editorial tags -> DfP content bundles. This PR will map over Editorial tags and add any content bundle matches.

Note:
- these new content bundle tags only touch YouTube and do not get pushed to capi
- this is temporary, the real solution would be to map over Tag Manager section tags

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/package.scala
+++ b/common/src/main/scala/com/gu/media/youtube/package.scala
@@ -66,13 +66,12 @@ package object youtube {
     def clean(): YouTubeMetadataUpdate = {
       // Editorial add "- video" for on platform SEO, but it isn't needed on a YouTube video title as its a video platform
       val cleanTitle = this.title.map(_.replaceAll(" (-|–) video$", ""))
+      this.copy(title = cleanTitle)
+    }
 
+    def withContentBundleTags(): YouTubeMetadataUpdate = {
       val contentBundledTags = getContentBundlingTags()
-
-      this.copy (
-        title = cleanTitle,
-        tags = contentBundledTags
-      )
+      this.copy(tags = contentBundledTags)
     }
 
     private def getContentBundlingTags(): List[String] = {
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/media/youtube/package.scala
```diff
@@ -56,6 +55,75 @@ package object youtube {
...
 
+  case class YouTubeMetadataUpdate(
+    title: Option[String],
+    categoryId: Option[String],
+    description: Option[String],
+    tags: List[String],
+    license: Option[String],
+    privacyStatus: Option[String]
+  ) {
+    def clean(): YouTubeMetadataUpdate = {
```
mbarton: Minor - not sure `clean` implies it will also add tags

common/src/main/scala/com/gu/media/youtube/package.scala
```diff
@@ -56,6 +55,75 @@ package object youtube {
...
+        "business" -> "gdnpfpbusiness",
+        "money" -> "gdnpfpmoney",
+        "fashion" -> "gdnpfpfashion",
+        "environment" -> "gdnpfpenvironment",
+        "technology" -> "gdnpfptechnology",
+        "travel" -> "gdnpfptravel",
+        "science" -> "gdnpfpscience"
+      )
+
+      val tagsAndContentBundleTags: ListBuffer[String] = new ListBuffer[String]()
```
mbarton: Could avoid the mutable list?

```scala
this.tags.flatMap { tag =>
    contentBundlingMap.get(tag.toLowerCase()) match {
        case Some(additionalTag) => List(tag, additionalTag)
        case None => List(tag)
    }
}
```

I ❤️ Scala and how much additional list allocation is required for `flatMap`

common/src/main/scala/com/gu/media/youtube/package.scala
akash1810: Yeah, I wasn't overly pleased with it. I've updated to use method chaining which is slightly better.

common/src/main/scala/com/gu/media/youtube/package.scala
```diff
@@ -56,6 +55,69 @@ package object youtube {
...
+      this.copy(title = cleanTitle)
+    }
+
+    def withContentBundleTags(): YouTubeMetadataUpdate = {
+      val contentBundledTags = getContentBundlingTags()
+      this.copy(tags = contentBundledTags)
+    }
+
+    private def getContentBundlingTags(): List[String] = {
+      val contentBundlingMap: Map[String, String] = Map (
```
mbarton: very minor: these could be a field on the object

PR #465, opened 2017-05-30
Switch to step functions!

mbarton:
This PR switches the video upload pipeline to only use step functions.

The `list()` end-point for uploads has been reworked to read the status of the step functions for a particular atom. The front-end now merely displays the `UploadStatus` information it is given.

If the invocation fails, an error is displayed on the upload page asking the user to retry the upload. It disappears after 10 minutes - the result is however still viewable from the AWS console.

Additional goodies:

- Pass the key to the `credentials` end-point as a GET parameter rather than a header (cleaner)
- Remove the `complete` end-point - no longer required
- Fix problem where the new IDs were too long for the STS role session name
- `UploadHandle` on the front-end has been removed in favour of a simpler recursive function call

Things still left to do:

- Don't fire an alert for test failures due to YouTube 503s (by looking at the `UploadStatus`)
- Cloudwatch alarm for failed step function runs
- Move the code called by each lambda into the lambda itself (since we don't need to share it any more)
- Remove the old Kinesis stream etc from the cloud-formation
- Properly poll for progress from the transcoder (when avoiding YouTube)
- Add the asset to the atom in DEV (maybe)

### DIFF
--- a/app/util/StepFunctions.scala
+++ b/app/util/StepFunctions.scala
@@ -1,0 +1,120 @@
+package util
+
+import java.time.Instant
+
+import com.amazonaws.services.stepfunctions.model._
+import com.fasterxml.jackson.core.JsonParseException
+import com.gu.media.upload.model._
+import play.api.libs.json.{JsResultException, JsSuccess, Json}
+
+import scala.collection.JavaConverters._
+
+class StepFunctions(awsConfig: AWSConfig) {
+  def getById(id: String): Option[Upload] = {
+    val arn = s"${awsConfig.pipelineArn.replace(":stateMachine:", ":execution:")}:$id"
+
+    try {
+      val request = new DescribeExecutionRequest().withExecutionArn(arn)
+      val result = awsConfig.stepFunctionsClient.describeExecution(request)
+      val json = Json.parse(result.getInput)
+
+      Some(json.as[Upload])
+    } catch {
+      case _: ExecutionDoesNotExistException =>
+        None
+    }
+  }
+
+  def getStatus(atomId: String): Iterable[UploadStatus] = {
+    val runningJobs = getExecutions(atomId, ExecutionStatus.RUNNING)
+    val failedJobs = getExecutions(atomId, ExecutionStatus.FAILED).filter(lessThan10MinutesOld)
+
+    val running = runningJobs.map(getRunningStatus)
+    val failed = failedJobs.map(getFailedStatus)
+
+    running ++ failed
+  }
+
+  def start(upload: Upload): Unit = {
+    val stepFunctionsRequest = new StartExecutionRequest()
+      .withName(upload.id)
+      .withStateMachineArn(awsConfig.pipelineArn)
+      .withInput(Json.stringify(Json.toJson(upload)))
+
+    awsConfig.stepFunctionsClient.startExecution(stepFunctionsRequest)
+  }
+
+  private def getExecutions(atomId: String, filter: ExecutionStatus): Iterable[ExecutionListItem] = {
+    val request = new ListExecutionsRequest()
+      .withStateMachineArn(awsConfig.pipelineArn)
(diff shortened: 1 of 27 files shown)


app/util/StepFunctions.scala
```diff
@@ -0,0 +1,120 @@
...
+    val events = getEvents(execution)
+
+    val cause = events.find(_.getType == "ExecutionFailed") match {
+      case Some(event) =>
+        val cause = event.getExecutionFailedEventDetails.getCause
+
+        try {
+          (Json.parse(cause) \ "errorMessage").as[String]
+        } catch {
+          case _: JsonParseException | _: JsResultException =>
```
akash1810: is this because AWS give dodgy responses?

common/src/main/scala/com/gu/media/upload/model/UploadStatus.scala
```diff
@@ -0,0 +1,14 @@
+package com.gu.media.upload.model
+
+import org.cvogt.play.json.Jsonx
+import play.api.libs.json.Format
+
+case class UploadStatus(id: String, status: String, current: Int, total: Int, failed: Boolean)
+
+object UploadStatus {
+  def indeterminate(id: String, status: String): UploadStatus = {
+    UploadStatus(id, status, current = -1, total = -1, failed = false)
```
akash1810: why `-1`?

public/video-ui/src/actions/UploadActions/s3Upload.js
```diff
@@ -34,38 +33,30 @@ function uploadError(error) {
   };
 }
 
-export function startUpload(id, file, completeFn, selfHost, useStepFunctions) {
+export function startUpload(id, file, selfHost, completeFn) {
```
akash1810: should this return a Promise rather than using the callback pattern?

public/video-ui/src/components/VideoUpload/VideoTrail.js
```diff
@@ -189,18 +196,24 @@ export default class VideoTrail extends React.Component {
...
+      upload => upload.id !== this.props.s3Upload.id
+    );
 
     uploads.forEach(upload => {
-      const total = upload.parts[upload.parts.length - 1].end;
-      const progress =
-        upload.progress.uploadedToS3 + upload.progress.uploadedToYouTube;
-
-      blocks.push(this.renderRemoteUpload(upload.id, total, progress));
+      const total = upload.total == -1 ? undefined : upload.total;
```
akash1810: `===` for stronger js comparison

public/video-ui/src/services/UploadsApi.js
```diff
@@ -31,10 +26,27 @@ class UploadFunctions {
...
+      const uploadPartRecursive = parts => {
+        if (parts.length === 0) {
+          resolve(true);
+        } else {
+          const part = parts[0];
+
+          this.uploadPart(upload, part, file, progressFn)
+            .then(s3Request => {
+              return s3Request.promise().then(() => {
+                uploadPartRecursive(parts.slice(1));
```
akash1810: minor: make this `resolve(uploadPartRecursive(parts.slice(1))` to be more explicit in the `Promise` you're returning

app/util/StepFunctions.scala
mbarton: Not all failures will have a nice error message (lambdas timing out for example).

common/src/main/scala/com/gu/media/upload/model/UploadStatus.scala
mbarton: Good point. I'll make these optional

public/video-ui/src/services/UploadsApi.js
mbarton: Not sure what you mean? I think I can only resolve the promise at the end of the chain

PR #467, opened 2017-05-31
add scribe

Reettaphant:
- add a scribe editor for adding links and lists to video atom description and trail text
- when updating youtube description, strip links and html tags off the description. If a published video page exists, add a link to the bottom of the description pointing to it

@akash1810 @jennysivapalan @mbarton 
out

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -147,7 +147,7 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
     }
   }
 
-  private def parseDescriptionForYoutube(description: String): String = {
+  private def removeHtmlTagsForYouTube(description: String): String = {
       val html = Jsoup.parse(description)
       html.select("a").remove()
       html.text()
@@ -174,7 +174,7 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
   private def updateYoutubeMetadata(previewAtom: MediaAtom, asset: Asset) = {
 
     val description = previewAtom.description.map(description => {
-      parseDescriptionForYoutube(description) + getComposerLinkText(previewAtom.id)
+      removeHtmlTagsForYouTube(description) + getComposerLinkText(previewAtom.id)
     })
 
     val metadata = YouTubeMetadataUpdate(
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -155,19 +155,19 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
 
   private def getComposerLinkText(atomId: String): String = {
 
-    try {
-      val usages: JsValue = (capi.capiQuery("/atom/media/" + atomId + "/usage", true) \ "response" \ "results").get
-      val usagesList = usages.as[List[String]]
 
-      val composerPage = usagesList.find(usage => ".*/video/.*".r.pattern.matcher(usage).matches())
+    val usages: JsValue = (capi.capiQuery("/atom/media/" + atomId + "/usage", true) \ "response" \ "results").get
+    val usagesList = usages.as[List[String]]
 
-      composerPage match {
-        case Some(page) => "\n View the video at https://www.theguardian.com/" + page
-        case None => ""
-      }
-    }
-    catch {
-      case _ => ""
+    val composerPage = usagesList.find(usage => {
+      val query = capi.capiQuery(usage, true)
+      val contentType = (capi.capiQuery(usage, true) \ "response" \ "content" \ "type").get.as[String]
+      contentType == "video"
+    })
+
+    composerPage match {
+      case Some(page) => "\n View the video at https://www.theguardian.com/" + page
+      case None => ""
     }
   }
 
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -4,6 +4,11 @@ import java.net.URL
 import java.time.Instant
 import java.util.Date
 
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.jsoup.nodes._
+import play.api.libs.json.{JsValue, JsLookupResult}
+
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.gu.atom.play.AtomAPIActions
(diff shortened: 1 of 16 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,40 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
     }
   }
 
+  private def parseDescriptionForYoutube(description: String): String = {
+      val html = Jsoup.parse(description)
```
mbarton: Just out of interest, what happens if we send HTML in a description to YouTube? I'm wondering if they clean it or if we just end up with garbage

app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,40 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
...
+
+      val composerPage = usagesList.find(usage => ".*/video/.*".r.pattern.matcher(usage).matches())
+
+      composerPage match {
+        case Some(page) => "\n View the video at https://www.theguardian.com/" + page
+        case None => ""
+      }
+    }
+    catch {
+      case _ => ""
```
mbarton: When would we hit this case?

public/video-ui/src/services/VideosApi.js
```diff
@@ -121,12 +121,15 @@ export default {
 
     function updateArticleField(stage, data, composerUrl, pageId) {
       if (data.value || data.belongsTo === 'settings') {
+        const value = data.isHtml
+          ? data.value.split('"').join('\\"')
```
mbarton: Does this work if we put quotes in the text?

app/model/commands/PublishAtomCommand.scala
Reettaphant: We just end up with an error from the youtube api so we can't even update it.

app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,40 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
...
+      html.text()
+  }
+
+  private def getComposerLinkText(atomId: String): String = {
+
+    try {
+      val usages: JsValue = (capi.capiQuery("/atom/media/" + atomId + "/usage", true) \ "response" \ "results").get
+      val usagesList = usages.as[List[String]]
+
+      val composerPage = usagesList.find(usage => ".*/video/.*".r.pattern.matcher(usage).matches())
```
Reettaphant: Yes, I will do that.

public/video-ui/src/services/VideosApi.js
Reettaphant: Yes, although mentioning this made me realise that having double quotes in any text fields (e.g. title) would break syncing back to composer so I'm fixing this to do it for all the fields wit text input.

public/video-ui/src/components/FormFields/ScribeEditor.js
```diff
@@ -0,0 +1,283 @@
...
+
+  componentDidMount() {
+    this.scribe = new Scribe(this.refs.editor);
+
+    this.configureScribe();
+
+    this.scribe.on('content-changed', this.onContentChange);
+    this.refs.editor.innerHTML = this.props.value;
+  }
+  //this.props.copiedValue
```
akash1810: is the linter happy with this?

public/video-ui/src/services/VideosApi.js
```diff
@@ -121,12 +121,15 @@ export default {
 
     function updateArticleField(stage, data, composerUrl, pageId) {
       if (data.value || data.belongsTo === 'settings') {
+        const value = data.isFreeText
+          ? data.value.split('"').join('\\"')
```
akash1810: this looks like a fancy `.replace('"', '\"')`?

app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,41 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
     }
   }
 
+  private def parseDescriptionForYoutube(description: String): String = {
+      val html = Jsoup.parse(description)
+      html.select("a").remove()
```
Reettaphant: All the html tags are stripped in the line below (`html.text()`) so all the list tags will be stripped away. This line removes all links with their content so in the example you shared, all the contents on the list would be removed. If the list contained text instead of a link, the text inside would remain but the list tags would be removed. I guess it would be simpler to always remove all lists, not sure if there is a case when lists are used without links.

public/video-ui/src/services/VideosApi.js
Reettaphant: Replace only replaces one character so this seemed like the simples way of doing it.

akash1810: `.replace(/"/g, '\\"')` should work, but its a minor point, so feel free to ignore 😄

app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,41 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
     }
   }
 
+  private def parseDescriptionForYoutube(description: String): String = {
```
jennysivapalan: Maybe the method could show more intent so, `removeHtmlTagsForYouTubeDescription`. That might help us in the future to remember why we do this?

app/model/commands/PublishAtomCommand.scala
```diff
@@ -141,11 +147,41 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
...
+      composerPage match {
+        case Some(page) => "\n View the video at https://www.theguardian.com/" + page
+        case None => ""
+      }
+    }
+  }
+
   private def updateYoutubeMetadata(previewAtom: MediaAtom, asset: Asset) = {
+
+    val description = previewAtom.description.map(description => {
```
Reettaphant: As we discussed, this becomes difficult because getting the youtube description requires querying capi so I'm just going to leave this here.

package.json
```diff
@@ -74,6 +74,11 @@
     "redux": "^3.6.0",
     "redux-thunk": "^2.1.0",
     "reqwest": "^2.0.5",
+    "scribe": "guardian/scribe#fcf4e16",
```
Reettaphant: Because scribe hasn't been released for a while so newer versions are not available on npm. I could look into releasing it later today.

PR #478, opened 2017-06-06
Add YouTube asset immediately after upload

mbarton:
This fixes the problem where the step functions would upload a video to YouTube and then fail before adding the asset to the atom. That's confusing because the video is on YouTube but not visible in the tool.

I've split the old add asset lambda into two: one for YouTube and one for self-hosted. The YouTube one runs immediately after the upload has completed. The upload is then hidden from the UI since the user can already see the asset.

The step functions pipeline now looks like this:

c08dbcee-50bb-4375-b805-b6323559fe0b

Check out the sick loop-the-loop on the self-hosted route 🏄 

I've also added code to fail the upload if the video is a duplicate.

NB: the self-hosted lambda is still a dummy to be completed in subsequent PRs


mbarton:
Closing for now. This needs re-working to avoid adding the asset to the atom until YouTube has processed or rejected it

PR #480, opened 2017-06-06
live capi endpoint should use live capi

akash1810:
FYI it looks like this endpoint isn't being used, but intellij could be lying to me...


LATaylor-guardian:
👍

PR #481, opened 2017-06-08
Add capi byline tags

Reettaphant:
- Adds a byline picker and syncs bylines with composer
- The byline picker is specific to bylines at the moment but might be adapted to picking tags that don't allow text input if this doesn't make it too complicated
- All input fields are made bigger in this pr so that the tag picker fields won't look out of place (they need to be bigger to display tags)
- depends on: https://github.com/guardian/atom-maker/pull/28
screen shot 2017-06-08 at 12 50 08


public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
```diff
@@ -0,0 +1,301 @@
+import React from 'react';
+import ContentApi from '../../services/capi';
+
+export default class FormFieldBylinePicker extends React.Component {
+  state = {
+    bylineTags: null,
```
mbarton: Could just be an empty array?

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
```diff
@@ -0,0 +1,301 @@
...
+
+  onUpdate = newValue => {
+    this.setState({
+      tagValue: newValue
+    });
+    this.props.onUpdateField(this.tagsToString(newValue));
+  };
+
+  updateInput = e => {
+    if (this.state.lastAction === 'SPACE') {
```
mbarton: Rather than tracking keypresses, is it possible to write a pure function that goes from the input value to the state required?

Reettaphant: We need a way of knowing if the user is deleting or adding content, I don't think there is a way of doing that without tracking the key presses.

public/video-ui/src/util/tagParsers.js
```diff
@@ -0,0 +1,31 @@
...
+      if (element.match('^profile/')) {
+        return ContentApi.getLivePage(element).then(capiResponse => {
+          const tag = capiResponse.response.tag;
+          return {
+            id: tag.id,
+            webTitle: tag.webTitle
+          };
+        });
+      } else {
+        return new Promise(resolve => resolve(element));
```
mbarton: Nitpick: can also be `Promise.resolve(element)`

public/video-ui/src/util/removeStringTagDuplicates.js
```diff
@@ -0,0 +1,37 @@
+export default function removeStringTagDuplicates(tag, tagValue) {
+  // Remove string input that appears in the last words of the tag input that is being added
```
mbarton: Might be useful to provide a few example inputs and outputs in the comment?

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
```diff
@@ -0,0 +1,241 @@
...
+      ContentApi.getBylineTags(searchText)
+        .then(capiResponse => {
+          const bylines = capiResponse.response.results.map(result => {
+            const tags = { id: result.id, webTitle: result.webTitle };
+            return tags;
+          });
+          this.setState({
+            bylineTags: bylines
+          });
+          return;
```
mbarton: Is this required?

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
```diff
@@ -0,0 +1,241 @@
...
+
+      this.setState({
+        inputString: lastInput,
+        lastAction: 'OTHER'
+      });
+
+      const newValue = this.state.tagValue.slice(
+        0,
+        this.state.tagValue.length - 1
+      );
```
jennysivapalan: Can the parameters for `slice` be on one line? Might be more readable?

public/video-ui/src/util/removeStringTagDuplicates.js
jennysivapalan: I think we should add a test!

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
Reettaphant: This is the pre-commit hook reorganising code. If it is difficult to read we should probably change how it works.

public/video-ui/src/util/removeStringTagDuplicates.js
Reettaphant: I've added a few tests to this now.

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
mbarton: I'd be up for removing the line length rule, it often does this to me

Reettaphant:
@jennysivapalan @mbarton I've added tests for removing duplicate input and split some of the functionality in the byline picker to utilities functions. We can't get rid of listening to key presses because its the only way we can know when the user is deleting input and when they are adding input. The build should be passing once https://github.com/guardian/content-atom/pull/91 and https://github.com/guardian/atom-maker/pull/28 have been merged and new versions released.

public/video-ui/src/test/stringTagDuplicates.spec.js
```diff
@@ -0,0 +1,56 @@
+import removeStringTagDuplicates from '../util/removeStringTagDuplicates';
+
+const simpleTag = { id: 'tag', webTitle: 'tag' };
+
+describe('Removing string tag duplicates', () => {
+  it('does not trim when imput is empty', () => {
```
akash1810: typo

public/video-ui/src/util/removeStringTagDuplicates.js
```diff
@@ -0,0 +1,40 @@
...
+  const valuesLength = addedValues.length;
+  let numberOfMatches = 0;
+
+  const firstTagMatch = tagWords.findIndex(tagWord => {
+    return addedValues[valuesLength - 1] === tagWord;
+  });
+
+  if (firstTagMatch !== -1) {
+    numberOfMatches = 1;
+    for (var i = firstTagMatch + 1; i < tagWords.length; i++) {
```
akash1810: could we use a `.forEach` here?

public/video-ui/styles/components/_forms.scss
```diff
@@ -219,6 +247,91 @@
   }
 }
 
+.form__field__tags {
```
akash1810: can be merged with the rules below

public/video-ui/styles/components/_forms.scss
```diff
@@ -219,6 +247,91 @@
...
+    border: 1px solid $color300Grey;
+    border-radius: 2px;
+    margin-right: 2px;
+    font-weight: bold;
+    cursor: pointer;
+    flex-shrink: 0;
+
+
+    &:after {
+      content: '×';
```
akash1810: minor: why not use the icons?

public/video-ui/src/components/FormFields/FormFieldBylinePicker.js
```diff
@@ -0,0 +1,261 @@
+import React from 'react';
+import ContentApi from '../../services/capi';
+import { tagsFromStringList, tagsToStringList } from '../../util/tagParsers';
+import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
+
+export default class FormFieldBylinePicker extends React.Component {
```
Reettaphant: I was going to change the whole name in the pr that makes this component a more generic tag picker so maybe it is enough to just do it then?

public/video-ui/src/util/removeStringTagDuplicates.js
Reettaphant: I want to break out of the loop immediately when a match is not found (e.g. if someone has typed in 'Mary Smith' but picks a tag with web title 'Mary Jones', we don't want to trim any words from the input string). I guess it could be done with a foreach as well, but I think this is simpler.

public/video-ui/styles/components/_forms.scss
Reettaphant: This is how atom workshop does multiselect elements, I think it is good to have consistent look on the pickers if possible.

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -68,6 +69,16 @@ class VideoData extends React.Component {
             >
               <ScribeEditorField />
             </ManagedField>
+
+            <ManagedField
+              fieldLocation="byline"
+              name="Byline Tags"
```
akash1810: could we make this `Byline` to be consistent with Composer?

public/video-ui/styles/components/_forms.scss
```diff
@@ -219,6 +247,88 @@
...
 }
 
+.form__field__tags {
+  border: 1px solid $color400Grey;
+  display: block;
+  border-top: 1px solid $color400Grey;
+  width: 100%;
+  padding: 5px 5px;
+  cursor: pointer;
+  border-top: 12px solid $color650Grey;
```
akash1810: we've 2 `border-top` rules now for the same class, can we delete one?

PR #484, opened 2017-06-13
Show details of YouTube upload failure

mbarton:
Adds a description of why the upload failed and removes the Activate link on the failed asset.

2951270c-16bd-4b9d-8a24-d3822a43d6dc

Also fixes a race condition introduced by #478 by filtering uploads client side (based on a new `assetAdded` field) rather than server side.


jennysivapalan:
Perhaps our failure messages could me more verbose, I'm not sure what `conversion` means, we could give a bit more detail that the user can forward on to the team or in the `duplicate` case perhaps state `This video upload failed as it is on YT already`?

mbarton:
@jennysivapalan I've added human readable strings for the most common errors

PR #488, opened 2017-06-14
Remove unused lambda parameters from step functions

mbarton:
Further to #487, removing the references to the old upload table from the step functions lambda.

I've also made the `AlertWebhook` parameter non-optional since it was breaking RiffRaff deploys. It's fine to set it to the blank string in DEV and CODE since it's not used anyway (due to conditional creation in the template).

### DIFF
--- a/uploader/src/main/resources/cfn-template.yaml
+++ b/uploader/src/main/resources/cfn-template.yaml
@@ -26,9 +26,6 @@ Parameters:
   UploadBucketName:
     Description: "The S3 bucket where videos are uploaded to"
     Type: "String"
-  UploadTrackingTable:
-    Description: "The Dynamo table to store progress"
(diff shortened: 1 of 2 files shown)


uploader/src/main/resources/cfn-template.yaml
```diff
@@ -39,7 +36,7 @@ Parameters:
   AlertWebhook:
     Description: "Where CloudWatch alerts are sent"
     Type: "String"
-    Default: ""
+    # Optional in DEV/CODE (just fill in with a dummy string)
```
jennysivapalan: I'd remove the comment and add `dummy-string` as the parameter for DEV and CODE.

mbarton: 👍

PR #492, opened 2017-06-19
Stop slack alerts for YouTube backend errors

mbarton:
YouTube can sometimes throw a wobbly and return 5XX errors from their API. This has made our integration tests fail sometimes and is annoying to have to investigate.

We're not going to add some kind of retry to the main app for this since we've never seen it happen to an actual user - only to automated tests. Even if it did happen to a real request, the user can press the Publish button again manually.

This PR translates the back-end errors into 503 responses and then suppresses slack notifications. I've also stopped the tests retrying 'publish' on failure, since this was pushing us over our API usage limit for no benefit.

### DIFF
--- a/app/model/commands/CommandException.scala
+++ b/app/model/commands/CommandException.scala
@@ -1,6 +1,7 @@
 package model.commands
 
-import play.api.mvc.{Results, Result}
+import com.google.api.client.googleapis.json.{GoogleJsonError, GoogleJsonResponseException}
+import play.api.mvc.{Result, Results}
 
 case class CommandException(msg: String, responseCode: Int) extends RuntimeException(msg)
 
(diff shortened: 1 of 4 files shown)


mbarton:
Not sure what you mean? The writing to disk (`NO_ALERTS` file) is just so the `int-test-runner` script can suppress the slack alerts

Reettaphant:
Not having these as slack notifications would be great. However, do we know that these youtube errors cannot hide any other problems in mam? Do we need to know if they suddenly happening much more frequently than they used to?

mbarton:
As discussed IRL, I agree it's not great to add code to identify and suppress errors since it's impossible to know if they will suppress other things in the future. I've done my best to avoid this but that's no guarantee.

I'm not so worried about frequency. If it starts happening a lot then YouTube is either down (which is obvious to the whole internet) or we have made a change that is sending them crap - which should result in 4XX errors in theory anyway.

PR #493, opened 2017-06-20
Youtube fields

Reettaphant:
- Move pluto id, youtube channel and youtube category to the video upload page 
- If moving the category over to this page does not fit users' workflow, we can change it back
- Don't allow uploading to youtube if channel or category id is missing
- youtube channels and pluto ids cannot be changed once they have been added


public/video-ui/src/components/VideoYoutube/VideoYoutube.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import { ManagedForm, ManagedField } from '../ManagedForm';
+import SelectBox from '../FormFields/SelectBox';
+
+export default class YoutubeMetaData extends React.Component {
```
mbarton: I think `YouTubeMetaData` is a clearer name than `VideoYoutube`, could we rename the file?

public/video-ui/src/components/VideoUpload/VideoUpload.js
```diff
@@ -98,13 +125,22 @@ class VideoUpload extends React.Component {
     }
   }
 
+  renderDataMissingMessage() {
+    if (this.state.file && this.videoDataMissing()) {
+      const dataInstructions =
+        'You have to add a channel and a category before you upload a video. ';
+      return <div className="error"> {dataInstructions} </div>;
```
akash1810: minor: don't really need the `const` here

PR #494, opened 2017-06-20
Add direct upload assets in DEV

mbarton:
Moving to step functions for the upload pipeline broke adding the resulting video as an asset to the atom. This PR brings it back, meaning DEV now operates exactly the same as CODE and PROD. This will make testing the forthcoming self-hosted asset work much easier.

`AddAssetToAtom` now gets the Thrift atom directly from the database, adds the asset, saves (also publishing to CAPI) and records what it has done in the audit trail.

A brief summary of changes:

- Moved `AuditDataStore` to `common`
- Added asset ID to `UploadStatus` for filtering out running jobs on the client side
    - This fixes a class of race conditions where the upload can disappear from the UI
- Pipeline Cloud Formation changes to give permission to access Dynamo and Kinesis
- Added `logback.xml` to the uploader
    - It decided logging the bytes of a video being uploaded at DEBUG level was a good idea!


app/model/commands/Command.scala
```diff
@@ -1,9 +1,17 @@
...
 
 trait Command extends UnpackedDataStores {
   type T
 
   def process(): T
+
+  def getUsername(user: PandaUser): String = {
+    user.email match {
+      case "" => user.firstName
+      case _ => user.email
```
jennysivapalan: When would email be `""`?

mbarton: It was there previously, I think to cover automated users @akash1810?

mbarton:
Yeah I should've been clearer about that. The audit code moved to common so it can be used by the uploader lambdas when they update the asset. The atom diffing code remained since it was only used in the app

app/model/commands/Command.scala
akash1810: HMAC authed users have a `""` email.

I did [try](https://github.com/guardian/panda-hmac/pull/2) to fix this a while ago but failed and this was a quick win.

PR #495, opened 2017-06-21
Add bash shebang

mbarton:
#492 caused the following error when run on TeamCity:

```
./integration-tests/int-test-runner.sh: 5: ./integration-tests/int-test-runner.sh: [[: not found
```

I believe this will be fixed by forcing the use of `bash`


shtukas:
:+1:

PR #500, opened 2017-06-23
Remove upload button permission check

7091lapS:
This makes direct upload accessible to anyone who uses the tool


app/controllers/AtomController.scala
```diff
@@ -35,12 +35,11 @@ trait AtomController extends Controller with UnpackedDataStores {
 
   import authActions.APIAuthAction
 
-  object CanUploadAsset extends ActionBuilder[UploadUserRequest] {
+  object ActionWithPermission extends ActionBuilder[UploadUserRequest] {
```
jennysivapalan: Could be called `HasPermission`?

jennysivapalan: Or `HasUploadPermssion` ?

mbarton: Yeah we struggled with the naming too. It won't block anyone based on a particular permission but will pass all the permissions to the next block so that can make a decision.

We also tried pulling out the code to be a simple function rather than an `ActionBuilder` but then we would have had to refactor the parsing helper method to take `Future[Result]` as well. Messy stuff.

Janus calls them `APIAuthAction`, perhaps `APIPermissionAction` is reasonable to follow the same convention?

app/controllers/UploadController.scala
```diff
@@ -34,9 +33,11 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
     Ok(Json.toJson(running))
   }
 
-  def create = CanUploadAsset { implicit raw =>
+  def create = ActionWithPermission { implicit raw =>
     parse(raw) { req: CreateRequest =>
-      if(canPerformUpload(raw.permissions, req.selfHost)) {
+      if(req.selfHost && !raw.permissions.addSelfHostedAsset) {
+        Unauthorized(s"User ${raw.user.email} is not authorised with permissions to upload asset")
```
akash1810: could the message be "is not authorised to self host" to be more explicit?

app/controllers/AtomController.scala
mbarton: We've settled on `LookupPermission`

PR #502, opened 2017-06-26
Support self-hosted assets in the UI

mbarton:
You've seen how to [add them](https://github.com/guardian/media-atom-maker/pull/497), you've seen how to [activate them](https://github.com/guardian/media-atom-maker/pull/501), now do all that **and more** direct from the media-atom-maker UI!

7b0ea22f-3030-495e-93ab-46beaed59aca

48906daf-06b2-4a9a-9da0-8f88741e28dc

Previously the UI would stitch together three pieces of info to form the video assets pages:

- Assets added to the atom
- Step function pipeline progress
- YouTube processing status

I found adding the additional work for self-hosted assets pushed the code over the edge of complexity. To fix this, I've introduced a new `ClientAsset` class that does the stitching server-side. The client is then free to simply render the asset or processing progress it is given.

See [ClientAssetTest](https://github.com/guardian/media-atom-maker/compare/mbarton/self-hosted-asset-ui-3?expand=1#diff-a8da3e10fead9a40fb874d1c574cc913) for all the combinations.

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -132,10 +132,13 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
 
   private def getRunning(job: ExecutionListItem): Option[ClientAsset] = {
     val events = stepFunctions.getEventsInReverseOrder(job)
-    val upload = stepFunctions.getUpload(events)
-    val error = stepFunctions.getError(events)
 
-    upload.map(ClientAsset(_, error))
+    val upload = stepFunctions.getTaskEntered(events)
+    val error = stepFunctions.getExecutionFailed(events)
+
+    upload.map { case(state, upload) =>
+      ClientAsset(state, upload, error)
+    }
   }
 }
 
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -36,7 +36,7 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
     val jobs = stepFunctions.getJobs(atomId)
     val running = jobs.flatMap(getRunning)
 
-    val assets = added ++ running
+    val assets = running ++ added
     Ok(Json.toJson(assets))
   }
 
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -2,13 +2,15 @@ package controllers
 
 import java.util.UUID
 
-import _root_.model.MediaAtom
+import _root_.model.{ClientAsset, MediaAtom}
 import _root_.model.commands.CommandExceptions._
+import com.amazonaws.services.stepfunctions.model.ExecutionListItem
 import com.gu.media.MediaAtomMakerPermissionsProvider
 import com.gu.media.logging.Logging
 import com.gu.media.model.{SelfHostedAsset, VideoSource}
 import com.gu.media.upload._
 import com.gu.media.upload.model._
+import com.gu.media.youtube.YouTubeVideos
 import com.gu.pandahmac.HMACAuthActions
 import com.gu.pandomainauth.model.User
 import controllers.UploadController.CreateRequest
@@ -18,7 +20,8 @@ import play.api.libs.json.{Format, Json}
 import util.{AWSConfig, CredentialsGenerator, StepFunctions}
 
 class UploadController(override val authActions: HMACAuthActions, awsConfig: AWSConfig, stepFunctions: StepFunctions,
-                       override val stores: DataStores, override val permissions: MediaAtomMakerPermissionsProvider)
+                       override val stores: DataStores, override val permissions: MediaAtomMakerPermissionsProvider,
+                       youTube: YouTubeVideos)
 
   extends AtomController with Logging with JsonRequestParsing with UnpackedDataStores {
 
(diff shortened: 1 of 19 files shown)


public/video-ui/src/components/VideoPreview/VideoPreview.js
```diff
@@ -1,33 +1,25 @@
...
-    }
-  };
-
   renderPreview() {
-    const activeAsset = this.getActiveAsset();
+    const { assets, activeVersion } = this.props.video;
+    const relevant = assets.filter(asset => asset.version === activeVersion);
 
-    if (!activeAsset) {
+    if (!assets || !activeVersion || relevant.length === 0) {
```
Reettaphant: If the assets do not exist (as is suggested here that they may not), will we get an exception from line 8?

public/video-ui/src/components/VideoPreview/VideoPreview.js
```diff
@@ -1,33 +1,25 @@
...
-      ) {
-        return this.props.video.assets[i];
-      }
-    }
-  };
-
   renderPreview() {
-    const activeAsset = this.getActiveAsset();
+    const { assets, activeVersion } = this.props.video;
+    const relevant = assets.filter(asset => asset.version === activeVersion);
```
Reettaphant: Can we have more than one active asset?

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -0,0 +1,90 @@
...
+      content: <YouTubeEmbed id={asset.id} />,
+      activate: active ? false : () => selectAsset(Number(id))
+    };
+  } else {
+    return {
+      title: `Version ${id}`,
+      active,
+      content: <VideoEmbed sources={asset.sources} />,
+      activate: active ? false : () => selectAsset(Number(id))
+    };
```
Reettaphant: Would this part be clearer if we had an assetProps class?

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -0,0 +1,90 @@
...
+        ? UPLOAD_FAILED_MSG
+        : <ProgressBar {...processing} />
+    };
+  } else if (asset.id) {
+    return {
+      title: `ID: ${asset.id}`,
+      active,
+      href: `https://www.youtube.com/watch?v=${asset.id}`,
+      content: <YouTubeEmbed id={asset.id} />,
+      activate: active ? false : () => selectAsset(Number(id))
```
Reettaphant: Are we ever going to call activate on asset that is active? Is the active check here needed?

Reettaphant:
The button for self hosted uploads is huge! Could it be made the same size as the button above?

public/video-ui/src/components/VideoPreview/VideoPreview.js
mbarton: We can only have one active version. That version might have multiple entries in the asset list - one for each transcode

mbarton:
@Reettaphant agreed. I've turned the button into a checkbox

app/controllers/UploadController.scala
```diff
@@ -27,10 +30,14 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
   def list(atomId: String) = APIAuthAction {
-    // Anyone can see the running uploads.
-    // Only users with permission can create/complete/delete them.
-    val running = stepFunctions.getStatus(atomId)
-    Ok(Json.toJson(running))
+    val atom = MediaAtom.fromThrift(getPreviewAtom(atomId))
+    val added = ClientAsset.byVersion(atom.assets, youTube.getProcessingStatus)
+
+    val jobs = stepFunctions.getJobs(atomId)
+    val running = jobs.flatMap(getRunning)
```
akash1810: minor: `runningJobs` would be more explicit

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -0,0 +1,90 @@
...
+const UPLOAD_FAILED_MSG = (
+  <p>
+    <strong>Upload Failed</strong><br />
+    You may retry your upload.<br />
+    This message will disappear after 10 minutes.
+  </p>
+);
+
+function ProgressBar({ current, total }) {
+  return total !== undefined && current !== undefined
```
akash1810: minor: could just be `total && current`

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -0,0 +1,90 @@
...
+
+export function Asset({ content, title, href, active, activateFn }) {
+  const button = activateFn
+    ? <button className="button__secondary button__active" onClick={activateFn}>
+        Activate
+      </button>
+    : false;
+
+  const link = href
+    ? <a href={href}><Icon icon="open_in_new" className="icon__assets" /></a>
```
akash1810: can we add the `target` attribute to open in a new tab

public/video-ui/src/components/utils/VideoEmbed.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export function VideoEmbed({ sources }) {
+  return (
+    <video className="video-player" controls>
```
akash1810: can we set the `preload` attribute to `metadata` to avoid downloading the video on page load

public/video-ui/src/components/utils/VideoEmbed.js
```diff
@@ -0,0 +1,13 @@
...
+      {sources.map(source => {
+        return (
+          <source key={source.src} src={source.src} type={source.mimeType} />
```
akash1810: If there is only one source, Safari will barf as it expects a single source to be set as the `src` attribute on the `video`.

We don't target Safari, so happy to keep as is. Maybe add a comment?

public/video-ui/src/components/VideoUpload/VideoAsset.js
mbarton: Unfortunately not! If `current` is `0`, I want it to render an empty bar

mbarton:
@akash1810 I'll do a follow up PR for your review feedback

public/video-ui/src/components/utils/VideoEmbed.js
mbarton: Wtf Safari. Is there a way I can support it transparently, by putting `src` on the `video` tag or something?

akash1810: We can set the `src` of the `video` tag. I believe this wins over the `source` set, so would have to be either or rather than setting both.

Something like:

```js
if (sources.length === 1) {
  <video src={source.src}>
} else {
  <video>
    {sources.map(source => <source src={source.src}>}
  </video>
}
```

should do the trick.

PR #505, opened 2017-06-27
Single template for atom defaultHtml

mbarton:
I wanted to remove the fixed `width` and `height` on the YouTube `iframe` in `defaultHtml`. This is so we can use the `defaultHtml` to display a preview in Workflow and Composer.

I then noticed we had duplicate templates for each embed, so I brought them together into a single one.

I also made the `defaultHtml` for an asset with no video be the poster image, which is what frontend was already doing.

### DIFF
--- a/app/views/MediaAtom/defaultEmbed.scala.html
+++ b/app/views/MediaAtom/defaultEmbed.scala.html
@@ -1,8 +1,7 @@
 @import com.gu.media.model.{VideoAsset, YouTubeAsset, SelfHostedAsset}
-
 @(asset: Option[VideoAsset], posterUrl: Option[String])
 
-@(asset, posterUrl) match {
+@defining((asset, posterUrl)) {
   case (None, Some(poster)) => {
     <img src="@poster"/>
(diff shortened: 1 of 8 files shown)


app/views/MediaAtom/defaultEmbed.scala.html
```diff
@@ -0,0 +1,24 @@
...
+  case (None, Some(poster)) => {
+    <img src="@poster"/>
+  }
+
+  case (Some(YouTubeAsset(id)), _) => {
+    <iframe frameborder="0" allowfullscreen="true" src="https://www.youtube.com/embed/@id"></iframe>
+  }
+
+  case (Some(SelfHostedAsset(sources)), Some(poster)) => {
+    <video controls="controls" poster="@poster">
```
akash1810: can we add `preload="metadata"` so the browser won't fetch the video on page load

mbarton:
@akash1810 wouldn't older atoms have a video so the impact of reindexing would be limited?

PR #507, opened 2017-06-28
Remove create form 

Reettaphant:
- Remove the create form. When create button is clicked, the atom edit page is displayed.
- Media atom can be created by closing the atom data form once the title has been entered. `News` is added to the atom as the default category. These are the only required fields for atoms. 
- Since a lot of fields can now be empty, I also took the opportunity to fix the display of placeholders when field value does not exist. Some field components did not display placeholders, others didn't display them in italics

screen shot 2017-06-28 at 09 52 59


public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -99,21 +99,30 @@ function Editor({ date, onChange }) {
...
+function Display({ date, placeholder }) {
+  const displayString = date ? date.format(DATETIME_FORMAT) : placeholder;
 
   return (
     <div>
       <p className="details-list__title">Expiry Date</p>
-      <p className="details-list__field">{displayString}</p>
+      <p
+        className={
+          'details-list__field' + (!date ? ' details-list__empty' : '')
```
mbarton: Maybe pull this out into a variable? it's a bit busy with all the newlines here

public/video-ui/src/components/FormFields/SelectBox.js
```diff
@@ -31,7 +31,19 @@ export default class SelectBox extends React.Component {
       return (
         <div>
           <p className="details-list__title">{this.props.fieldName}</p>
-          <p className="details-list__field">{displayValue}</p>
+          <p
+            className={
+              'details-list__field ' +
```
mbarton: as above, probably wants to be a variable

public/video-ui/src/components/ReactApp.js
```diff
@@ -59,7 +59,12 @@ class ReactApp extends React.Component {
           currentPath={this.props.location.pathname}
           video={this.props.video || {}}
           publishedVideo={this.props.publishedVideo || {}}
-          showPublishedState={this.props.params.id ? true : false}
+          showPublishedState={
+            this.props.params.id ||
+              this.props.location.pathname === '/videos/create'
```
mbarton: Why did we need to make this change?

public/video-ui/src/services/pandaReqwest.js
```diff
@@ -42,6 +42,15 @@ export function pandaReqwest(reqwestBody, timeout = 0) {
 
     // prettier-ignore
     if (payload.contentType === 'application/json' && typeof payload.data === 'object') {
+
+      //Replace all empty strings with nulls so that dynamo can save all the data we
+      //send to the server
+      Object.keys(payload.data).forEach(key => {
```
mbarton: Could we do this at the component or placeholder level? Seems easy to break in the future if it's done here

Reettaphant: I wasn't sure about this one, wouldn't want to have to do it in multiple components either. Maybe it should be in the VideoApi service so we could make sure that videos with empty strings for properties never get saved?

public/video-ui/src/components/ReactApp.js
Reettaphant: Otherwise we'd get a `Create new video` button in the header when creating a video which doesn't make much sense. Since it is the same video data form that we are displaying, I think it makes sense to display the same header as when editing the atom.

public/video-ui/src/services/pandaReqwest.js
mbarton: Yeah if we have to do it lets limit it to just the services that require it

public/video-ui/src/components/FormFields/ScribeEditor.js
```diff
@@ -24,6 +24,10 @@ export default class ScribeEditorField extends React.Component {
   };
 
   getWords = text => {
+    if (!text) {
+      return 0;
```
akash1810: should this be an `int`? `getWords` suggests it should be a `string`

akash1810: `updateWordCount` also does `this.getWords(text).length` which won't work if `getWords` returns `0`

public/video-ui/src/util/replaceEmptyStrings.js
```diff
@@ -0,0 +1,9 @@
+export function replaceEmptyStrings(data) {
```
akash1810: minor: `nullifyEmptyStrings` may be more explicit as we always replace with `null`

public/video-ui/src/components/FormFields/ScribeEditor.js
Reettaphant: Yeah, this should be an empty array.

paperboyo:
Do all free text fields use Scribe? Asking, because it polices proper typographic entities like real apostrophes etc. And so that we won’t have to [fix things laterz](https://trello.com/c/IeZioLRs/117-convert-straight-quotes-to-curlies-automatically-in-headlines)...

Reettaphant:
@paperboyo The headline field does not use scribe, only the byline and the trail fields. It tries to replicate the functionality in composer. If the headline field in composer starts using scribe as suggested it should in the trello card you linked to, then we should switch over to using scribe as well.

paperboyo:
@Reettaphant When/if https://github.com/guardian/flexible-content/pull/2848 gets permanent, we will land in an infinite loop of ugly typographic features, though 😜.

Just joking, your call, most obviously!

PR #508, opened 2017-06-28
Third party YouTube video atoms

mbarton:
Allows the addition of any YouTube video to an atom via the "Add Asset to URL" text box. If the channel is not the same as that already attached to the atom, it will reject the add.

NB: this behaviour should be restricted to Hosted atoms. The UI changes for that will be in a follow up PR.

I've also made `RuntimeUploadMetadata` not optional and moved the channel into `YouTubeUploadMetadata` since it is not required for self-hosted assets.

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -99,9 +99,19 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
   }
 
   private def updateYouTube(previewAtom: MediaAtom, asset: Asset): Future[MediaAtom] = {
-    updateYoutubeMetadata(previewAtom, asset)
-    updateYoutubeThumbnail(previewAtom, asset)
-    createOrUpdateYoutubeClaim(previewAtom, asset)
+    previewAtom.channelId match {
+      case Some(channel) if youTube.allowedChannels.contains(channel) =>
+        updateYoutubeMetadata(previewAtom, asset)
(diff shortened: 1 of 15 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -99,9 +99,19 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youTu
   }
 
   private def updateYouTube(previewAtom: MediaAtom, asset: Asset): Future[MediaAtom] = {
-    updateYoutubeMetadata(previewAtom, asset)
-    updateYoutubeThumbnail(previewAtom, asset)
-    createOrUpdateYoutubeClaim(previewAtom, asset)
+    previewAtom.channelId match {
+      case Some(channel) if youTube.allowedChannels.contains(channel) =>
```
akash1810: As the main channel isn't in the allowed list, we're now able to add an asset with the YT url but will never update YT metadata.

We store a list of channels linked to the cms [in the cache](https://github.com/guardian/media-atom-maker/blob/master/app/controllers/Youtube.scala#L21). We could see if the video channel is in that list, and then treat it as a first party video? Might need some ux attention though. Lets ship this and see.

mbarton: Just gave it a go and I think it would be rather confusing because the initial headline and description are not imported when you manually add the asset.

I agree, ship this and see. We may need to do a bit more additional work since selecting a YouTube channel is mandatory right now, which blocks this. That won't be true once #507 is merged

PR #511, opened 2017-06-29
Fix video preview and review feedback

mbarton:
#502 had a bug where it would try to preview self-hosted assets using the YouTube play if the asset was just a single file. This PR fixes that and also addresses review feedback.

### DIFF
--- a/public/video-ui/src/components/VideoUpload/VideoAsset.js
+++ b/public/video-ui/src/components/VideoUpload/VideoAsset.js
@@ -43,13 +43,8 @@ export function Asset({ content, title, href, active, activateFn }) {
     : false;
 
   const link = href
-    ? <a href={href}>
-        <Icon
-          icon="open_in_new"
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -43,7 +43,14 @@ export function Asset({ content, title, href, active, activateFn }) {
     : false;
 
   const link = href
-    ? <a href={href}><Icon icon="open_in_new" className="icon__assets" /></a>
+    ? <a href={href}>
+        <Icon
+          icon="open_in_new"
+          className="icon__assets"
+          target="_blank"
```
akash1810: `target` and `rel` should be on the `a`

mbarton: oops

PR #512, opened 2017-06-29
Byline fixes

Reettaphant:
A few fixes to the byline picker

- Don't allow empty strings as input to bylines
- Add a scrollbar to tag list and improve its appearance
- Add instructions on how to add text input to byline. Initially I wanted to change this so that you would not have to press space to actually get text input added to the byline. I 
screen shot 2017-06-29 at 15 12 48
thought it is too easy to forget to do this. I then realised that this might become confusing with the other tag pickers that allow for entering text on the input field but this text does not get saved. So I've added a little reminder to press space to add text input instead.

### DIFF
--- a/public/video-ui/src/components/FormFields/ComposerTagPicker.js
+++ b/public/video-ui/src/components/FormFields/ComposerTagPicker.js
@@ -63,7 +63,8 @@ export default class ComposerTagPicker extends React.Component {
   };
 
   updateInput = e => {
-    if (this.state.lastAction === UserActions.space) {
+    const onlyWhitespace = !/\S/.test(this.state.inputString);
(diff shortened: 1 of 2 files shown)


public/video-ui/src/components/FormFields/ComposerTagPicker.js
```diff
@@ -269,6 +270,16 @@ export default class ComposerTagPicker extends React.Component {
     );
   }
 
+  renderBylineInstructions() {
+    if (this.props.tagType === 'contributor') {
+      return (
+        <span className="form__field__instructions">
+          Press space to add byline as text
```
akash1810: 👍

PR #514, opened 2017-06-30
Youtube picker

Reettaphant:
Remove youtube keyword picker and use the new tag picker for picking youtube keywords for consistency.
screen shot 2017-06-30 at 10 00 19


public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -93,29 +98,47 @@ export default class ComposerTagPicker extends React.Component {
...
 
   processTagInput = e => {
-    if (e.keyCode === keyCodes.space) {
-      this.setState({
-        lastAction: UserActions.space
-      });
+    if (e.keyCode === keyCodes.enter) {
+      const onlyWhitespace = !/\S/.test(this.state.inputString);
+      if (!onlyWhitespace) {
+        let newInput;
```
akash1810: can use a ternary here to avoid the `let`.

PR #520, opened 2017-07-03
Update package.scala

akash1810:
update content bundles for commercial


akash1810:
ideally this'll all be managed by tag manager and be self serve, but ⏲

PR #522, opened 2017-07-05
Hosted atom support

mbarton:
Updates to the UI to fully support hosted atoms.

When the user selects Hosted as the category, the UI drops out the fields that will no longer have any effect since we don't own the YouTube channel. The video is added to the atom using the existing "Add Asset by URL" box.

22a96bb3-60ee-4d8f-9ecd-39a6a27f53e9

The button to 'Create A Composer Page' is hidden if the atom is Hosted. Once an atom is Hosted and has assets, it cannot be switched back to a different category.

This PR also fixes support for managed videos (on Guardian channels that are not present in the drop-down). The process is the same but with a category that isn't Hosted. The UI drops out the fields and will not modify the video on publish.

I've moved around the components to make it clearer that YouTube information is only required when uploading to a managed channel:

28b8a544-23f2-475f-9c8c-2e2e4f8efcc5

5c477bb9-64ae-49c7-89f0-61bea6326a89

In either case, the user can always upload a self-hosted asset providing they have permission. Once an atom has a YouTube asset, it is locked to that YouTube channel forever.

🤖 *TECH NOTES* 🤖 

I've moved the cache of YouTube channels and categories to be used in all cases in the main Play app not just for the Youtube route controller. This means we can change the hacky `isMyVideo` method to just check the list of cached channels. The UI still only shows the `allowedChannels` but the filtering is done in the controller.

I've removed the filter so that Hosted atoms are shown on the front page.

Since #507 (removing the create form) has not yet been merged, I've had to add special code in `AddAssetCommand` that changes the YouTube channel to the right one if there are no current assets. Without this, you'd never be able to add any video that's not on one of the channels in the dropdown.

As before, videos on channels in the drop-down, deleting an atom on a channel in drop-down will make the YouTube assets private. The same will also happen if the atom expires. For atoms on other channels, the assets are unaffected.

Finally I refactored `VideoUpload` into separate components because it had got too big.


Reettaphant:
I've now merged getting rid of the create form so you are free to use the logic there.

app/controllers/Youtube.scala
```diff
@@ -1,25 +1,23 @@
...
-      youTube.categories}
-    Ok(Json.toJson(categories))
+    Ok(Json.toJson(youTube.categories))
   }
 
   def listChannels() = AuthAction {
-    val channels = cache.getOrElse[List[YouTubeChannel]]("channels", 1.hours) {
-      youTube.channels}
+    val channels = youTube.allowedChannels match {
+      case Nil => youTube.channels
```
Reettaphant: Why would the youtube channels not exist?

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -42,15 +49,22 @@ class VideoData extends React.Component {
...
             </ManagedField>
             <ManagedField
               fieldLocation="description"
-              name="Standfirst (YouTube description)"
+              name={
+                notOnManagedChannel
+                  ? 'Standfirst'
+                  : 'Standfirst (YouTube description)'
+              }
+              placeholder="No Description"
```
Reettaphant: These placeholders no longer need to get passed to the form fields.

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -70,7 +85,7 @@ class VideoData extends React.Component {
 
             <ManagedField
               fieldLocation="byline"
-              name="Byline"
+              name="Byline Tags"
```
Reettaphant: This should be just `Byline`

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -62,6 +76,7 @@ class VideoData extends React.Component {
               fieldLocation="trailText"
               derivedFrom={this.props.video.description}
               name="Trail Text"
+              placeholder="No Trail Text"
```
Reettaphant: Not needed

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -98,10 +113,10 @@ class VideoData extends React.Component {
             <ManagedField
               fieldLocation="source"
               name="Video Source"
+              placeholder="No video source"
```
Reettaphant: Not needed

public/video-ui/src/components/VideoUpload/AddAssetFromURL.js
```diff
@@ -0,0 +1,50 @@
...
+        <div className="video__detailbox__header__container">
+          <header className="video__detailbox__header">Asset URL</header>
+        </div>
+        <div className="form__group">
+          <div className="form__row">
+            <div>
+              <input
+                className="form__field"
+                type="text"
+                placeholder="Paste YouTube URL here"
```
Reettaphant: If you want to keep this, you'll have to override the generic one in managed field.

public/video-ui/src/components/VideoUpload/AddYouTubeAsset.js
```diff
@@ -0,0 +1,97 @@
...
+
+  render() {
+    const { video, channels, startUpload } = this.props;
+    const hasAssets = video.assets.length !== 0;
+
+    const isHosted = video.category === 'Hosted';
+    const isManaged = channelAllowed(video, channels);
+
+    const editable =
+      !hasAssets && (!video.channelId || !!video.youtubeCategoryId);
```
Reettaphant: I think the check here should be checking for youtube assets rather than all assets.

public/video-ui/src/components/VideoUpload/AddYouTubeAsset.js
```diff
@@ -0,0 +1,97 @@
...
+        <div className="video__detailbox__header__container">
+          <header className="video__detailbox__header">
+            YouTube Video
+          </header>
+        </div>
+        <div className="form__group">
+          <ManagedForm
+            data={video}
+            updateData={this.props.saveVideo}
+            editable={editable}
```
Reettaphant: Does this prevent from the youtube category as well? We should be able to change it.

app/controllers/Youtube.scala
mbarton: If there's no `allowedChannels` in config then the tool should work with all channels for the content owner

public/video-ui/src/components/VideoData/VideoData.js
mbarton: Yep sorry about that. merge mess-up

public/video-ui/src/components/VideoUpload/AddAssetFromURL.js
mbarton: Not sure why? It's working fine at the moment

public/video-ui/src/components/VideoUpload/AddYouTubeAsset.js
mbarton: Agreed. I've split them into two ManagedForms

public/video-ui/src/components/VideoUpload/AddAssetFromURL.js
Reettaphant: Ah ok, I remembered incorrectly!

PR #530, opened 2017-07-11
endpoint for Commercial diagnosis

akash1810:
As part of the Commercial training for selling YouTube, one of the troubleshooting topics is: "Why isn't this video showing in DfP?".

One way to answer that is to login to the YouTube CMS and find the video. However this requires adding users to the CMS. This endpoint will call the YouTube Partner API and display the advertising options for a video and error if the video isn't claimed or isn't owned by the content owner.

The endpoint is `/api/youtube/video-info/:id` and is Panda Authed.

As part of this, I've refactored YouTubeClaims as a trait, following the YouTubeVideos pattern. I've also simplified the `YouTubeChannel` case class, removing `logo` as its never used.

Example response:

```json
{
  "video": {
    "id": "cDoV2d4cMfI",
    "title": "Mohamed Salah: I joined Liverpool to win trophies",
    "duration": 41,
    "publishedAt": "2017-06-23T12:47:02.000+01:00",
    "privacyStatus": "public",
    "tags": [
      "Mohamed Salah",
      "Salah",
      "Liverpool",
      "Premier League",
      "Football",
      "gdnpfpsportfootball",
      "Sport"
    ],
    "contentBundleTags": [
      "gdnpfpsportfootball"
    ],
    "channel": {
      "id": "UCNHqb1IRxQ5WBsWtO53JL2g",
      "title": "Guardian Football"
    }
  },
  "advertising": {
    "id": "cDoV2d4cMfI",
    "adFormats": [
      "standard_instream",
      "trueview_instream",
      "product_listing",
      "display"
    ],
    "adBreaks": [
      "preroll"
    ]
  }
}
```

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/package.scala
+++ b/common/src/main/scala/com/gu/media/youtube/package.scala
@@ -21,43 +21,7 @@ package object youtube {
       YouTubeVideoCategory(category.getId.toInt, category.getSnippet.getTitle)
     }
   }
-
-  case class YouTubeAdFormat(format: String, description: String)
-
-  object YouTubeAdFormat {
-    implicit val reads: Reads[YouTubeAdFormat] = Json.reads[YouTubeAdFormat]
-    implicit val writes: Writes[YouTubeAdFormat] = Json.writes[YouTubeAdFormat]
-
-    def build(format: String): YouTubeAdFormat = {
-      // definitions lifted from https://developers.google.com/youtube/partner/docs/v1/videoAdvertisingOptions#adFormats[]
-      val description = format match {
-        case "display"              => """Appears to the right of the feature video and above the video suggestions list
-                                      |on www.youtube.com. Not shown on O&O platforms."""
-        case "long"                 => """Long video ads which would be enabled to run for videos that can show standard
-                                      |in-stream ads. Requires standard_instream ads to be enabled."""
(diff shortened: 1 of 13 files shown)


common/src/main/scala/com/gu/media/youtube/package.scala
```diff
@@ -23,26 +21,80 @@ package object youtube {
...
+     title: String,
+     duration: Long,
+     publishedAt: DateTime,
+     privacyStatus: String,
+     tags: Seq[String],
+     contentBundleTags: Seq[String],
+     channel: YouTubeChannel
+  )
+
+  object YouTubeVideo {
```
Reettaphant: Maybe we could call this something else? The youtube videos trait already has a `getVideo` method that returns objects of type `Video`so this feels a bit confusing.

akash1810: Yep, I was struggling to think of an alternative name. We already have `YouTubeChannel` and `YouTubeVideoCategory` so was following this pattern.

`Video` and `YouTubeVideo` are different types at least...

Reettaphant: Could it be `YouTubeVideoDetail`?

akash1810: 👍  done.

PR #533, opened 2017-07-18
Video poster size

Reettaphant:
Fixes the video poster ratio to 16 * 9. 

Not sure if using viewport could cause problems later, but I set the width of the poster image container as well so that they always have the same width. 

I didn't want to use jquery and other approaches suggested [here](https://stackoverflow.com/questions/5445491/height-equal-to-dynamic-width-css-fluid-layout) caused the whole player to disappear. 
screen shot 2017-07-18 at 08 47 58


akash1810:
I think this could be done with flex-box...

Reettaphant:
Not sure about flex-box but this seems to work. It is almost the correct ratio, I'm not sure how important this is so I could keep playing with decimal points to get it even closer.

PR #534, opened 2017-07-19
Pick composer image

Reettaphant:
Allow for picking a composer trail image of the correct crop size in mam and add this trail image to the composer video page.

Depends on: 

https://github.com/guardian/grid/pull/2072
https://github.com/guardian/flexible-content/pull/2858
https://github.com/guardian/atom-maker/pull/30

### DIFF
--- a/app/model/Image.scala
+++ b/app/model/Image.scala
@@ -13,23 +13,26 @@ object ImageAssetDimensions {
   def fromThrift(dim: ThriftImageAssetDimensions) = ImageAssetDimensions(dim.height, dim.width)
 }
 
-case class ImageAsset(mimeType: Option[String], file: String, dimensions: Option[ImageAssetDimensions], size: Option[Long]) {
-  def asThrift = ThriftImageAsset(mimeType, file, dimensions.map(_.asThrift), size)
+case class ImageAsset(mimeType: Option[String], file: String, dimensions: Option[ImageAssetDimensions],
+                      size: Option[Long], aspectRatio: Option[String]) {
+  def asThrift = ThriftImageAsset(mimeType, file, dimensions.map(_.asThrift), size, aspectRatio)
 }
 
 object ImageAsset {
   implicit val imageAssetFormat = Jsonx.formatCaseClass[ImageAsset]
 
-  def fromThrift(imgAsset: ThriftImageAsset) = ImageAsset(imgAsset.mimeType, imgAsset.file, imgAsset.dimensions.map(ImageAssetDimensions.fromThrift), imgAsset.size)
+  def fromThrift(imgAsset: ThriftImageAsset) = ImageAsset(imgAsset.mimeType, imgAsset.file,
+    imgAsset.dimensions.map(ImageAssetDimensions.fromThrift), imgAsset.size, imgAsset.aspectRatio)
 }
 
-case class Image(assets: List[ImageAsset], master: Option[ImageAsset], mediaId: String) {
-  def asThrift = ThriftImage(assets.map(_.asThrift), master.map(_.asThrift), mediaId)
+case class Image(assets: List[ImageAsset], master: Option[ImageAsset], mediaId: String, source: Option[String]) {
+  def asThrift = ThriftImage(assets.map(_.asThrift), master.map(_.asThrift), mediaId, source)
 }
 
 object Image {
   implicit val imageFormat = Jsonx.formatCaseClass[Image]
 
-  def fromThrift(img: ThriftImage) = Image(img.assets.map(ImageAsset.fromThrift).toList, img.master.map(ImageAsset.fromThrift), img.mediaId)
+  def fromThrift(img: ThriftImage) =
+    Image(img.assets.map(ImageAsset.fromThrift).toList, img.master.map(ImageAsset.fromThrift), img.mediaId, img.source)
 }
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   val awsVersion = "1.11.125"
   val pandaVersion = "0.4.0"
   val mockitoVersion = "2.0.97-beta"
-  val atomMakerVersion = "1.0.5-SNAPSHOT"
+  val atomMakerVersion = "1.1.2"
   val slf4jVersion = "1.7.21"
   val typesafeConfigVersion = "1.3.0" // to match what we get from Play transitively
   val scanamoVersion = "0.9.1" // to match what we get from atom-publisher-lib transitively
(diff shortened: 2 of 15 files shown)


project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   val awsVersion = "1.11.125"
   val pandaVersion = "0.4.0"
   val mockitoVersion = "2.0.97-beta"
-  val atomMakerVersion = "1.1.0"
+  val atomMakerVersion = "1.0.5-SNAPSHOT"
```
Reettaphant: This will be updated once the new version of atom maker is released

public/video-ui/src/components/utils/GridImageSelect.js
```diff
@@ -56,56 +56,36 @@ export default class GridEmbedder extends React.Component {
...
-          </Modal>
-        </div>
-      );
+
+    let gridUrl;
+
+    if (this.props.posterImage && this.props.isComposerImage) {
+      const imageIdParts = this.props.posterImage.mediaId.split("/");
+      const imageGridId = imageIdParts[imageIdParts.length - 1];
+      gridUrl = this.props.gridUrl + '/images/' + imageGridId + '/crop?cropType=composer';
```
akash1810: could `cropType` here be `landscape` to match the nomenclature of grid?

public/video-ui/src/components/utils/GridImageSelect.js
```diff
@@ -56,56 +56,36 @@ export default class GridEmbedder extends React.Component {
...
-            </div>
-          </div>
-
-          <Modal isOpen={this.state.modalOpen} dismiss={this.closeModal}>
-            <iframe className="gridembedder__modal" src={this.props.gridUrl} />
-          </Modal>
-        </div>
-      );
+
+    let gridUrl;
```
akash1810: could this logic be pushed to a function to avoid the `let`?

```js
function getGridUrl() {
  if (this.props.posterImge & this.props.isComposerImage) {
    const imageIdParts = this.props.posterImage.mediaId.split("/");
    const imageGridId = imageIdParts[imageIdParts.length - 1];
    return `${this.props.gridUrl}/images/${imageGridId}/crop?cropType=composer`;
  }
  else {
    return `${this.props.gridUrl}?cropType=video`;
  }
}

const gridUrl = getGridUrl();
```

public/video-ui/src/util/parseGridMetadata.js
```diff
@@ -10,22 +10,69 @@ function parseMimeType(mimeType) {
...
+function parseAsset(asset, aspectRatio) {
   return {
     file: asset.secureUrl,
     mimeType: parseMimeType(asset.mimeType),
     size: asset.size,
+    aspectRatio: aspectRatio,
     dimensions: {
       width: asset.dimensions.width,
-      height: asset.dimensions.height
+      height: asset.dimensions.height,
```
akash1810: is the linter happy with the trailing `,`?

public/video-ui/src/util/parseGridMetadata.js
```diff
@@ -10,22 +10,69 @@ function parseMimeType(mimeType) {
...
+    mediaId: cropData.specification.uri,
+    source: imageData.data.metadata.credit
+  };
 }
 
-export function parseImageFromGridCrop(cropData) {
+export function parseComposerDataFromImage(image, trail) {
+
+  const urlParts = image.mediaId.split('/');
+  const mediaId = urlParts[urlParts.length - 1];
```
akash1810: this is really similar to https://github.com/guardian/media-atom-maker/pull/534/files#diff-49e3809db1c472d315c7c0d48cccf491R67, could we abstract it into a shared function?

public/video-ui/src/util/parseGridMetadata.js
```diff
@@ -10,22 +10,69 @@ function parseMimeType(mimeType) {
...
+    };
+  }
+
+  function getComposerMasterAsset(asset) {
+    const composerAsset = getComposerAsset(asset);
+    composerAsset.isMaster = 'true';
+    return composerAsset;
+  }
+
+  const tempDiv = document.createElement('div');
```
akash1810: why do we need to do this?

Reettaphant: This was the best way I found of extracting the text from the trail text (which is html) without using jquery or complicated regexes. Happy to change it if there is saner way of doing it.

public/video-ui/src/components/FormFields/ImageSelector.js
```diff
@@ -19,33 +19,15 @@ class ImageSelector extends React.Component {
...
-        </div>
-      );
-    } else {
-      return (
-        <div className="form__row">
-          <div className="form__imageselect">
-            {this.renderImage()}
-          </div>
+    return (
+      <div className="form__row">
```
Reettaphant: Because it wasn't used for anything, just always set to true. The image picker only has an edit mode, we don't have to open up the form to be able to pick a poster image.

public/video-ui/src/components/utils/GridImageSelect.js
Reettaphant: Yes, I'll have to change that in the grid as well.

public/video-ui/src/util/parseGridMetadata.js
Reettaphant: Probably since the build passed but I seemed to have removed that anyway on my branch without pushing up the changes...

app/model/Image.scala
```diff
@@ -13,23 +13,26 @@ object ImageAssetDimensions {
   def fromThrift(dim: ThriftImageAssetDimensions) = ImageAssetDimensions(dim.height, dim.width)
 }
 
-case class ImageAsset(mimeType: Option[String], file: String, dimensions: Option[ImageAssetDimensions], size: Option[Long]) {
-  def asThrift = ThriftImageAsset(mimeType, file, dimensions.map(_.asThrift), size)
+case class ImageAsset(mimeType: Option[String], file: String, dimensions: Option[ImageAssetDimensions],
+                      size: Option[Long], aspectRatio: Option[String]) {
```
Reettaphant: Not sure. Currently it is possible to pick something different from the grid by navigating away from the cropping page and selecting a differently sized image. Maybe we shouldn't enforce it on the server since we are not enforcing on the client, only strongly suggesting that these crops should be used.

Reettaphant:
I think we decided to never disable the create video page button due to missing fields to allow flexibility for use of the tool and because having different buttons disabled for different reasons might be confusing. If this becomes a problem we can always add this in later.

PR #536, opened 2017-07-21
Draggable tags

Reettaphant:
- Allow keyword tags to be sorted by dragging them around once they have been added.

### DIFF
--- a/public/video-ui/src/components/FormFields/TagPicker.js
+++ b/public/video-ui/src/components/FormFields/TagPicker.js
@@ -5,6 +5,7 @@ import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
 import { keyCodes } from '../../constants/keyCodes';
 import UserActions from '../../constants/UserActions';
 import TagTypes from '../../constants/TagTypes';
+import DragSortableList from 'react-drag-sortable';
 
 export default class TagPicker extends React.Component {
   state = {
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -298,6 +308,76 @@ export default class TagPicker extends React.Component {
     }
   }
 
+  onSort = (sortedList) => {
+    const newTagValues = [];
+
+    sortedList.forEach(sortedValue => {
```
akash1810: this could be a [`reduce`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce)

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -298,6 +308,76 @@ export default class TagPicker extends React.Component {
...
+
+      //For each component in the list of dragged elements,
+      //we have to extract the name of the tag it represents.
+      const child = sortedValue.content.props.children;
+
+      let tagTitle;
```
akash1810: minor: could be a ternary to avoid the `let`

Reettaphant: But then it would end up being a really long statement...

akash1810: True, though we could break it down a little:

```js
const child = child[0];
cons tagTitle = typeof child === 'string' ? child : child.props.children[0];
```

You're call though!

PR #537, opened 2017-07-21
fix integration tests

akash1810:
The json blob we were POSTing wasn't deserialising to the new `MediaAtomBeforeCreation` case class so the tests were failing. Adding the missing fields and we're winning!


philmcmahon:
👍

akash1810:
related to https://github.com/guardian/media-atom-maker/pull/532

PR #538, opened 2017-07-23
min duration for ads

akash1810:
As requested by Commercial, automatically block ads when a video is less than 30 seconds.

We only do this on Publish rather than on Update and Publish so as to keep interaction with an external API to a minimum, thus keeping Update fast.

Displays a tooltip too, for clarity to the User.

image

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -40,8 +40,11 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youtu
 
     getActiveAsset(previewAtom) match {
       case Some(asset) if asset.platform == Youtube =>
-        val atomWithDuration = previewAtom.copy(duration = youtube.getDuration(asset.id))
-        updateYouTube(atomWithDuration, asset).map(atomWithYoutubeUpdates => {
+        val duration = youtube.getDuration(asset.id)
+        val blockAds = if (duration.getOrElse(0L) < youtube.minDurationForAds) false else previewAtom.blockAds
+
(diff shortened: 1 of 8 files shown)


akash1810:
[don't monetize videos <30 seconds](https://trello.com/c/oGMlKA82/53-dont-monetize-videos-30-seconds)

app/model/commands/PublishAtomCommand.scala
```diff
@@ -40,8 +40,11 @@ case class PublishAtomCommand(id: String, override val stores: DataStores, youtu
 
     getActiveAsset(previewAtom) match {
       case Some(asset) if asset.platform == Youtube =>
-        val atomWithDuration = previewAtom.copy(duration = youtube.getDuration(asset.id))
-        updateYouTube(atomWithDuration, asset).map(atomWithYoutubeUpdates => {
+        val duration = youtube.getDuration(asset.id)
+        val blockAds = if (duration.getOrElse(0L) < youtube.minDurationForAds) false else previewAtom.blockAds
```
mbarton: This would automatically block ads on videos without a duration?

akash1810: yes...

mbarton: Yeah I suppose if YouTube doesn't give us a duration what decision could we make anyway

akash1810: Indeed. `youtube.getDuration(asset.id)` returns `None` if there was a problem getting the video from the YT Data API. The chances of a call to the Partner API succeeding at this point is very low...

PR #541, opened 2017-07-27
use babel-loader

akash1810:
so we can use .babelrc rather than inline querystrings

also uses [`cacheDirectory`](https://github.com/babel/babel-loader) for faster builds in DEV

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,11 @@
 {
-  "env": {
-    "test": {
-      "plugins": ["transform-es2015-modules-commonjs"]
-    }
-  }
+  "presets": [
+    "es2015",
+    "react"
+  ],
+  "plugins": [
+    "transform-object-assign",
+    "transform-class-properties",
+    "transform-es2015-modules-commonjs"
+  ]
 }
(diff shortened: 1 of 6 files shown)


akash1810:
this is all in an attempt to use `async/await` 😈

.babelrc
```diff
@@ -1,7 +1,11 @@
 {
-  "env": {
-    "test": {
-      "plugins": ["transform-es2015-modules-commonjs"]
-    }
-  }
+  "presets": [
+    "es2015",
```
gtrufitt: Maybe using [`babel-preset-env`](http://2ality.com/2017/02/babel-preset-env.html) with our targets might bring our bundle size down?

You may not need `transform-es2015-modules-commonjs` as I think that's the default in preset-env? though I would have expected webpack to handle the module side of things?

akash1810: Tried using `babel-preset-env` and the build is still 2.24MB. I suspect it because of the [browser versions targeted](https://github.com/guardian/media-atom-maker/commit/24b1d8fc572daa38b147bcc59a004961e62e3f14); this list was provided from ESD.

PR #542, opened 2017-07-28
add series tags to keyword tags list

Reettaphant:
- Series tags appear at the top of the composer keyword tag list
screen shot 2017-07-28 at 09 00 22

### DIFF
--- a/public/video-ui/src/services/capi.js
+++ b/public/video-ui/src/services/capi.js
@@ -35,15 +35,18 @@ export default class ContentApi {
     });
   }
 
-  static getTagsByType(query, type) {
-    if (query === '*') {
+  static getTagsByType(query, types) {
+
+    return Promise.all(types.map(type => {
+      if (query === '*') {
+        return pandaReqwest({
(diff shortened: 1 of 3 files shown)


public/video-ui/src/services/capi.js
```diff
@@ -35,15 +35,18 @@ export default class ContentApi {
...
   }
 
-  static getTagsByType(query, type) {
-    if (query === '*') {
+  static getTagsByType(query, types) {
+
+    return Promise.all(types.map(type => {
+      if (query === '*') {
+        return pandaReqwest({
+          url: `${ContentApi.proxyUrl}/tags?page-size=100&type=${type}` //TODO this is likely to change based on CAPI work to search by prefix on webTitle
```
Reettaphant: Not sure if we want to adjust the page size here. If we have a search term we only get a handful of series tags but here when we are querying for all tags, we will end up with 200 tags (100 keyword and 100 series)

akash1810: I think this is ok as the list should scroll.

PR #543, opened 2017-07-28
smaller web assets

akash1810:
use digest for hashing and gzip for compression

this gets the js down from ~2mb to ~400kb

TODO:
- [x] make sure the angular app still works


akash1810:
image

PR #546, opened 2017-08-04
Tag picker fixes

Reettaphant:
out
- Apologies for the long PR: by the time I realised I needed to rearrange the tag picker for this, I had already started on many of the improvements. 

- This PR makes many improvements to the usability of the tag pickers
- It hides tag search results when user is no longer focuses on the input
- It allows for navigating up and down search results using arrow keys
- It fixes padding issues in the byline picker
- It changes the commissioning and keyword tag picker presentation to be more in line with what is on composer (showing selected tag as a vertical list, not on the input element itself)
- It also breaks the tag picker into multiple components to make it more usable and to allow for more customisation.

### DIFF
--- a/public/video-ui/src/components/CapiSearch/CapiSearch.js
+++ b/public/video-ui/src/components/CapiSearch/CapiSearch.js
@@ -1,0 +1,44 @@
+ import React from 'react';
+import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
+
+export default class CapiSearch extends React.Component {
+
+  renderTags(tag) {
+    const addTag = () => {
+
+      const valueWithoutStringDupes = this.props.removeDupes(
+        tag,
+        this.props.tagValue
+      );
+
+      const newFieldValue = valueWithoutStringDupes.concat([tag]);
+
+      this.props.selectNewTag(newFieldValue);
+
+    };
+
+    return (
+      <a
+        className="form__field__tags"
+        key={tag.id}
+        title={tag.id}
+        onClick={addTag}
+      >
+        {' '}{tag.webTitle}{' '}
+      </a>
+    );
+  }
+
+  render() {
+
+    if (this.props.capiTags.length !== 0 && this.props.showTags) {
+      return (
+        <div className="form__field__tags" onMouseDown={this.props.tagsToVisible}>
+          {this.props.capiTags.map(tag => this.renderTags(tag))}
+        </div>
+      );
+    }
+
+    return null;
+  }
+}
--- a/public/video-ui/src/components/CapiSearch/CapiSearch.js
+++ b/public/video-ui/src/components/CapiSearch/CapiSearch.js
@@ -7,7 +7,10 @@ export default class CapiSearch extends React.Component {
     if (this.props.selectedTagIndex !== nextProps.selectedTagIndex && nextProps.selectedTagIndex !== null) {
       const selectedTag = this.props.capiTags[nextProps.selectedTagIndex];
       const listNode = this.refs.list;
-      listNode.scrollTop = 35 * (nextProps.selectedTagIndex === 0 ? 0 : nextProps.selectedTagIndex - 1)
+      const elementHeight = listNode.children[0].offsetHeight;
+      if (listNode) {
+        listNode.scrollTop = elementHeight * (nextProps.selectedTagIndex === 0 ? 0 : nextProps.selectedTagIndex - 1)
+      }
     }
   }
 
--- a/public/video-ui/src/components/CapiSearch/CapiSearch.js
+++ b/public/video-ui/src/components/CapiSearch/CapiSearch.js
@@ -17,10 +17,9 @@ export default class CapiSearch extends React.Component {
   renderTags(tag, index) {
 
     const getTagClassName = () => {
-      const name = "form__field__tags" + (index === this.props.selectedTagIndex ? " form__field__tags--selected" : "");
-
       return "form__field__tags" + (index === this.props.selectedTagIndex ? " form__field__tags--selected" : "");
     }
+
     const addTag = () => {
 
       const valueWithoutStringDupes = this.props.removeDupes(
@@ -35,14 +34,14 @@ export default class CapiSearch extends React.Component {
     };
 
     return (
-      <a
+      <li
         className={getTagClassName()}
         key={tag.id}
         title={tag.id}
         onClick={addTag}
       >
         {' '}{tag.webTitle}{' '}
-      </a>
+      </li>
     );
   }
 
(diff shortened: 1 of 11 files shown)


public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -34,359 +43,264 @@ export default class TagPicker extends React.Component {
...
-      })
-    ) {
-      return {
-        id: this.state.inputString,
-        webTitle: this.state.inputString
-      };
-    }
-    return [];
+  removeFn = (tag) => {
+    const newFieldValue = this.state.tagValue.filter(oldField => {
```
akash1810: `filter` returns a list, does this work as expected?

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -1,19 +1,28 @@
...
+
   state = {
     capiTags: [],
-    inputString: '',
-    lastAction: UserActions.other,
     tagValue: [],
-    capiUnavailable: false
+    capiUnavailable: false,
+    showTags: true,
+    tagsVisible: false,
```
akash1810: it looks like `showTags` and `tagsVisible` are related and can be merged into a single property

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -34,359 +43,264 @@ export default class TagPicker extends React.Component {
...
-  renderTags(tag) {
-    const addTag = () => {
-      const valueWithoutStringDupes = removeStringTagDuplicates(
-        tag,
-        this.state.tagValue
-      );
-      const newFieldValue = valueWithoutStringDupes.concat([tag]);
+    if (e.keyCode === keyCodes.enter && this.state.selectedTagIndex !== null) {
+      const newTag = this.state.capiTags[this.state.selectedTagIndex];
+      const valueWithoutDupes = this.props.tagType === TagTypes.contributor ? removeStringTagDuplicates(newTag, this.state.tagValue) : removeTagDuplicates(newTag, this.state.tagValue);
```
akash1810: the linter would cry about this line length 😜

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -34,359 +43,264 @@ export default class TagPicker extends React.Component {
...
+      const newFieldValue = valueWithoutDupes.concat([newTag]);
 
       this.setState({
-        inputString: ''
+        selectedTagIndex: null,
+        inputClearCount: this.state.inputClearCount + 1
```
akash1810: what's `inputClearCount` used for?

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -34,359 +43,264 @@ export default class TagPicker extends React.Component {
...
-  renderCapiUnavailable() {
-    if (this.state.capiUnavailable) {
-      return (
-        <span className="form__field--external__error">
-          Tags are currently unavailable
-        </span>
-      );
-    }
+      newTagValues.push(tagValue);
+      return newTagValues;
```
akash1810: minor: could use a [spread operator](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator) here:

```js
return [...newTagValues, tagValue];
```

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -34,359 +43,264 @@ export default class TagPicker extends React.Component {
...
+    }
   }
 
-  renderValue = (field, i) => {
-    const removeFn = () => {
-      const newFieldValue = this.state.tagValue.filter(oldField => {
-        return field.id !== oldField.id;
-      });
+  onSort = (sortedList) => {
+    const newTagValues = sortedList.reduce((newTagValues, sortedValue) => {
```
akash1810: could we use a distinct name for the accumulator for clarity?

public/video-ui/src/components/FormFields/TagPicker.js
Reettaphant: Do you mean two different properties in the same object?

public/video-ui/src/components/FormFields/TagPicker.js
Reettaphant: Yes, `onUpdate` converts the list into a string before saving the value but the tagpicker handles the tags as lists.

public/video-ui/src/components/FormFields/TagPicker.js
Reettaphant: To send a message from the `TagPicker` to its children to clear the input string because a new tag has been selected by navigating the tag list with arrow keys. If the count changes, the children know they should set the input string to an empty string.

public/video-ui/src/components/CapiSearch/CapiSearch.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+  }
+
+  renderTags(tag, index) {
+
+    const getTagClassName = () => {
+      const name = "form__field__tags" + (index === this.props.selectedTagIndex ? " form__field__tags--selected" : "");
+
+      return "form__field__tags" + (index === this.props.selectedTagIndex ? " form__field__tags--selected" : "");
```
Reettaphant: I like this pattern. It saves a lot of typing. This would only be worth doing I think it were abstracted out to a utility function which adds a new classname to the end of the string if a condition is satisfied. But maybe this would be another pr?

public/video-ui/src/components/CapiSearch/CapiUnavailable.js
```diff
@@ -0,0 +1,17 @@
+ import React from 'react';
+
+ export default class CapiUnavailable extends React.Component {
+
+  render() {
+    if (this.props.capiUnavailable) {
+      return (
+        <div className="form__field--external__error">
```
Reettaphant: Yes, it doesn't make any sense. I'll fix it

public/video-ui/src/components/FormFields/TagPicker.js
akash1810: from what I understood from the code, `showTags` is always the inverse of `tagsVisible` which made me think it could be one boolean value?

Reettaphant: It's not. `tagsVisible` is used to check if `onBlu` event was triggered by clicking on one of the pickers children (this should not hide the tags) or elsewhere (this should hide the tags). `showTags` is set to true/false depending on whether or not we should display the tag list. I should probably rename that variable because it is not very descriptive.

PR #549, opened 2017-08-04
always edit the main block

Reettaphant:
- If a composer video page is live, any changes to media atom fields should update the live composer page on publish

- Just editing the live data of a published composer page seems to be enough to accomplish this.

- I've added checks and warning messages to make sure we don't create or update composer when required composer fields are not filled in. The warning messages look pretty ugly but since the logic is a little bit complicated (publishing is only disabled if a composer page exists, otherwise creation of composer page is restricted), I didn't want to rely on only exclamation marks on buttons or tooltips. @akash1810 

screen shot 2017-08-09 at 08 15 10

### DIFF
--- a/public/video-ui/src/components/ReactApp.js
+++ b/public/video-ui/src/components/ReactApp.js
@@ -72,6 +72,7 @@ class ReactApp extends React.Component {
           usages={this.props.usages}
           presenceConfig={this.props.config.presence}
           isTrainingMode={this.props.config.isTrainingMode}
+          formFieldsWarning={this.props.formFieldsWarning}
         />
         {this.props.error
           ? <div className="error-bar">{this.props.error}</div>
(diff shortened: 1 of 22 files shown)


public/video-ui/src/components/ReactApp.js
```diff
@@ -109,7 +110,8 @@ function mapStateToProps(state) {
     checkedFormFields: state.checkedFormFields,
     videoEditOpen: state.videoEditOpen,
     usages: state.usage,
-    config: state.config
+    config: state.config,
```
akash1810: ah! that's where `state.config` is 😄

public/video-ui/src/util/validateField.js
```diff
@@ -18,10 +21,13 @@ const validateField = (
     );
   }
 
-  if (isDesired && !fieldValue) {
+  if (
+    isDesired &&
+    (!fieldValue || (fieldValue instanceof Array && fieldValue.length === 0))
```
akash1810: minor: could do `Array.isArray(fieldValue)` here instead of `instanceof`

PR #550, opened 2017-08-04
link to logs in slack message

akash1810:
link to the logs in the slack message sent when the integration tests fail


akash1810:
ideally we'd use attachments, but alas json manipulation in bash is 🤢

PR #554, opened 2017-08-09
No lists and links in trail scribe editor

Reettaphant:
- Only bold and italics should be allowed for scribe trail text editor
- Copy to trail button should be removed because it's not being used much and stripping lists and links out on copy is too much work for functionality which does not get use
- I noticed that the `textArea` component was still here, I've removed it because we are now using the scribe editor instead.


public/video-ui/src/components/FormComponents/CopyTrailButton.js
```diff
@@ -1,17 +0,0 @@
-import React from 'react';
```
akash1810: 👌

public/video-ui/src/components/FormFields/ScribeEditor.js
```diff
@@ -184,29 +158,37 @@ export class ScribeEditor extends React.Component {
   }
 
   configureScribe() {
+
+    const allKeyboardShortcuts = {
+      bold: function(event) {
+        return event.metaKey && event.keyCode === 66;
```
akash1810: could we add these [here](https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/constants/keyCodes.js)?

Reettaphant: Yes!

public/video-ui/src/components/FormFields/ScribeEditor.js
```diff
@@ -174,39 +144,38 @@ export class ScribeEditor extends React.Component {
...
-      this.refs.editor.innerHTML = nextProps.copiedValue;
-      this.setState({
-        copiedValue: nextProps.copiedValue
-      });
+  configureScribe() {
+
+    const allKeyboardShortcuts = {
+      bold: function(event) {
+        return event.metaKey && event.keyCode === keyCodes.b
+      }, // b
```
akash1810: probably don't need these comments anymore...

PR #557, opened 2017-08-10
set composer link text

akash1810:


Reettaphant:
Is this field being updated in composer on page create as well?

akash1810:
Not tested 😬 but did look at the request made by composer to set this field. Will test this morning.

akash1810:
Tested now and it works 😄

PR #559, opened 2017-08-11
training mode

akash1810:
Add a training mode to the app. When in this mode:
- You can only use the dev youtube channel
- Created composer pages are embargoed indefinitely

# Why?
- Prevent test videos landing on public, heavily subscribed YouTube channels
- Prevent video page launch
- Publishing is scary!

# User flow
- To enter training mode, visit `/training/on`
- To exit training mode, visit `/training/off`
- Links are on `/training` and on `/help`

# UX
Training mode is denoted by a stripy header:
image

Additionally, `/training` has some information:
image

# How it works
Visiting `/training/on` sets session storage in a cookie and its value is injected onto the DOM for client-side use.

Visiting `/training/off` sets the session storage value to `false`.

After `/training/on` and `/training/off` has set session storage, they redirect to `/` for normal app operations... this results in a small "loading..." flash whilst the React app is loaded, but I think that's fine for now (we could always style that loading screen...).

### DIFF
--- a/public/video-ui/src/components/Training/Training.js
+++ b/public/video-ui/src/components/Training/Training.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { getStore } from '../../util/storeAccessor';
+
+export default class Training extends React.Component {
+  render() {
+    const trainingModeState = getStore().getState().config.isTrainingMode
+      ? 'on'
+      : 'off';
+
+    return (
+      <div className="container">
+        <h1>Training mode is currently {trainingModeState}</h1>
+        <ul>
+          <li>
+            <a href="/training/on">Turn on</a>
+          </li>
+          <li>
+            <a href="/training/off">Turn off</a>
+          </li>
+        </ul>
+      </div>
+    )
+  }
+}
--- a/public/video-ui/src/services/VideosApi.js
+++ b/public/video-ui/src/services/VideosApi.js
@@ -258,5 +258,22 @@ export default {
       }
       return '';
     });
+  },
+
+  preventPublication(composerId, composerUrl) {
+    function doEmbargoIndefinitely(stage) {
+      return pandaReqwest({
+        url: `${composerUrl}/api/content/${composerId}/${stage}/settings/embargoedIndefinitely`,
+        method: 'put',
+        crossOrigin: true,
+        withCredentials: true,
+        data: `"true"`
+      });
+    }
+
+    return Promise.all([
+      doEmbargoIndefinitely('preview'),
+      doEmbargoIndefinitely('live')
+    ]);
   }
 };
--- a/public/video-ui/src/util/getComposerData.js
+++ b/public/video-ui/src/util/getComposerData.js
@@ -1,11 +1,12 @@
 import { parseComposerDataFromImage } from './parseGridMetadata';
+import { getStore } from '../util/storeAccessor';
 
 function asBooleanString(value) {
   return value ? 'true' : 'false';
 }
 
 export function getComposerData(video) {
-  return [
+  const coreFields = [
     {
       name: 'headline',
       value: video.title,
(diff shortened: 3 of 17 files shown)


public/video-ui/styles/components/_advanced.scss
```diff
@@ -11,10 +11,9 @@
         align-items: center;
         padding: 5px 10px;
         width: 100%;
-        background: $color700Grey;
```
akash1810: use the background of the toolbar

public/video-ui/src/services/VideosApi.js
```diff
@@ -258,5 +258,22 @@ export default {
...
+  },
+
+  preventPublication(composerId, composerUrl) {
+    function doEmbargoIndefinitely(stage) {
+      return pandaReqwest({
+        url: `${composerUrl}/api/content/${composerId}/${stage}/settings/embargoedIndefinitely`,
+        method: 'put',
+        crossOrigin: true,
+        withCredentials: true,
+        data: `"true"`
```
akash1810: 🤢

public/video-ui/src/components/Training/Training.js
```diff
@@ -0,0 +1,24 @@
...
+    const trainingModeState = getStore().getState().config.isTrainingMode
+      ? 'on'
+      : 'off';
+
+    return (
+      <div className="container">
+        <h1>Training mode is currently {trainingModeState}</h1>
+        <ul>
+          <li>
+            <a href="/training/on">Turn on</a>
```
akash1810: cannot be `Link`s as they need to go back to the server

public/video-ui/src/util/getComposerData.js
```diff
@@ -77,6 +78,19 @@ export function getComposerData(video) {
...
   ];
+
+  const isTrainingMode = getStore().getState().config.isTrainingMode;
+
+  // block training content being published
+  const embargoedIndefinately = {
+    name: 'embargoedIndefinitely',
+    value: asBooleanString(true),
+    belongsTo: 'settings'
+  };
```
Reettaphant: Do we need to add this here? We are already embargoing the article on page creation, does this do anything other than also embargo it on subsequent publications?

public/video-ui/styles/layout/_topbar.scss
```diff
@@ -3,6 +3,16 @@ $topbarHeight: 50px;
...
   background: $color700Grey;
   border-bottom: 1px solid $greyBorderColor;
+
+  &--training-mode {
+    background: repeating-linear-gradient(
+        -45deg,
+        $color500Grey,
+        $color500Grey 10px,
+        $color800Grey 10px,
+        $color800Grey 20px
```
Reettaphant: Would adding opacity here make the title a little bit more visible when in training mode?

akash1810: Unfortunately opacity hits the all the elements 😢 . I've forced the background colour back instead:

image

public/video-ui/styles/layout/_topbar.scss
```diff
@@ -12,6 +26,9 @@ $topbarHeight: 50px;
   input {
     background-color: transparent;
     border: none;
+    &::placeholder {
+      color: $color400Grey;
```
akash1810: make the placeholder more legible

mbarton:
Might be a daft question but why not get people to use CODE for training?

akash1810:
> why not get people to use CODE for training?

CODE is a pre-prod environment, there's no guarentee it will be operational 100% of the time. Being operational is important particularly as we train overseas offices.

Also, having to remember a different domain is painful.

public/video-ui/src/services/VideosApi.js
Reettaphant: Alternatively you could change the creation method in composer to embargo if the embargo is set to true in the create request body (which should be sent automatically). But that would require making another (simple) pr.

akash1810: oooh yeah, that's a **lot** nicer! I'll do this as a secondary PR.

PR #560, opened 2017-08-11
copy youtube tags to composer keywords

Reettaphant:


Reettaphant:
@akash1810 This pr has now been updated. There is now a button a separate button for copying composer keywords over and composer keyword tags are matched with the bundle names. I couldn't get react-tooltip working for displaying the tooltip when hovering over the keyword copy button, I'll fix that in a separate pr.

Reettaphant:
Tooltip is now working here as well!

app/controllers/YoutubeTagController.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.gu.pandahmac.HMACAuthActions
+import play.api.libs.json.Json
+import play.api.mvc.Controller
+import com.gu.media.youtube.contentBundlingMap
+
+class YoutubeTagController(val authActions: HMACAuthActions) extends Controller {
+
+  import authActions.APIHMACAuthAction
+
+  def getById(id: String) = APIHMACAuthAction { implicit req =>
```
akash1810: why `APIHMACAuthAction` over `APIAuthAction`?

conf/routes
```diff
@@ -36,6 +36,7 @@ GET     /api2/pluto/projects/:id        controllers.PlutoProjectController.getPr
 POST    /api2/pluto/projects            controllers.PlutoProjectController.createProject()
 PUT     /api2/pluto/projects/:id        controllers.PlutoProjectController.updateProject(id)
 
+GET     /api2/youtube-tag/:id           controllers.YoutubeTagController.getById(id)
```
akash1810: Would `/api2/youtube/content-bundle-tag` be more intention revealing? Also, could this be kept in the `/api2/youtube/` namespace?

common/src/main/scala/com/gu/media/youtube/package.scala
```diff
@@ -11,6 +11,44 @@ import org.joda.time.DateTime
...
+  def contentBundlingMap: Map[String, String] = Map (
+    "uk" -> "gdnpfpnewsuk",
+    "us" -> "gdnpfpnewsus",
+    "au" -> "gdnpfpnewsau",
+    "world" -> "gdnpfpnewsworld",
+    "politics" -> "gdnpfpnewspolitics",
+    "opinion" -> "gdnpfpnewsopinion",
+    "football" -> "gdnpfpsportfootball",
+    "cricket" -> "gdnpfpsportcricket",
+    "rugby union" -> "gdnpfpsportrugbyunion",
```
akash1810: should this be `rugby-union` and `rugby-league`?

app/controllers/YoutubeTagController.scala
Reettaphant: No reason, just absent mindedness.

public/video-ui/src/services/KeywordsApi.js
```diff
@@ -0,0 +1,12 @@
+import { pandaReqwest } from './pandaReqwest';
+import ContentApi from './capi';
+
+export default {
+  composerTagToYouTube: (tagId) => {
+    const encodedId = encodeURIComponent(tagId)
+    return pandaReqwest({
+      url: '/api2/youtube/content-bundle/' + encodedId,
+      method: 'get'
```
akash1810: minor: I think we default as `get`

PR #564, opened 2017-08-14
remove video title from header

akash1810:
Numerous reasons for this:
- the title is already on the page in the video metadata form, so its use is questionable
- it renders with ellipses when the title is long so isn't that useful
- it keeps the header more in-line with other Tools, such as Composer
- it frees up space to add a "Track in Workflow" button

# Before
image

# After
image

### DIFF
--- a/public/video-ui/src/components/Header.js
+++ b/public/video-ui/src/components/Header.js
@@ -66,14 +66,6 @@ export default class Header extends React.Component {
     );
   }
 
-  renderVideoTitle() {
-    return (
-      <div className="flex-container">
-        <span className="header__video__title">{this.props.video.title}</span>
-      </div>
-    );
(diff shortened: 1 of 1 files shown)


akash1810:
[track in Workflow button](https://trello.com/c/0qyN0s3B/56-track-in-workflow-button)

public/video-ui/src/components/Header.js
```diff
@@ -66,14 +66,6 @@ export default class Header extends React.Component {
     );
   }
 
-  renderVideoTitle() {
-    return (
-      <div className="flex-container">
-        <span className="header__video__title">{this.props.video.title}</span>
```
akash1810: can't delete this class as its used on the `/upload` page

PR #566, opened 2017-08-14
move to css grid

akash1810:
partly to experiment with grid, partly because flexbox isn't great

Grid is great because:
- the `fr` unit
- `grid-gap` means we don't have to account for padding/margins when determining how much space we have for content (you do have to in flexbox)
- css becomes more expressive as the layout is literally written in css
- encourages thinking in components
- makes responsive layouts easier

There's a lot of dead css in the repo now, which I'll look at removing separately (hopefully there's some tooling for this)

Result:
image


paperboyo:
💎

PR #567, opened 2017-08-14
[to discuss] allow async/await code to be written

akash1810:
This adds 80kb to the build 😱  but it allows us to write pretty code!

Opted against using `babel-plugin-transform-regenerator` after reading http://madole.xyz/async-await-es7/

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,6 +4,14 @@
     "react"
   ],
   "plugins": [
+    ["transform-runtime", {
+      "polyfill": false,
+      "regenerator": true
+    }],
+    ["transform-async-to-module-method", {
+      "module": "bluebird",
+      "method": "coroutine"
+    }],
     "transform-object-assign",
     "transform-class-properties",
     "transform-es2015-modules-commonjs"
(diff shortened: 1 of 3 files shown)


mbarton:
[caniuse](http://caniuse.com/#search=await) is pretty much green for this now so is it worth the extra weight of the build?

That is of course assuming we don't have to support old Safari (which I bet we do)

gtrufitt:
If we do have to support old Safari... bluebird is MASSIVE. Maybe we should [transpile into native Promises](https://www.npmjs.com/package/babel-plugin-async-to-promises) and let polyfill.io to the hard work of supporting older browsers? 

Worth talking to Platform as we maintain our own polyfill io https://assets.guim.co.uk/polyfill.io/v2/polyfill.min.js?rum=0&features=es6,es7,es2017,default-3.6&flags=gated&callback=guardianPolyfilled but we probably don't want to call that one directly (maybe we can make it a seperate CDN endpoint for tools to use).

akash1810:
Other Tools officially [support Chrome and FF](https://github.com/guardian/flexible-content/blob/master/composer/src/js/services/helper/supported-browser-collection.js); I'll revisit this to go native...

paperboyo:
I’m pretty sure Grid doesn’t work correctly in Microsoft Safari 6.0. FWIW.

akash1810:
closing as getting old and was more experimental than anything...

PR #572, opened 2017-08-20
track in workflow button

akash1810:
# Screenshots
## When not tracked yet
image

## When tracked in Workflow:
image

TODO
- [x] screenshots
- [x] deploy to CODE
- [x] fix tests
- [x] detect production office


akash1810:
[track in Workflow button](https://trello.com/c/0qyN0s3B/56-track-in-workflow-button)

public/video-ui/src/reducers/workflowReducer.js
```diff
@@ -0,0 +1,49 @@
...
+      });
+    case 'WORKFLOW_STATUS_GET_RECEIVE':
+      return Object.assign({}, state, {
+        status: action.status || {}
+      });
+    case 'WORKFLOW_STATUS_NOT_FOUND':
+      return Object.assign({}, state, {
+        status: action.status
+      });
+    case 'WORKFLOW_GET_PRODUCTION_OFFICE':
```
Reettaphant: Does the production office need to be saved in the redux store? Could we just save it in the state of the workflow component and have the logic for computing it in a utility function?

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -0,0 +1,62 @@
...
+      url: `${WorkflowApi.workflowUrl}/api/sections`,
+      crossOrigin: true,
+      withCredentials: true
+    }).then(response => {
+      return response.data
+        .map(section => Object.assign({}, section, { title: section.name }))
+        .sort((a, b) => {
+          if (a.title.toLowerCase() < b.title.toLowerCase()) return -1;
+          if (a.title.toLowerCase() > b.title.toLowerCase()) return 1;
+          return 0;
```
Reettaphant: Could we use `sortBy` from lodash instead here?

public/video-ui/src/reducers/workflowReducer.js
akash1810: good point, doesn't really make sense to have it in the store

public/video-ui/src/services/WorkflowApi.js
Reettaphant: I guess it's just a bit neater but doesn't really make much of a difference.

akash1810: The [lodash syntax](https://lodash.com/docs/4.17.4#sortBy) looks largely similar to this, I'm tempted to stay with the native `.sort` function. What do you think?

PR #573, opened 2017-08-21
set poster image for self hosted video preview

akash1810:
Makes the video preview more consistent to how it'll render on site.

# Before
image

# After
image


akash1810:
> I'm mainly just approving whatever on earth the test video is

🐑  🐑  🐑  🐑  🐑  🐑  🐑  🐑  🐑  🐑

PR #574, opened 2017-08-22
re-id the blockAds DOM element

akash1810:
to prevent ad blockers from removing it from the DOM


akash1810:
After sleeping on it, I think the better solution is to stop using adblockers (or at least whitelist the Tools).

Merging this PR enforces the idea that adblockers are good. Last time I looked, there was a big money shaped hole downstairs...

I'm almost tempted to add `advert` to the class list of `body` to make the entire app unusable with adblockers installed, but that's probably too extreme 🤣

PR #576, opened 2017-08-25
remove legacy routes no longer in use

akash1810:
img


akash1810:
> Is this the old angular app?

Nah, its the old react app when it was `media-atom-maker.foo.bar.co.uk`

akash1810:
although, it is part of the route53 healthcheck, so we'll need to update that too...

PR #578, opened 2017-08-25
Spell YouTube correctly

paperboyo:
Too many capitalisations all over the app, but it may be subjective... Spelling “YouTube” isn’t.

EDIT: BTW, isn’t this used as a poster frame for when the atom is embedded on our frontends? “YouTube Poster Image” would be a bit misleading/irrelevant, then.

### DIFF
--- a/public/video-ui/src/components/VideoPreview/VideoPreview.js
+++ b/public/video-ui/src/components/VideoPreview/VideoPreview.js
@@ -53,7 +53,7 @@ export default class VideoPreview extends React.Component {
         <div className="video__detailbox">
           <div className="video__detailbox__header__container">
             <header className="video__detailbox__header">
-              Youtube Poster Image
+              YouTube Poster Image
             </header>
             <GridImageSelect
               updateVideo={this.saveAndUpdateVideoPoster}
@@ -68,7 +68,7 @@ export default class VideoPreview extends React.Component {
             formName={formNames.posterImage}
             updateErrors={this.props.updateErrors}
             fieldLocation="posterImage"
-            name="Youtube Poster Image"
+            name="YouTube Poster Image"
           />
         </div>
         <div className="video__detailbox">


akash1810:
> isn’t this used as a poster frame for when the atom is embedded on our frontends? “YouTube Poster Image” would be a bit misleading/irrelevant, then.

👍 maybe we should use a similar pattern to the headline and description fields and rename it `Main Image (YouTube poster)`?

image

paperboyo:
So: Composer Trail Image shows up _only_ on cards (not carousel) on fronts for when the:
– Video page is created and
– added to the front
In all other circumstances (incl. all external uses), YouTube Poster Image is used.

If the above is right, YouTube Poster Image is massively more important, sometimes (when there is no Video page) accounting for all uses.

For old Videos, it’s called “Main media (Holding image)” in Composer. Calling it Main Media would be misleading, cause that refers mainly to the (optional now) Video page. “Poster” wins in popularity with “Holding”.

So I like your “Main Image (YouTube poster)” best. If anything, it’s a bit verbose. Maybe we could use 🛈 with a tooltip explaining some of that?

paperboyo:
BTW, I let you change it, because I have no idea if I should escape the parentheses. Closing this...

akash1810:
> So: Composer Trail Image shows up only on cards (not carousel) on fronts for when the:
– Video page is created and
– added to the front
In all other circumstances (incl. all external uses), YouTube Poster Image is used.

Correct. Perhaps its easier to talk in ratios? YouTube poster is 16x9 and Composer Trail is 5x3*. A video atom can only be on fronts if a video page is created for it, and so the usual rules apply:
- 5x3 is used in cards
- 16x9 is used in carousel

*these ratios are actually enforced now too as we open Grid with them pre-selected and other options disabled

> BTW, I let you change it, because I have no idea if I should escape the parentheses. Closing this...

😭 your change was perfect! It had been approved too!

paperboyo:
Sorry, had bad experience with not escaping and here it affects a field called `name`, so wasn’t sure... See https://github.com/guardian/media-atom-maker/pull/579.

> we open Grid with them pre-selected and other options disabled

Tested this – is nice! I think it would speed up working on new atoms if the corresponding image could be automatically cropped (this would also crop them [in alignment](https://trello.com/c/O9rKRLU7/23-figure-out-what-to-do-about-trail-image), as now the second crop has to be performed manually and does not remember the placement of the first one – although it shouldn’t cause any twitching like in the old days...). Of course, Trail is just needed for Composer page, so I’m not sure if carrying it over for all atoms is not a waste, but most videos will probably have the Composer page created at some point. Not sure what Publish means in Training Mode on CODE, so haven’t tested the whole process...

paperboyo:
Found one that has no Trail crop. But it looks like there is no warning and creation of a Composer page is allowed, so Trail is not treated as compulsory (it is in Composer)...
Atom (PROD): 892b74da-bd0c-49a1-a9be-6656be1eafb3.

akash1810:
> Found one that has no Trail crop. But it looks like there is no warning and creation of a Composer page is allowed, so Trail is not treated as compulsory (it is in Composer)...

I think the Composer creation validation is mistaken here then... Will take a look next week.

paperboyo:
> Will take a look next week.

💯 Thank you!

Also, I was wrong. Atoms are also susceptible to twitching (16:9 and 5:3 crops not being [aligned by default](https://trello.com/c/O9rKRLU7/23-figure-out-what-to-do-about-trail-image) or on entering the second image cropping stage):

atom twitch
[atom: cba921cd-0d30-4a72-b410-161bdb3007750), image: e8f2c6e91596b26a73f176632d87ac48da6042a4); another example: atom: f0a31bdc-10a8-4fb9-a489-28488b7399e0; image: ffc9d75c7d28e23b9b51b9a810f937846d31de39]

Automagically creating a second crop correctly aligned may or not be a waste (depending on how many atoms end up _not_ having a Composer page). But entering a cropping stage for a second image with a crop frame correctly aligned would be a good idea irregardless, IMHO.

As it stands, it’s impossible to correctly align the pics, unless one knows and remembers about it every time and carefully observes cropping coordinates and dimensions in Grid. And uses a calculator 😆

PR #587, opened 2017-09-01
turn on pluto sync

akash1810:
TODO
- [x] send atom title in the message to Pluto
- [ ] have this flag as a server side config rather than a client-side POST body field
- [x] update IAM Role, granting Pluto permission to create temporary download links for simpler ingestion by videspine

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -80,7 +80,8 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
       assetVersion = -1,
       atomId = atom.id,
       title = atom.title,
-      user = user.email
+      user = user.email,
+      posterImageUrl = atom.posterImage.flatMap(_.master).map(_.file)
     )
 
     val metadata = UploadMetadata(
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -78,7 +78,9 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
       projectId = atom.plutoData.flatMap(_.projectId),
       s3Key = CompleteUploadKey(awsConfig.userUploadFolder, id).toString,
       assetVersion = -1,
-      atomId = atom.id
+      atomId = atom.id,
+      title = atom.title,
+      user = user.email
     )
 
     val metadata = UploadMetadata(
--- a/cloudformation/media-atom-maker.yaml
+++ b/cloudformation/media-atom-maker.yaml
@@ -199,7 +199,7 @@ Resources:
             Action:
             - kinesis:Get*
             - kinesis:DescribeStream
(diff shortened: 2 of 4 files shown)


akash1810:
cc @fredex42

app/controllers/UploadController.scala
```diff
@@ -78,7 +78,10 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
       projectId = atom.plutoData.flatMap(_.projectId),
       s3Key = CompleteUploadKey(awsConfig.userUploadFolder, id).toString,
       assetVersion = -1,
```
mbarton: I don't think we use this now?

fredex42: assetVersion, you mean?

mbarton: Yeah, I'm working on a different feature branch and I could see it being assigned but never read

app/controllers/UploadController.scala
```diff
@@ -78,7 +78,10 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
-      atomId = atom.id
+      atomId = atom.id,
+      title = atom.title,
+      user = user.email,
+      posterImageUrl = atom.posterImage.flatMap(_.master).map(_.file)
```
akash1810: Why these new properties? As far as I can tell, we're not using them?

cloudformation/media-atom-maker.yaml
```diff
@@ -199,6 +199,17 @@ Resources:
             Action:
             - kinesis:Get*
             - kinesis:DescribeStream
+      - PolicyName: PlutoIntegrationS3Policy
```
akash1810: can this be `PlutoIntegrationS3GetPolicy` to be more explicit?

app/controllers/UploadController.scala
fredex42: i am using them at the pluto end, to make it easier to find masters if one does not make it through or the "launch detector" phase does not work.  They are what we are likely to want to search on.

cloudformation/media-atom-maker.yaml
fredex42: sorry I don't understand?

fredex42: on re-reading I think i get it now :)

app/controllers/UploadController.scala
fredex42: it's just there 'cos it was - never touched it. Can remove if you like or let it get removed in another update, up to you

mbarton: I'll remove it in my branch. No worries

akash1810:
👍

fredex42:
cool. I'll merge this and deploy tomorrow morning then?

akash1810:
feel free to merge now, it'll go to CODE automatically

PR #588, opened 2017-09-04
fix usage bug on create form

akash1810:
Usage aren't getting updated when you create a new atom. This is ok when there is nothing in the store (blankUsageData), however if you've looked at an atom, then go to create a new atom, the usages do not change, that is, the usages of the newly created atom are the same as the atom you were previously looking at.

This fixes this by refreshing usages when creating a new atom.

relates to https://github.com/guardian/media-atom-maker/pull/583 and https://github.com/guardian/media-atom-maker/pull/584

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -28,6 +28,7 @@ class VideoDisplay extends React.Component {
   saveAndUpdateVideo = video => {
     if (this.props.route.mode === 'create') {
       this.props.videoActions.createVideo(video);
+      this.props.videoActions.getUsages(this.props.video.id);
     } else {
       this.props.videoActions.saveVideo(video);
     }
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -28,6 +28,7 @@ class VideoDisplay extends React.Component {
   saveAndUpdateVideo = video => {
     if (this.props.route.mode === 'create') {
       this.props.videoActions.createVideo(video);
+      this.props.videoActions.getUsages(this.props.video.id);
```
Reettaphant: Can we do something to ensure that the video is created before usages are fetched? If the video takes a long time to create is it still possible here to fetch the usages before we have an id?

akash1810: Good call! Have updated to do this once `createVideo` promise has finished

PR #590, opened 2017-09-05
refactor image components

akash1810:
Refactors the image components to be simpler and smaller:
- removes the forms as we're only really displaying an image now in the components
- pushes complex logic up
  - the Grid iframe URL is now constructed within `VideoImages`, this makes `GridImageSelect` simpler because there's no need to pass in `posterImage` for the `trailImage` variant
- add `propTypes` for sanity
- adds an empty state to `blankVideoData` for images to bring this object closer to an API response:
```js
const blankVideoData = {
  ...,
  posterImage: {
    assets: []
  },
  trailImage: {
    assets: []
  }
}
```
unfortunately, it does mean we have to do this https://github.com/guardian/media-atom-maker/pull/590/files#diff-7de3ab410c6924a5fb3d477157e30facR10

TODO
- [x] test on CODE


akash1810:
tested on CODE now 🎉

PR #591, opened 2017-09-05
more detailed composer update error message

akash1810:
also allow error messages as html

image


akash1810:
this is similar to what we do in Composer - https://github.com/guardian/flexible-content/blob/master/composer/src/js/controllers/content/common/status.js#L121-L123

philmcmahon:
👍

PR #593, opened 2017-09-06
rename 'hosted' to 'hosted by'

akash1810:
this is just for the display value, we're still saving 'hosted' per the enum

this is towards fully supported Labs with their hosted and paid for categories

image


akash1810:
[support for paid for and hosted](https://trello.com/c/swV75Es1/86-support-for-paid-for-and-hosted)

PR #600, opened 2017-09-14
i hate dates

akash1810:
fix the office location sniffing


akash1810:
this doesn't work in Opera Developer for some bizarre reason, works fine in Chrome though so 🎉

PR #603, opened 2017-09-15
merge two columns in video data form

akash1810:
saving space and improving layout

Before
image

After
image


akash1810:
the diff is incredible!!

PR #604, opened 2017-09-15
extract common sixteen by nine styling

akash1810:
apply style to video preview and search image

fixes the empty state of video preview

before
image

after
image

### DIFF
--- a/public/video-ui/styles/main.scss
+++ b/public/video-ui/styles/main.scss
@@ -19,6 +19,7 @@
 
 // 4. Layout-related sections
 @import
+  'layout/_common.scss',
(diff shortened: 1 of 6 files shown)


public/video-ui/styles/main.scss
```diff
@@ -19,6 +19,7 @@
 
 // 4. Layout-related sections
 @import
+  'layout/_common.scss',
```
mbarton: Is this different to having the name missing the underscore (as below)?

akash1810: yes, in that this is wrong 😛

PR #607, opened 2017-09-18
allow for scoped channels

akash1810:
Editorial have experienced problems where they've pasted a link to a video on the main channel, however as its not in the allowedChannel list, updates don't go to YouTube. This results in them editing the Atom and then going to youtube.com to do the same thing. This is time-consuming and not scalable because a lot of the subs don't have yt credentials.

We removed the main channel to prevent too many public uploads.

This PR updates the `/api/youtube/channels` endpoint, returning a list of privacyStates that a channel can be uploaded to. That is, we can add the main channel to the config `youtube.channels.unlisted` and allow Editorial to treat this Atom like any other, safe in the knowledge that its always unlisted.

channel

TODO
- [x] test on CODE
- [x] confirm behaviour with Adam and Katie
- [ ] update configs in S3 before deploying

### DIFF
--- a/app/controllers/Youtube.scala
+++ b/app/controllers/Youtube.scala
@@ -17,15 +17,9 @@ class Youtube (val authActions: HMACAuthActions, youtube: YouTube) extends Contr
     val isTrainingMode = isInTrainingMode(req)
 
     val channels = if (isTrainingMode) {
-      youtube.trainingChannels match {
-        case Nil => List()
-        case trainingList => youtube.channels.filter(c => trainingList.contains(c.id))
-      }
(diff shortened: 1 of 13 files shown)


akash1810:
Before:

```json
[
  {
    "id": "UC-pCzg7zq1Nahs0lENOqyiw",
    "title": "gdn-dev-acc"
  }
]
```

After:
```json
[
  {
    "id": "UC-pCzg7zq1Nahs0lENOqyiw",
    "title": "gdn-dev-acc",
    "privacyStates": [
      "Unlisted",
      "Private"
    ]
  }
]
```

PR #611, opened 2017-09-20
presence indicators

akash1810:
The positioning isn't consistent across all pages, but I think there's benefit in shipping this now.

Also refactors out the publish state element into its own component to position the indicators in the right place on the video page.

# Search page
Nothing, obviously.
image

# Video page
image

# Upload page
image

# Audit page
image

# Help page
image

### DIFF
--- a/app/controllers/VideoUIApp.scala
+++ b/app/controllers/VideoUIApp.scala
@@ -77,16 +77,12 @@ class VideoUIApp(val authActions: HMACAuthActions, conf: Configuration, awsConfi
   }
 
   private def presenceConfig(user: User): Option[Presence] = {
-    (awsConfig.stage, conf.getString("presence.domain")) match {
-      case ("DEV", _ ) =>
-        None
-
-      case (_, None) =>
+    conf.getString("presence.domain") match {
+      case Some(origin) =>
+        Some(Presence(origin, user.firstName, user.lastName, user.email))
+      case None =>
         log.warn("Presence disabled. Missing presence.domain in config")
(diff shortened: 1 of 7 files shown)


app/controllers/VideoUIApp.scala
```diff
@@ -77,16 +77,12 @@ class VideoUIApp(val authActions: HMACAuthActions, conf: Configuration, awsConfi
   }
 
   private def presenceConfig(user: User): Option[Presence] = {
-    (awsConfig.stage, conf.getString("presence.domain")) match {
-      case ("DEV", _ ) =>
```
akash1810: allow presence in def if config exists

public/video-ui/styles/components/_presence.scss
```diff
@@ -0,0 +1,18 @@
+.presence-list {
```
akash1810: stolen from atom workshop...

public/video-ui/src/components/Presence.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 
 export class Presence extends React.Component {
-  state = null;
+  state = {
+    client: null,
+    visitors: []
```
akash1810: this felt like the simplest thing to do, rather than throw it to the store... because I'm lazy 😄

shaundillon: It's easy enough to push up if it's needed elsewhere. Local state is for this kind of thing anyway

PR #614, opened 2017-09-22
always make inactive videos private on youtube

akash1810:
as requested from Editorial for legal reasons


mbarton:
I'm not against this but we should bear in mind the propagation delay for updates changing an asset.

It's unlikely but I'm thinking of a video atom that is on a front and we switch assets for benign reasons and people see a cached but now private YouTube video

akash1810:
Agreed. The delay shouldn't be any longer than for articles however, as it's as fast as CAPI processes the message.

mbarton:
Even for front presses? Although I guess videos are likely to be embedded in fronts that press with high priority

Reettaphant:
I can't remember how often fronts are pressed but it depends on the the kind of fronts. Network fronts get pressed fast but on other fronts you'd be able to see it.

akash1810:
Let's confirm with Editorial what their preferred behaviour is; I can chase today.

PR #615, opened 2017-09-24
remove angular app

akash1810:
- remove old angular app
- remove gulp too
- remove old api endpoints
- separate out login and healthcheck to their own controllers

TODO:
- [x] confirm with Labs this is OK to do


akash1810:
after a week of GA tracking in place, the only visitor to the angular app was me 😂  should be safe to remove this then!

philmcmahon:
Woohoo! 🔥

akash1810:
this branch is breaking the integration tests for some reason, will fix them before merging

PR #618, opened 2017-10-02
Better upload metadata

mbarton:
Adds the original filename, time and user to each upload in the video assets page.

The information was already available in the metadata we pass through the step functions. The trick to getting it into the UI was to change the ID of each step function execution to include the asset version number rather than a UUID.

This means that if an upload fails, that version number is skipped. I don't think it would be a problem but should be discussed in this review.

I've also made some changes to how the YouTube upload is rendered client-side. By removing the `processing` data for completed videos we should see less of a spinner saying "Uploading" and more of "Uploading to YouTube" with an actual progress bar.

These changes mean that we now call the step functions API for every upload, not just those in progress. Requests on a test atom with ~4 videos were taking around 1 second so caching is most likely a benefit, especially since a completed execution can be cached indefinitely.


paperboyo:
Cool! +1 for [filenames](https://github.com/guardian/grid/pull/937#issuecomment-121681833)!

Some emails can be much, much longer. Grid uses this pseudo-presence pattern to avoid visual noise (visible only on hover as it’s only for the interested):

pseudo-presence

Most obviously, future rich searching aids should allow for using this (and all other metadata) while performing searches.

common/src/main/scala/com/gu/media/youtube/YouTubeProcessingStatus.scala
```diff
@@ -9,25 +9,26 @@ case class YouTubeProcessingStatus(id: String, status: String, total: Long, proc
...
-  def apply(video: Video): YouTubeProcessingStatus = {
+  def apply(video: Video): Option[YouTubeProcessingStatus] = {
     val base = YouTubeProcessingStatus(video.getId, status = "", total = 0, processed = 0, timeLeftMs = 0, failure = None)
 
     val summary = video.getStatus
 
     summary.getUploadStatus match {
-      case "uploaded" => parseProcessingStatus(base, video.getProcessingDetails)
-      case "failed" => base.copy(status = "failed", failure = Option(summary.getFailureReason).map(humanizeFailureReason))
+      case "uploaded" =>
```
akash1810: minor: would be good to extract these as case objects, rather than using strings

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -65,25 +71,51 @@ export function Asset({ content, title, href, active, activateFn }) {
...
+  if (processing && !processing.failed) {
+    return processing.status;
+  } else if (metadata) {
+    const { startTimestamp, user, originalFilename } = metadata;
+    const startDate = moment(startTimestamp);
+
+    return (
+      <div>
+        <strong>
+          {startDate.format('DD/MM/YY HH:mm:ss')}
```
akash1810: can we use `yyyy-mm-dd` for easier internationalisation?

public/video-ui/src/services/pandaReqwest.js
```diff
@@ -8,24 +8,24 @@ function poll(reqwestBody, timeout) {
...
-            setTimeout(makeRequest, interval, resolve, reject);
-          },
-          error => {
-            throw error;
-          }
-        );
-      } else {
-        if (Number(new Date()) < endTime) {
-          setTimeout(makeRequest, interval, resolve, reject);
+      if (Number(new Date()) < endTime) {
```
akash1810: any reason for this change? linter?

mbarton: Ah should've mentioned it in the PR comment. It stops retrying forever if the response is a 419 (like not being logged into Workflow)

app/controllers/UploadController.scala
```diff
@@ -29,28 +25,27 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
...
-    parse(raw) { req: CreateRequest =>
+    parse(raw) { req: UploadRequest =>
       if(req.selfHost && !raw.permissions.addSelfHostedAsset) {
         Unauthorized(s"User ${raw.user.email} is not authorised with permissions to upload self-hosted asset")
       } else {
         log.info(s"Request for upload under atom ${req.atomId}. filename=${req.filename}. size=${req.size}, selfHosted=${req.selfHost}")
 
         val atom = MediaAtom.fromThrift(getPreviewAtom(req.atomId))
-        val upload = buildUpload(atom, raw.user, req.size, req.selfHost, req.syncWithPluto)
+        val version = if(atom.assets.isEmpty) { 1 } else { atom.assets.map(_.version).max + 1 }
```
akash1810: minor: would be nice if there was a `getNextVersion` helper function on `MediaAtom` for this

app/model/ClientAsset.scala
```diff
@@ -118,3 +119,83 @@ object ClientAssetProcessing {
...
   }
 }
+
+object ClientAssetMetadata {
+  implicit val format: Format[ClientAssetMetadata] = Jsonx.formatCaseClass[ClientAssetMetadata]
+}
+
+object ClientAssetTestData {
+  // SO MANY PERMUTATIONS! This helps test the UI without having to upload videos every time
+  def youTubeAsset(version: Long): ClientAsset = ClientAsset(
```
akash1810: feels odd to check this in...

public/video-ui/src/components/VideoUpload/VideoAsset.js
```diff
@@ -1,92 +1,148 @@
...
 import { VideoEmbed } from '../utils/VideoEmbed';
 
-const UPLOAD_FAILED_MSG = (
-  <p>
-    <strong>Upload Failed</strong><br />
-    You may retry your upload.<br />
-    This message will disappear after 10 minutes.
-  </p>
-);
+function presenceInitials(email) {
```
akash1810: minor: `getPresenceInitials` to be more expressive

test/model/ClientAssetTest.scala
```diff
@@ -13,130 +11,31 @@ class ClientAssetTest extends FunSuite with MustMatchers {
...
-      val processing = ClientAssetProcessing("YouTube Processing", failed = false, None, None)
-
-      val expected = ClientAsset(ytAsset.version.toString, asset = None, Some(processing))
-      val actual = ClientAsset(List(ytAsset), youTube)
-
-      actual must contain(expected)
-    }
-  }
-
   test("Group assets by version (newest first)") {
```
akash1810: would `latest first` be more descriptive?

uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
```diff
@@ -26,16 +27,28 @@ class AddAssetToAtom extends LambdaWithParams[Upload, Upload] with DynamoAccess
...
 
     saveAtom(after)
     audit.auditUpdate(atomId, "media-atom-pipeline", s"Added YouTube video $asset")
 
     upload
   }
 
+  private def getVersion(atom: MediaAtom, upload: Upload): Long = {
+    upload.metadata.version.getOrElse {
+      if (atom.assets.isEmpty) {
```
akash1810: see above re helper function

app/model/ClientAsset.scala
mbarton: It's pretty useful to see the different types of result though (eg a find usages) as well as being useful for testing the UI.

I'd like to keep it if that's ok?

common/src/main/scala/com/gu/media/youtube/YouTubeProcessingStatus.scala
mbarton: I'd need to have manual JSON Writes then though

mbarton:
@akash1810 I've updated with review fixes

PR #620, opened 2017-10-04
remove nbsp too

akash1810:
somehow some videos have &nbsp; in their title which was failing the regex


akash1810:
thanks GH...

gh

akash1810:
> somehow some videos have   in their title which was failing the regex

I've worked it out! `alt + -` gives you the elongated hyphen (–) which is part of the Guardian style guide. `alt + space` gives you `nbsp;`, so I guess `alt` is held down for too long...

PR #621, opened 2017-10-05
Temporarily disable Pluto sync

mbarton:
There are a number of issues with the Pluto sync flow as it stands:

- https://video.code.dev-gutools.co.uk/videos/pluto-list is hardcoded to `[1,2,3,4,5,6,7]`
- The project selector on the upload page shows all projects (gonna get massive)
- US and AU don't even use Pluto but still get emails about it

I'm disabling it until we review how to move forward with this feature.


akash1810:
could we turn of the email rather than pluto sync, just so @fredex42 's work on the other side isn't halted?

commenting out https://github.com/guardian/media-atom-maker/blob/5ca359b0e1f2bbaf5ca5eeffdbfa6add91d01087/common/src/main/scala/com/gu/media/upload/PlutoUploadActions.scala#L23-L27 should work?

mbarton:
@akash1810 agreed. I've updated the PR

PR #624, opened 2017-10-10
allow hosted to add paid-content tags

akash1810:
if you're a hosted atom, you should be allowed to add a paid-content tag, because, well, that's what they're for 😁

the implementation feels a bit hacky...

TODO
- [x] test on CODE

### DIFF
--- a/public/video-ui/src/constants/TagTypes.js
+++ b/public/video-ui/src/constants/TagTypes.js
@@ -23,7 +23,7 @@ export default class TagTypes {
     return 'tone';
   }
 
-  static get paidContent() {
+  static get paidContentKeywords() {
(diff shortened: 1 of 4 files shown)


akash1810:
this'll make Tim Hughes really happy 😄

public/video-ui/src/constants/TagTypes.js
```diff
@@ -22,4 +22,8 @@ export default class TagTypes {
   static get tone() {
     return 'tone';
   }
+
+  static get paidContent() {
```
Reettaphant: This is very minor so feel free to ignore but `paidContentKeywords` might be a better name - this type includes more than just paid content tags.

PR #628, opened 2017-10-17
filter channel list based on category

akash1810:
Now that we have `isCommercial` in the channel api, the `Hosted` and `Paid` categories should only be able to upload to commercial channels.

Also:
- refactors useful functions to `VideoUtils`
- adds `GLabs - Paid for` category to UI (the thrift model already has this as a value in the category enum)

labs

### DIFF
--- a/public/video-ui/src/components/VideoData/VideoData.js
+++ b/public/video-ui/src/components/VideoData/VideoData.js
@@ -10,8 +10,7 @@ import TagTypes from '../../constants/TagTypes';
 import { fieldLengths } from '../../constants/videoEditValidation';
 import { videoCategories } from '../../constants/videoCategories';
 import PrivacyStates from '../../constants/privacyStates';
-import { channelAllowed } from '../../util/channelAllowed';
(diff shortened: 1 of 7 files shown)


public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -31,20 +30,13 @@ class VideoData extends React.Component {
...
-    const privacyStates = this.props.video.channelId && this.props.youtube.channels.length > 0 && canUpdateYouTube
-      ? this.props.youtube.channels.find(_ => _.id === this.props.video.channelId).privacyStates
-      : PrivacyStates.defaultStates;
-
-    const minDurationForAds = getStore().getState().config.minDurationForAds;
-
-    const tooShortForAds =
-      this.props.video.duration > 0 &&
-      this.props.video.duration < minDurationForAds;
+    const isYoutubeAtom = VideoUtils.isYoutube(this.props.video);
```
shaundillon: Much cleaner!

PR #629, opened 2017-10-17
Open composer page in new tab

7091lapS:
When composer page exists, show "Open in Composer" button in place of "Create Video Page" button

screen shot 2017-10-17 at 10 28 29

### DIFF
--- a/public/video-ui/src/components/Videos/ComposerPageCreate.js
+++ b/public/video-ui/src/components/Videos/ComposerPageCreate.js
@@ -14,8 +14,9 @@ export default class ComposerPageCreate extends React.Component {
 
   getComposerId = () => {
     const usages = getStore().getState().usage.data;
-    if (usages.preview.video.length && !usages.published.video.length) {
-      return usages.preview.video[0].fields.internalComposerCode;
+    const videoPages = [...usages.preview.video, ...usages.published.video];
+    if (videoPages.length !== 0) {
+      return videoPages[0].fields.internalComposerCode;
     }
     else {
       console.log("Could not find composer id");
--- a/public/video-ui/src/components/Videos/ComposerPageCreate.js
+++ b/public/video-ui/src/components/Videos/ComposerPageCreate.js
@@ -12,9 +12,13 @@ export default class ComposerPageCreate extends React.Component {
     return getStore().getState().config.composerUrl;
   };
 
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/Videos/ComposerPageCreate.js
```diff
@@ -12,10 +12,14 @@ export default class ComposerPageCreate extends React.Component {
     return getStore().getState().config.composerUrl;
   };
 
+  getComposerId = () => getStore().getState().usage.data.preview.video[0].fields.internalComposerCode;
```
akash1810: this'll fail if there's a published usage, could we check if `preview.video.length === 0` and then check `published.video.length`?

public/video-ui/src/components/Videos/ComposerPageCreate.js
```diff
@@ -12,10 +12,14 @@ export default class ComposerPageCreate extends React.Component {
...
+
   isHosted = () => {
     return this.props.video.category === 'Hosted';
   };
 
+  openComposerPage = () => window.open(`${this.getComposerUrl()}/content/${this.getComposerId()}`, '_blank');
```
akash1810: could we render an `href` in this scenario?

```js
render() {
  if (showOpenPage) {
    return (<a href="" target="_blank">yo</a>)
  }
  else {
    <button ...></button>
  }
}

```

public/video-ui/src/components/Videos/ComposerPageCreate.js
```diff
@@ -12,10 +12,22 @@ export default class ComposerPageCreate extends React.Component {
     return getStore().getState().config.composerUrl;
   };
 
+  getComposerId = () => {
+    const usages = getStore().getState().usage.data;
+    if (usages.preview.video.length && !usages.published.video.length) {
```
akash1810: IIUC if the Composer page has been published, we will fail?

Maybe we update the [usage response](https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/services/VideosApi.js#L134-L145) or check `if (totalVideoPages > 0)`?

akash1810: even better, we could do something like:

```js
const videoPages = [...usages.preview.video, ...usages.published.video]
if (videoPages.length !== 0) {
  return videoPages.[0].fields.internalComposerCode
}
```

PR #631, opened 2017-10-17
refactor tag display name function

akash1810:
- reduces the number of branches
- always append sectionName to tag regardless how may spaces there are

should help avoid [this scenario](https://www.theguardian.com/books/video/2017/oct/17/its-just-marmite-taste-testing-the-new-vegemite-blend-17-video) in the future where a video has been launched with the `Books` `food and drink` tag and thus has `books` in its url

image


akash1810:
> unnecessarily long tag names (e.g. http://internal.content.code.dev-guardianapis.com/social-care-network/international) 

oh wow!

PR #632, opened 2017-10-17
stricter comparison

akash1810:


akash1810:
don't really need a review, right? merging...

PR #635, opened 2017-10-19
Youtube characters limit

Reettaphant:
Youtube has a character limit on it's tag list. Mam should make sure that youtube keywords don't go over this character limit so that we don't end up in a situation where updates to youtube are failing because we are exceeding this limit
screen shot 2017-10-19 at 08 59 43
.

### DIFF
--- a/public/video-ui/src/components/FormFields/TagPicker.js
+++ b/public/video-ui/src/components/FormFields/TagPicker.js
@@ -349,6 +349,7 @@ export default class TagPicker extends React.Component {
   render() {
 
     const hasWarning = this.props.hasWarning(this.props) && this.state.capiTags.length === 0;
+    const hasError = this.props.hasError(this.props);
 
     if (!this.props.editable) {
       if (!this.state.tagValue || this.state.tagValue.length === 0) {
@@ -389,6 +390,11 @@ export default class TagPicker extends React.Component {
         {this.renderAddedTags()}
         {hasWarning
           ? <p className="form__message form__message--warning">
+          {this.props.notification.message}
+          </p>
+            : ''}
+        {hasError
+          ? <p className="form__message form__message--error">
           {this.props.notification.message}
           </p>
             : ''}
--- a/public/video-ui/src/components/FormFields/TagPicker.js
+++ b/public/video-ui/src/components/FormFields/TagPicker.js
@@ -14,6 +14,8 @@ import removeTagDuplicates from '../../util/removeTagDuplicates';
 import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
 import {requiredForComposerWarning} from '../../constants/requiredForComposerWarning';
 import ReactTooltip from 'react-tooltip';
+import { getYouTubeTagCharCount } from '../../util/getYouTubeTagCharCount';
+import YouTubeKeywords from '../../constants/youTubeKeywords';
 
 export default class TagPicker extends React.Component {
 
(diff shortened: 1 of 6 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -110,6 +112,22 @@ class VideoDisplay extends React.Component {
...
 
+  validateYouTubeKeywords = youTubeKeywords => {
+    const charLimit = YouTubeKeywords.maxCharacters;
+    const numberOfChars = getYouTubeTagCharCount(youTubeKeywords);
+
+    if (numberOfChars > charLimit) {
+
+      return new FieldNotification(
+        'required',
+        `Maximum characters allowed in youTube tags is ${charLimit}.`,
```
gtrufitt: Minor but `youTube` should be `YouTube`

public/video-ui/src/util/getYouTubeTagCharCount.js
```diff
@@ -0,0 +1,8 @@
+export function getYouTubeTagCharCount(tags) {
+  if (Array.isArray(tags)) {
+    return tags.reduce((charCount, keyword) => {
```
gtrufitt: Could be:
```js
return tags.reduce((charCount, keyword) => charCount += keyword.length, 0);
```

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -330,6 +331,17 @@ export default class TagPicker extends React.Component {
     }
   }
 
+  renderCharCount() {
+    if (this.props.tagType === TagTypes.youtube) {
+      const charCount = getYouTubeTagCharCount(this.props.fieldValue);
+      return (
+        <span>
+          Character count: {charCount}
```
gtrufitt: Could you put `{charCount} / {YouTubeKeywords.maxCharacters}` so it's clearer what the limit is as the user is typing?

public/video-ui/src/components/FormFields/TagPicker.js
```diff
@@ -392,6 +406,11 @@ export default class TagPicker extends React.Component {
           {this.props.notification.message}
           </p>
             : ''}
+        {hasError
```
gtrufitt: I wonder if using the logical `&&` here is cleaner than the ternary: 

```js
hasError && <p className="form__message form__message--error">
           {this.props.notification.message}
           </p>
```

Though it looks as if there's a consistency of using ternary operators with empty string in else, so not major

RichieAHB:
Does the YT limit include separators too between keywords (i.e. the comma). If not then this is fine but if it does then you might want:

`return tags.reduce((charCount, keyword) => charCount += keyword.length, 0) + tags.length - 1`

Reettaphant:
@RichieAHB Yes, separators are included in the character count.

public/video-ui/src/components/Video/Display.js
Reettaphant: 👍

public/video-ui/src/util/getYouTubeTagCharCount.js
Reettaphant: 👍

public/video-ui/src/components/FormFields/TagPicker.js
Reettaphant: I'll leave these alone for the reason you stated.

PR #637, opened 2017-10-23
Publish button

Reettaphant:
- Fix checks to see if video has unpublished changes 
- Adding poster and trail images to these checks is annoying. The function finds out what the other fields that need checking are by looking at validated form fields on a page so that we don't have to hard code new fields to a list every time we add them but the image fields are not part of a form so they need adding manually.

### DIFF
--- a/public/video-ui/src/util/hasUnpublishedChanges.js
+++ b/public/video-ui/src/util/hasUnpublishedChanges.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import { appUpdatedFields } from '../constants/appUpdatedFields';
+import { imageFields } from '../constants/imageFields';
 
 export function hasUnpublishedChanges(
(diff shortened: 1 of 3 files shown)


public/video-ui/src/util/hasUnpublishedChanges.js
```diff
@@ -14,7 +15,7 @@ export function hasUnpublishedChanges(
     return true;
   }
 
-  const allFields = editableFields.concat(appUpdatedFields);
+  const allFields = editableFields.concat(appUpdatedFields).concat(imageFields);
```
akash1810: minor: a [spread](https://davidwalsh.name/spread-operator) operator may read a bit nicer here?

```js
const allFields = [
  ...editableFields,
  ...appUpdatedFields,
  ...imageFields
];
```

PR #638, opened 2017-10-23
don't count spaces

Reettaphant:
Adam informs me that youtube doesn't count spaces in keywords as characters - this pr strips those from the count.


Reettaphant:
I'm now counting the quotation marks around the keywords - just confirming with Adam that this is what he wants.

Issue #391, opened 2017-04-19
Upgrade to coursier 1.0.0-RC1


mbarton:
cc @akash1810 @Reettaphant 

This was previously attempted in #386 but lead to build instability around test artifacts from the atom manager lib:

```
[Step 11/11] /opt/teamcity/buildAgent/work/92ffd722307ca558/test/AuthTests.scala:3: object test is not a member of package com.gu.atom.play
[14:14:47][Step 11/11] import com.gu.atom.play.test.AtomSuite
[14:14:47][Step 11/11]                         ^
[14:14:47][Step 11/11] /opt/teamcity/buildAgent/work/92ffd722307ca558/test/AuthTests.scala:12: not found: type AtomSuite
[14:14:47][Step 11/11] trait AuthTests extends AtomSuite {
[14:14:47][Step 11/11]                         ^
[14:14:47][Step 11/11] /opt/teamcity/buildAgent/work/92ffd722307ca558/test/AuthTests.scala:15: value customOverrides is not a member of AnyRef
[14:14:47][Step 11/11]     super.customOverrides :+ (bind(classOf[HMACAuthActions]).to(classOf[TestPandaAuth]))
[14:14:47][Step 11/11]           ^
[14:14:47][Step 11/11] /opt/teamcity/buildAgent/work/92ffd722307ca558/test/AuthTests.scala:19: not found: type AtomTestConf
[14:14:47][Step 11/11]   def getAuthActions(implicit conf: AtomTestConf): HMACAuthActions = conf.iget[HMACAuthActions]
[14:14:47][Step 11/11]                                     ^
[14:14:47][Step 11/11] /opt/teamcity/buildAgent/work/92ffd722307ca558/test/AuthTests.scala:29: not found: type AtomTestConf
[14:14:47][Step 11/11]   def requestWithCookies(implicit conf: AtomTestConf) =
[14:14:47][Step 11/11]                                         ^
[14:14:48][Step 11/11] 5 errors found
```

We don't know exactly what the issues are. I suspect it might be related to using a cache created with an older version of coursier with a newer version, or vice versa. Certainly the build errors only seemed to appear sometimes.

Potentially relevant:

https://github.com/coursier/coursier/issues/461
https://github.com/coursier/coursier/issues/444

PR #647, opened 2017-10-25
update coursier

akash1810:
should fix https://github.com/guardian/media-atom-maker/issues/391?


akash1810:
Left this on CODE overnight and the integration tests had no issue, so yes, fixes #391 🎉

PR #648, opened 2017-10-26
compare apples to apples

akash1810:
upload.id is a string...

we changed from `==` to `===` [recently](https://github.com/guardian/media-atom-maker/pull/640/files#diff-e7f07317cc9574be7a20ed5d64c64a9fR60)...


mbarton:
I think this will fail when you are uploading a video since the ID is a string then. Sorry for the confusing code!

akash1810:
tested in DEV and works, because `parseInt("f39727df-b210-45c6-900d-2b2aeae0b623-4") === NaN` oh javascript 🤕

PR #651, opened 2017-10-27
Remember upload metadata even if state machine changed

mbarton:
tl;dr - we can now change the upload pipeline without losing metadata (user and filename) about each upload.

The Cloud Formation support for Step Functions will delete the old state machine rather than modify it. We rely on this `Upload` data to show the user and filename for each upload in the assets page.

This PR adds a dynamo table to cache the `Upload` data. It is written to at the start and end of each run of the state machine. The upload controller will then look in the cache first **for metadata only**. This does not affect looking up the progress of an upload which is still read from the state machine.

To support this, I've added  `startTimestamp` to `Upload`. This was originally read directly from the state machine execution and is backfilled from there if not present.

TODO:
- [x] ~save executions from CODE~
- [x] deploy CFN to CODE
- [x] ~upload executions to new cache table in CODE~
- [x] test in CODE
- [x] save executions from PROD
- [ ] deploy pipeline CFN to PROD
- [ ] deploy app CFN to PROD
- [ ] merge and deploy to PROD
- [ ] upload executions to new cache table in PROD

### DIFF
--- a/app/util/UploadDecorator.scala
+++ b/app/util/UploadDecorator.scala
@@ -1,0 +1,39 @@
+package util
+
+import com.gu.media.aws.{DynamoAccess, UploadAccess}
+import com.gu.media.upload.model.Upload
+import com.gu.scanamo.{Scanamo, Table}
+import com.gu.scanamo.syntax._
+import model.{ClientAsset, ClientAssetMetadata}
+
+class UploadDecorator(aws: DynamoAccess with UploadAccess, stepFunctions: StepFunctions) {
+  private val table = Table[Upload](aws.cacheTableName)
+
+  def addMetadata(atomId: String, video: ClientAsset): ClientAsset = {
+    val id = s"$atomId-${video.id}"
+
+    getUpload(id) match {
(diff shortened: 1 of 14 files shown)


app/util/UploadDecorator.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def addMetadata(atomId: String, video: ClientAsset): ClientAsset = {
+    val id = s"$atomId-${video.id}"
+
+    getUpload(id) match {
+      case Some(upload) =>
+        video.copy(metadata = Some(
+          ClientAssetMetadata(
+            upload.metadata.originalFilename,
+            upload.metadata.startTimestamp.getOrElse(0),
```
akash1810: `0`? 😱

mbarton: Yeah this is laziness on my part. It happens to not render in the UI (probably due to an `if(...)` check somewhere) but I should actually make that field optional

mbarton:
@akash1810 I'll merge this Monday morning since it's slightly involved

PR #652, opened 2017-10-27
allow blocking ads on create form

akash1810:
Request from James C.

The current workflow is:
- Create atom
- Block ads field is disabled
- Upload video
- Block ads field becomes enabled

A faster process is:
- Block ads field is enabled in the creation form from the start

### DIFF
--- a/public/video-ui/src/util/video.js
+++ b/public/video-ui/src/util/video.js
@@ -67,12 +67,15 @@ export default class VideoUtils {
   }
 
   static isEligibleForAds(atom) {
-    const minDurationForAds = getStore().getState().config.minDurationForAds;
-
     if (VideoUtils.isCommercialType(atom)) {
       return false;
     }
 
+    if (!VideoUtils.hasAssets(atom)) {
+      return true;
+    }
+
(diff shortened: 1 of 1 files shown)


public/video-ui/src/util/video.js
```diff
@@ -67,12 +67,15 @@ export default class VideoUtils {
   }
 
   static isEligibleForAds(atom) {
-    const minDurationForAds = getStore().getState().config.minDurationForAds;
-
     if (VideoUtils.isCommercialType(atom)) {
       return false;
     }
 
+    if (!VideoUtils.hasAssets(atom)) {
```
akash1810: a newly created atom doesn't have assets

public/video-ui/src/util/video.js
```diff
@@ -67,12 +67,15 @@ export default class VideoUtils {
...
+      return true;
+    }
+
+    const minDurationForAds = getStore().getState().config.minDurationForAds;
```
akash1810: refactor: hit store at last possible moment

PR #653, opened 2017-10-30
Avoid long multipart copies

mbarton:
We combine all the chunks in S3 using multipart copy so that Pluto can pull it down in a single go.

For really big videos this takes longer than the maximum execution time for a lambda. This PR changes the state machine to do each copy (100MB chunk) in a single lambda invocation, just like we do with YouTube uploads.


mbarton:
[Create multipart upload in S3 in multiple lambda invocations](https://trello.com/c/lQNzNinR/139-create-multipart-upload-in-s3-in-multiple-lambda-invocations)

PR #654, opened 2017-10-31
send a payload to credentials endpoint

akash1810:
allows for easier extensibility for uploading other files (xml!) as we can pattern match on the request body rather than using query string values

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -58,14 +58,16 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
     }
   }
 
-  def credentials(id: String, key: String) = LookupPermissions { implicit req =>
-    getPart(id, key) match {
-      case Some(part) =>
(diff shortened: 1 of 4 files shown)


app/controllers/UploadController.scala
```diff
@@ -89,9 +91,9 @@ class UploadController(override val authActions: HMACAuthActions, awsConfig: AWS
       video
   }
 
-  private def getPart(id: String, key: String): Option[UploadPart] = for {
-    upload <- stepFunctions.getById(id)
-    part <- upload.parts.find(_.key == key)
+  private def getPart(credentialsRequest: UploadCredentialsRequest): Option[UploadPart] = for {
+    upload <- stepFunctions.getById(credentialsRequest.atomId)
```
akash1810: turns out this is the wrong ID 😢  step function invocations are identified by `s"$atomId-$version"`. Closing this PR for now, will reopen another once I've got it working.

PR #656, opened 2017-11-01
Automatic composer updates

Reettaphant:
- Push updates to canonical video page automatically when saving an atom if video page exists
- If canonical video page exists, required missing composer fields are displayed as errors and atom cannot be saved before they are added back in. If the page does not exist, they are rendered as warnings as before.

### DIFF
--- a/public/video-ui/src/components/FormFields/ScribeEditor.js
+++ b/public/video-ui/src/components/FormFields/ScribeEditor.js
@@ -6,7 +6,7 @@ import scribePluginToolbar from 'scribe-plugin-toolbar';
 import scribePluginLinkPromptCommand from 'scribe-plugin-link-prompt-command';
 import scribePluginSanitizer from 'scribe-plugin-sanitizer';
 import {keyCodes} from '../../constants/keyCodes';
-import {requiredForComposerWarning} from '../../constants/requiredForComposerWarning';
+import RequiredForComposer from '../../constants/requiredForComposer';
 import ReactTooltip from 'react-tooltip';
 
 export default class ScribeEditorField extends React.Component {
@@ -106,6 +106,7 @@ export default class ScribeEditorField extends React.Component {
 
   renderField() {
     const hasWarning = this.state.wordCount === 0  && this.props.fieldLocation === 'trailText';
+    const hasError = false;
 
     if (!this.props.editable) {
       if (this.state.wordCount === 0) {
@@ -139,7 +140,12 @@ export default class ScribeEditorField extends React.Component {
         </div>
         {hasWarning
           ? <p className="form__message form__message--warning">
-            {requiredForComposerWarning}
+            {RequiredForComposer.warning}
+            </p>
+          : ''}
+        {hasError
+          ? <p className="form__message form__message--warning">
+            {RequiredForComposer.error}
             </p>
           : ''}
       </div>
--- a/public/video-ui/src/components/FormFields/ScribeEditor.js
+++ b/public/video-ui/src/components/FormFields/ScribeEditor.js
@@ -105,8 +105,9 @@ export default class ScribeEditorField extends React.Component {
   };
 
   renderField() {
-    const hasWarning = this.state.wordCount === 0  && this.props.fieldLocation === 'trailText';
-    const hasError = false;
+    const requiresValidation = this.state.wordCount === 0  && this.props.fieldLocation === 'trailText';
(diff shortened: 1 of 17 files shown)


public/video-ui/src/components/FormFields/ScribeEditor.js
```diff
@@ -105,7 +105,8 @@ export default class ScribeEditorField extends React.Component {
   };
 
   renderField() {
-    const hasWarning = this.state.wordCount === 0  && this.props.fieldLocation === 'trailText';
+    const hasWarning = this.props.isDesired && (this.state.wordCount === 0  && this.props.fieldLocation === 'trailText');
+    const hasError = this.props.isRequired && (this.state.wordCount === 0  && this.props.fieldLocation === 'trailText');
```
akash1810: minor: could refactor to be a little DRYer:

```js
const requiresValidation = this.state.wordCount === 0  && this.props.fieldLocation === 'trailText';
const hasWarning = requiresValidation && this.props.isDesired;
const hasError = requiresValidation && this.props.isRequired;
```

public/video-ui/src/components/Video/Display.js
```diff
@@ -28,14 +29,38 @@ class VideoDisplay extends React.Component {
     }
   }
 
+  canonicalVideoPageExists = () => {
+    return this.props.usages.totalVideoPages > 0;
```
akash1810: we do this in a few places, should we add it to [video utils](https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/util/video.js) so its shared?

public/video-ui/src/components/Video/Display.js
```diff
@@ -28,14 +29,38 @@ class VideoDisplay extends React.Component {
...
       this.props.videoActions.createVideo(video)
         .then(() => {
           this.props.videoActions.getUsages(this.props.video.id);
         });
     } else {
-      this.props.videoActions.saveVideo(video);
+      this.props.videoActions.saveVideo(video)
+      .then(() => {
+        if (this.canonicalVideoPageExists()) {
+          const videoBlock = getVideoBlock(
```
akash1810: are we sure we want to update published Composer pages on atom save too?

public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -81,11 +81,13 @@ class VideoData extends React.Component {
               name="Trail Text"
               maxCharLength={fieldLengths.description.charMax}
               maxLength={fieldLengths.description.max}
-              isDesired={true}
+              isDesired={this.props.canonicalVideoPageExists ? false : true}
```
akash1810: can do `isDesired={!this.props.canonicalVideoPageExits}` instead of a ternary

public/video-ui/src/components/Video/Display.js
Reettaphant: Good point, we probably shouldn't because otherwise publishing an atom doesn't make much sense. I'll need to update my composer endpoint pr to support this.

public/video-ui/src/services/VideosApi.js
```diff
@@ -235,26 +237,28 @@ export default {
...
-            const videoPagePromise = this.addVideoToComposerPage({
-              composerId,
-              previewData: videoBlock,
-              composerUrlBase
-            });
-
-            return Promise.all([...fieldPromises, videoPagePromise]);
-          })
-        );
+        if (updatesTo === 'preview' && state === 'preview' || state === 'published' && updatesTo === 'published') {
```
akash1810: can this be `if state === updatesTo`?

public/video-ui/src/components/Video/Display.js
```diff
@@ -29,14 +31,38 @@ class VideoDisplay extends React.Component {
...
+      this.props.videoActions.saveVideo(video)
+      .then(() => {
+        if (canonicalVideoPageExists(this.props.usages)) {
+          const videoBlock = getVideoBlock(
+            this.props.video.id,
+            this.props.video.title,
+            this.props.video.source
+          );
+
+          const updateLive = this.props.usages.data.published.video.length === 0;
```
akash1810: these two variables don't seem to be used?

Reettaphant:
@akash1810 This has now been rebased and is ready to be merged once I've asked central prod to sent out comms.

PR #657, opened 2017-11-02
Rv scheduled launch

7091lapS:
Adds a scheduled launch date functionality. The feature is currently activated with the query `showScheduler=true`

### DIFF
--- a/public/video-ui/src/components/FormFields/DatePicker.js
+++ b/public/video-ui/src/components/FormFields/DatePicker.js
@@ -3,11 +3,12 @@ import Picker from 'react-datepicker';
 import moment from 'moment';
 import 'react-datepicker/dist/react-datepicker.css';
 import Icon from '../Icon';
+import _range from 'lodash/fp/range';
 
-const DATE_FORMAT = 'YYYY/MM/DD';
+const DATE_FORMAT = 'DD MMM YYYY';
 const DATETIME_FORMAT = `${DATE_FORMAT} HH:mm`;
 
-const MINUTES = [0, 15, 30, 45].map(minute =>
+const MINUTES = _range(0, 60).map(minute =>
   moment().minute(minute).format('mm')
 );
 const HOURS = [...new Array(24).keys()].map(hour =>
@@ -78,7 +79,7 @@ function Editor({ date, onChange, fieldName }) {
 
   return (
     <div>
-      <label className="form__label">{fieldName}</label>
+      {fieldName && <label className="form__label">{fieldName}</label>}
       <div className="expiry-date-picker">
         <div className="expiry-date-picker__date">
           <DateSelector date={date} onChange={onChange} />
@@ -89,11 +90,14 @@ function Editor({ date, onChange, fieldName }) {
         <div className="expiry-date-picker__number">
           <MinuteSelector date={date} onChange={onChange} />
         </div>
-        <Icon
-          icon="cancel"
-          className="icon__edit expiry-date-picker__reset"
-          onClick={reset}
-        />
+        {
+          date &&
+            <Icon
+            icon="cancel"
+            className="icon__edit icon__cancel"
+            onClick={reset}
+          />
+        }
       </div>
     </div>
   );
@@ -130,6 +134,7 @@ export default function DatePicker({
       <Editor
         fieldName={fieldName}
         date={date}
+        placeholder={placeholder}
         onChange={newDate => {
           if (newDate) {
             onUpdateField(newDate.valueOf());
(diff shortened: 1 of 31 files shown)


public/video-ui/src/components/VideoPublishBar/VideoPublishBar.js
```diff
@@ -20,6 +21,7 @@ export default class VideoPublishBar extends React.Component {
 
   isPublishingDisabled() {
     return (
+      this.props.video.contentChangeDetails.scheduledLaunch ||
```
Reettaphant: I think this line is breaking this check. I saw this branch on code and I was never able to publish the video.

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -1,6 +1,7 @@
 import { pandaReqwest } from './pandaReqwest';
 import { getStore } from '../util/storeAccessor';
 import getProductionOffice from '../util/getProductionOffice';
+// import VideoUtils from '../util/video';
```
Reettaphant: Why is this commented out?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -0,0 +1,101 @@
...
+            disabled={!video || videoEditOpen}
+          >
+            {scheduledLaunch ? 'Reschedule' : 'Schedule'}
+          </button>
+        }
+        {
+          showDatePicker &&
+          <button
+            className="button__secondary--confirm"
+            onClick={() => this.saveScheduledLaunch()}
```
Reettaphant: Why is the call to `saveScheduledLaunch` inside a function?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -0,0 +1,101 @@
...
+            disabled={!selectedDate || !isFutureDate}
+          >
+            Save
+          </button>
+        }
+        {
+          scheduledLaunch && showDatePicker &&
+          <button className="button__secondary--remove" onClick={() => this.removeScheduledLaunch()}>
```
Reettaphant: Why is the call to `removeScheduledLaunch` inside a function?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -0,0 +1,101 @@
...
+    const isFutureDate = selectedDate && moment(selectedDate).isAfter(moment());
+    const scheduledLaunch = contentChangeDetails && contentChangeDetails.scheduledLaunch && contentChangeDetails.scheduledLaunch.date;
+    return (
+      <div className="flex-container topbar__scheduler">
+        {scheduledLaunch && !showDatePicker && <span className="topbar__launch-label">Scheduled: {moment(scheduledLaunch).format('Do MMM YYYY HH:mm')}</span> }
+        {
+          showDatePicker &&
+          <DatePicker
+            editable={true}
+            onUpdateField={(date) => this.setState({ selectedDate: date })}
```
Reettaphant: Could we not use the `updateVideo` function here that saves the date to the redux store but not the database?

public/video-ui/src/services/WorkflowApi.js
7091lapS: the new stuff related to the scheduledLaunch in this file will have to be commented out until we get rid of scheduling in the workflow component

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
7091lapS: I decided to use the component's state to avoid incurring in the situation in which we have the value in the database out of sync with the value in the redux store. In this component's logic we are checking against the existence of that value in the database in order to show the scheduler vs the existing saved date. If the value in the store changed every time the value is updated but not saved, these checks would not work. To change this would require us to change the structure of the redux store, so I think keeping like this is easier.

public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -3,11 +3,12 @@ import Picker from 'react-datepicker';
...
 import 'react-datepicker/dist/react-datepicker.css';
 import Icon from '../Icon';
+import _range from 'lodash/fp/range';
 
-const DATE_FORMAT = 'YYYY/MM/DD';
+const DATE_FORMAT = 'DD MMM YYYY';
 const DATETIME_FORMAT = `${DATE_FORMAT} HH:mm`;
 
-const MINUTES = [0, 15, 30, 45].map(minute =>
+const MINUTES = _range(0, 60).map(minute =>
```
akash1810: We probably also want to change the expiry lambda to run every minute now too, as opposed to [every 15 minutes](https://github.com/guardian/media-atom-maker/blob/master/cloudformation/media-atom-maker.yaml#L521).

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
Reettaphant: 👍

PR #665, opened 2017-11-03
move MediaAtom model to common

akash1810:
also moves:
- `ThriftUtil`
- `JsonConversions`
- `MediaAtomImplicits`

This'll mean we can use our version of `MediaAtom` in `common`, which has lots of useful helpers, instead of the thrift model


common/src/main/scala/com/gu/media/util/MediaAtomImplicits.scala
```diff
@@ -18,11 +18,29 @@ trait MediaAtomImplicits extends AtomImplicits[MediaAtom] {
     }
   }
 
+  // why not use a Play template? `common` doesn't have a dependency on Play
   def defaultMediaHtml(atom: MediaAtom): String = {
     val asset = getActiveAsset(atom)
     val posterUrl = atom.posterUrl
 
-    views.html.MediaAtom.defaultEmbed(asset, posterUrl).toString()
+    (asset, posterUrl) match {
```
akash1810: this change isn't pretty, but felt like the simplest solution

akash1810:
deployed on CODE and this doesn't seem to break anything 😌

PR #668, opened 2017-11-03
default source field to The Guardian

akash1810:
This is frequently requested. Defaulting means a faster workflow and fewer mistakes.

Create page now looks like this:
image

### DIFF
--- a/public/video-ui/src/reducers/videoReducer.js
+++ b/public/video-ui/src/reducers/videoReducer.js
@@ -4,7 +4,7 @@ export default function video(state = null, action) {
   switch (action.type) {
     case 'VIDEO_GET_RECEIVE':
       return action.video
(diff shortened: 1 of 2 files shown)


public/video-ui/src/reducers/videoReducer.js
```diff
@@ -4,7 +4,7 @@ export default function video(state = null, action) {
   switch (action.type) {
     case 'VIDEO_GET_RECEIVE':
       return action.video
-        ? Object.assign({}, blankVideoData, action.video)
+        ? Object.assign({}, blankVideoData, {source: null}, action.video)
```
akash1810: don't artificially default the value if the server says its blank

akash1810:
[Video source drop-down](https://trello.com/c/0Y78zrJI/98-video-source-drop-down)

PR #680, opened 2017-11-14
Privacy status

Reettaphant:
Determine the privacy statuses of a channel according to the permissions of the user. 

Not everything that needs a list of youtube channels needs to know about privacy statuses so this pr also separates adding additional data to the channels from fetching them from youtube and caching the results. 

See also:
https://github.com/guardian/permissions/pull/87

### DIFF
--- a/app/util/YouTube.scala
+++ b/app/util/YouTube.scala
@@ -5,7 +5,6 @@ import com.gu.media.model.PrivacyStatus
 import com.gu.media.youtube._
 import com.typesafe.config.Config
 import play.api.cache.CacheApi
-import com.gu.pandomainauth.model.{User => PandaUser}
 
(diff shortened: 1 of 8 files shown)


app/util/YouTube.scala
```diff
@@ -20,6 +20,10 @@ trait YouTube extends Logging with YouTubeAccess with YouTubeVideos with YouTube
     cache.getOrElse("channels", duration) { super.channels }
   }
 
+  override def channelsWithData(setAllVideosPublic: Boolean): List[YouTubeChannelWithData] = {
```
mbarton: Do we need this since it's just forwarding on to the superclass?

Reettaphant: No we don't. It is gone.

PR #681, opened 2017-11-15
parse json

Reettaphant:
Workflow will not be returning JSON in the future so we need to parse workflow responses. 

Checks whether we are getting back json or a string before parsing so that we can deploy this before changes to workflow go live.


marialivia16:
@akash1810 I've updated [wf-frontend to use Circe](https://github.com/guardian/workflow-frontend/pull/110) and the `JSON` from Circe is not supported by the play's results. That's why we have to send strings instead 😞

PR #684, opened 2017-11-17
Av add embargo

7091lapS:
Atoms can now be embargoed.
NB: Embargo and Schedule Launch are non-blocking for now.

### DIFF
--- a/public/video-ui/src/components/Icon.js
+++ b/public/video-ui/src/components/Icon.js
@@ -91,7 +91,7 @@ export default class Icon extends React.Component {
 
     return (
       <span className={props.className}>
-        <i className="icon responsive--primary">{props.icon}</i>
+        <i className="icon responsive--primary" onClick={props.onClick}>{props.icon}</i>
         {this.renderText()}
       </span>
     );
--- a/public/video-ui/src/components/ScheduleRecap/ScheduleRecap.js
+++ b/public/video-ui/src/components/ScheduleRecap/ScheduleRecap.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import moment from 'moment';
+import PropTypes from 'prop-types';
+
+const ScheduleRecap = ({ scheduledLaunch, embargo, hasPreventedPublication }) => (
+  !hasPreventedPublication ?
+  <div className="topbar__launch-label">
+    <div>
+      <span className="scheduledSummary--scheduledLaunch">
+        {'Scheduled: '}
+      </span>
+      {scheduledLaunch
+        ? moment(scheduledLaunch).format('Do MMM YYYY HH:mm')
+        : '-'}
+    </div>
+    <div>
+      <span className="scheduledSummary--embargo">
+        {'Embargoed: '}
+      </span>
+      {embargo ? moment(embargo).format('Do MMM YYYY HH:mm') : '-'}
+    </div>
+  </div>
+  :
+  <div className="topbar__launch-label">
+    <span className="scheduledSummary--embargo">
+      Embargoed indefinitely
+    </span>
+  </div>
+);
+
+ScheduleRecap.propTypes = {
+  scheduledLaunch: PropTypes.number,
+  embargo: PropTypes.number,
+  hasPreventedPublication: PropTypes.bool
+};
+
+export default ScheduleRecap;
--- a/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
+++ b/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
@@ -1,7 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import DatePicker from '../FormFields/DatePicker';
-import moment from 'moment';
+import Icon from '../Icon';
+import ScheduleRecap from '../ScheduleRecap/ScheduleRecap';
+import {
+  isFutureDate,
+  isSameOrAfter,
+  isAfter
+} from '../../util/dateHelpers';
+import { impossiblyDistantDate }  from '../../constants/dates';
+import datesProperties from '../../constants/datesProperties';
 
 export default class ScheduledLaunch extends React.Component {
   static propTypes = {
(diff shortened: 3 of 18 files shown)


public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+    this.setState({ invalidDateError: null });
+
+    if (!isFutureDate(date)) {
+      this.setState({ invalidDateError: 'Date must be in the future!' });
+    }
+
+    if (
+      actionType === 'schedule' &&
+      embargo &&
+      isFutureDate(date) &&
```
Reettaphant: Was this check not already made on line 58?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+    ) {
+      this.setState({
+        invalidDateError: "Scheduled launch can't be earlier than embargo!"
+      });
+    }
+
+    if (
+      actionType === 'embargo' &&
+      scheduledLaunch &&
+      isFutureDate(date) &&
```
Reettaphant: Was this check not already made on line 58?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+      showDatePicker: true,
+      actionType,
+      showScheduleOptions: false
+    });
+  };
+
+  preventPublication = () => {
     const video = this.props.video;
-    this.props.saveVideo(Object.assign({}, video, {
+    const impossiblyDistantDate = 16725225600000;
```
Reettaphant: I'd put this somewhere more visible, to the top of the file or in the constants folder.

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.js
```diff
@@ -52,6 +53,9 @@ export default class VideoPublishBar extends React.Component {
     this.props.publishVideo();
   };
 
+  hasPublishedVideoUsages = () =>
+    this.props.usages.data.published.video.length > 0;
```
Reettaphant: I just started thinking about this and maybe we don't need this check at all now that we are going to allow embedding unpublished atoms and we are going to have warnings in composer when embedding unpublished atoms. Maybe we should trust our users to read these warnings. Maybe check this with editorial when we next catch up with them.

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+        })
+      })
+    );
+    this.setState({ showDatePicker: false, [selectedDate]: null });
+  };
+
+  /* Render functions */
+
+  renderScheduleOptions = (video, videoEditOpen, scheduledLaunch, embargo) => {
+    const hasPreventedPublication = embargo && embargo > 16693689600000;
```
Reettaphant: The future date should be a separate named constant.

public/video-ui/src/components/Icon.js
```diff
@@ -85,7 +85,7 @@ export default class Icon extends React.Component {
       
     return (
       <span className={props.className}>
-        <i className="icon responsive--primary">{props.icon}</i>
+        <i className="icon responsive--primary" onClick={props.onClick}>{props.icon}</i>
```
Reettaphant: Is this going to cause errors if we do not pass in onClick?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+    this.validateDate(date, actionType);
+  };
+
+  saveDate = actionType => {
+    const isSchedule = actionType === 'schedule';
+    const key = isSchedule ? 'scheduledLaunch' : 'embargo';
+    const selectedDate =
+      actionType === 'schedule'
+        ? 'selectedScheduleDate'
+        : 'selectedEmbargoDate';
```
Reettaphant: Could we somehow reduce the amount of different variable names going on here? Instead of actionType, could we just have a string called propertyName and could the propertyName be also used to identify the selectedDate?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +17,338 @@ export default class ScheduledLaunch extends React.Component {
...
+      contentChangeDetails &&
+      contentChangeDetails.scheduledLaunch &&
+      contentChangeDetails.scheduledLaunch.date;
+    const embargo =
+      contentChangeDetails &&
+      contentChangeDetails.embargo &&
+      contentChangeDetails.embargo.date;
+    const hasPreventedPublication = embargo && embargo > 16693689600000;
+
     return (
```
Reettaphant: I'd split what comes next into separate render functions e.g. renderPreventedPublication, renderDates, renderDataPicker. You could nest these functions as well and end up with less complicated statements as well when deciding which components to render.

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
akash1810: in addition to @Reettaphant 's suggestion, I'd either add a comment explaining its value (year 3000) or express it as a moment function `moment('3000-01-01).toString()` (or whatever the correct syntax is...)

public/video-ui/src/components/Videos/VideoItem.js
```diff
@@ -40,29 +41,56 @@ export default class VideoItem extends React.Component {
 
   render() {
     const video = this.props.video;
-    const scheduledLaunch = video.contentChangeDetails.scheduledLaunch && video.contentChangeDetails.scheduledLaunch.date;
+    const scheduledLaunch =
```
akash1810: could we use this to make things DRYer? https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/util/video.js#L96

public/video-ui/src/components/Videos/VideoItem.js
```diff
@@ -40,29 +41,56 @@ export default class VideoItem extends React.Component {
...
+      video.contentChangeDetails.scheduledLaunch &&
+      video.contentChangeDetails.scheduledLaunch.date;
+    const embargo =
+      video.contentChangeDetails.embargo &&
```
akash1810: could we extract this logic to the `VideoUtils` helper? https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/util/video.js

public/video-ui/src/components/Icon.js
7091lapS: I don't think so, `onClick` will just be `undefined`

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +21,310 @@ export default class ScheduledLaunch extends React.Component {
...
+    const scheduledLaunch =
+      contentChangeDetails &&
+      contentChangeDetails.scheduledLaunch &&
+      contentChangeDetails.scheduledLaunch.date;
+    const embargo =
+      contentChangeDetails &&
+      contentChangeDetails.embargo &&
+      contentChangeDetails.embargo.date;
+
+    this.setState({ invalidDateError: null });
```
Reettaphant: It feels  a bit strange that we are setting this to null here and then potentially setting it to something else below.

public/video-ui/src/components/ScheduleRecap/ScheduleRecap.js
```diff
@@ -0,0 +1,37 @@
...
+
+const ScheduleRecap = ({ scheduledLaunch, embargo, hasPreventedPublication }) => (
+  !hasPreventedPublication ?
+  <div className="topbar__launch-label">
+    <div>
+      <span className="scheduledSummary--scheduledLaunch">
+        {'Scheduled: '}
+      </span>
+      {scheduledLaunch
+        ? moment(scheduledLaunch).format('Do MMM YYYY HH:mm')
```
Reettaphant: This format seems to be used in quite a few places as well - it could also live in a constants file.

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +21,310 @@ export default class ScheduledLaunch extends React.Component {
...
+
+    if (!isFutureDate(date)) {
+      this.setState({ invalidDateError: 'Date must be in the future!' });
+    }
+
+    if (
+      propertyName === 'selectedScheduleDate' &&
```
Reettaphant: The property name could also be defined in constants, like [this](https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/constants/TagTypes.js).

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -2,6 +2,16 @@ import React from 'react';
...
+import Icon from '../Icon';
+import ScheduleRecap from '../ScheduleRecap/ScheduleRecap';
+
+const isFutureDate = date => date && moment(date).isAfter(moment());
+const isSameOrAfter = (dateA, dateB) =>
+  moment(dateA).isSameOrAfter(moment(dateB));
+const isAfter = (dateA, dateB) => moment(dateA).isAfter(moment(dateB));
+
+const impossiblyDistantDate = moment('2499-01-01T00:00:00+00:00').valueOf();
+const laterThanImpossiblyDistantDate = moment('2500-01-01T00:00:00+00:00').valueOf();
```
Reettaphant: Not sure why we need both `impossiblyDistantDate` and `laterThanImpossiblyDistantDate`. Can't we just check if a date is not before `impossiblyDistantDate` to find out if it is embargoed indefinitely?

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -11,89 +21,310 @@ export default class ScheduledLaunch extends React.Component {
...
-    const scheduledLaunch = contentChangeDetails && contentChangeDetails.scheduledLaunch && contentChangeDetails.scheduledLaunch.date;
+    const invalidDateError = this.state.invalidDateError;
+    const scheduledLaunch =
+      contentChangeDetails &&
+      contentChangeDetails.scheduledLaunch &&
+      contentChangeDetails.scheduledLaunch.date;
+    const embargo =
+      contentChangeDetails &&
+      contentChangeDetails.embargo &&
+      contentChangeDetails.embargo.date;
```
Reettaphant: Did we not already access the scheduled launch of the video on lines 50-58?

Reettaphant:
Just remembered that we also need to make sure that scheduled launch/expiry/indefinite embargo get sent to the composer video pages correctly. I'm happy to help with this!

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
7091lapS: I need to access them in the `render()` function, but I'd like to avoid having to call `validateDate()` with two extra arguments, as it is called by both `setDate()` and `onSelectOption()`, so then I would have to pass the extra arguments to these two functions as well.

public/video-ui/src/components/ScheduleRecap/ScheduleRecap.js
7091lapS: I think I'd rather keep this explicit, because in `VideoItem` we are using a slightly different format (`'D MMM HH:mm'` vs `'Do MMM HH:mm'`), and now it's easy to spot the difference

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -2,6 +2,16 @@ import React from 'react';
...
 import moment from 'moment';
+import Icon from '../Icon';
+import ScheduleRecap from '../ScheduleRecap/ScheduleRecap';
+
+const isFutureDate = date => date && moment(date).isAfter(moment());
+const isSameOrAfter = (dateA, dateB) =>
+  moment(dateA).isSameOrAfter(moment(dateB));
+const isAfter = (dateA, dateB) => moment(dateA).isAfter(moment(dateB));
+
+const impossiblyDistantDate = moment('2499-01-01T00:00:00+00:00').valueOf();
```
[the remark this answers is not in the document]
7091lapS: 👍

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
7091lapS: 👍

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
7091lapS: I think this was to make sure that the alerts from previous failed validations would be cleared after closing the dropdown and selecting a new option.

Reettaphant: Ok, that makes sense. You could put the whole logic setting the `invalidDateError` inside an if-else with the else clause setting `invalidDateError` to null

public/video-ui/src/components/ScheduleRecap/ScheduleRecap.js
Reettaphant: We could make them both constants of course can keep them like this as well

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.js
7091lapS: we're adding a card for this and it will be in a separate PR

PR #690, opened 2017-11-20
refactor to avoid passing in undefined args

akash1810:
hasYoutubeWriteAccess is also used in AddYouTubeAsset, however the second arg wasn't being set, lets just calculate the availablePrivacyStates within the function


akash1810:
> I'm assuming that keeping the signature as an object is so that you could feed in a video and `channelId` is a property of a video?

Exactly. This PR just keeps things a bit neater.

PR #691, opened 2017-11-21
allow for a PAC form to be uploaded next to a video

akash1810:
You can choose an xml file next to your video in S3 for eventual consumption by a Pluto service.

You can upload the xml file independently of the video.

Upload to the server that then pushes it to S3 and sends a message on the kinesis stream for Pluto to consume.

image


public/video-ui/src/components/VideoUpload/YoutubeUpload.js
```diff
@@ -0,0 +1,67 @@
...
+
+    return (
+      <div className="video__detailbox video__detailbox__assets">
+        <div className="video__detailbox__header__container">
+          <header className="video__detailbox__header">
+            Upload to YouTube
+          </header>
+        </div>
+        <div className="form__group">
+          <ManagedForm
```
Reettaphant: Do we get anything useful from the managed form here when we are not using the managed field component?

public/video-ui/src/components/VideoUpload/YoutubeUpload.js
```diff
@@ -0,0 +1,67 @@
...
+            data={video}
+            updateData={this.props.saveVideo}
+            editable={true}
+            formName="YouTubeUpload"
+          >
+            <input
+              className="form__field"
+              type="file"
+              onChange={this.setFile}
+              disabled={!canUploadToYouTube || this.props.uploading}
```
Reettaphant: if they cannot upload to youtube because of missing data, should we tell them why - it could be confusing because teh missing fields are on the main video page.

akash1810: 👍 

image

public/video-ui/src/components/VideoUpload/YoutubeUpload.js
akash1810: good point! removed

akash1810:
@Reettaphant I've redone the server side logic after talking with @fredex42 about how/when to send a message on the stream

common/src/main/scala/com/gu/media/upload/PlutoUploadActions.scala
```diff
@@ -2,32 +2,39 @@ package com.gu.media.upload
...
-//          config.replyToAddresses)
-//
-//        plutoStore.put(plutoData)
+  def sendToPluto(plutoIntegrationData: PlutoIntegrationData): Unit = {
+    plutoIntegrationData match {
+      case plutoData: PlutoSyncMetadata => {
+        plutoData.projectId match {
+          case Some(_) => sendKinesisMessage(plutoIntegrationData)
+          case None => {
+            // TODO MRB: re-enable this once the project selector has been fixed
```
Reettaphant: Is the project selector fixed in this branch already?

akash1810: sadly no, I'm doing that now in a different PR

PR #692, opened 2017-11-21
move create page from /videos/create to /create

akash1810:
- its shorter to type
- edit /videos/create to a redirect to /create


akash1810:
> can't see where the original /videos/create was in conf/routes

Yeah, that's confused me too. https://github.com/guardian/media-atom-maker/blob/master/conf/routes#L51 implicitly catches it.

PR #693, opened 2017-11-22
linty mclintface

akash1810:
lint the js


akash1810:
We [had a hook](https://github.com/guardian/media-atom-maker/pull/515) but it was painful when WIPing a branch.

Maybe it should be a pre-post hook rather than a pre-commit one?

PR #697, opened 2017-11-27
Av move expiry date

7091lapS:
Add an expiry field to contentChangeDetails, as part of the work to be able to query capi by date range

### DIFF
--- a/public/video-ui/src/components/VideoData/VideoData.js
+++ b/public/video-ui/src/components/VideoData/VideoData.js
@@ -125,7 +125,7 @@ class VideoData extends React.Component {
             </ManagedField>
           </ManagedSection>
           <ManagedSection>
-            <ManagedField fieldLocation="expiryDate" name="Expiry Date">
+            <ManagedField fieldLocation="contentChangeDetails.expiry.date" name="Expiry Date">
               <DatePicker />
             </ManagedField>
             <ManagedField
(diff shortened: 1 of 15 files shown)


public/video-ui/src/components/VideoData/VideoData.js
```diff
@@ -125,7 +125,7 @@ class VideoData extends React.Component {
             </ManagedField>
           </ManagedSection>
           <ManagedSection>
-            <ManagedField fieldLocation="expiryDate" name="Expiry Date">
+            <ManagedField fieldLocation="contentChangeDetails.expiry.date" name="Expiry Date">
```
akash1810: I don't think this'll work for older atoms as this field will be empty until we re-index.

public/video-ui/src/reducers/publishedVideoReducer.js
```diff
@@ -3,9 +3,33 @@ import { blankVideoData } from '../constants/blankVideoData';
...
+        /*
+        this is necessary to make sure that if the old expiryDate field already exists in the db,
+        the new expiry value is initialised to the saved one. This won't be necessary
+        once we remove the old field
+        */
+        const newExpiryDate =
+          video.contentChangeDetails.expiry &&
+          video.contentChangeDetails.expiry.date;
+        if (!newExpiryDate && video.expiryDate) {
+          video = Object.assign({}, video, {
```
akash1810: could we just `return` rather than mutate `video` (also means we can use a `const` instead of a `let` above)

PR #700, opened 2017-11-30
allow for embedding preview atoms

Reettaphant:


public/video-ui/src/components/Video/Display.js
```diff
@@ -207,11 +208,14 @@ class VideoDisplay extends React.Component {
   }
 
   renderSelectBar(video) {
+
+    const videoToSelect = isVideoPublished(this.props.publishedVideo) ?
```
akash1810: this is a weird statement `isVideoPublished(this.props.publishedVideo)` 😂

Reettaphant: Yeah, makes more sense to pass in `this.props.video` here.

PR #701, opened 2017-11-30
update client side, allowing to select a commission and project

akash1810:
requires #696 

pluto


public/video-ui/src/components/Pluto/PlutoProjectPicker.js
```diff
@@ -0,0 +1,94 @@
...
+            data={video}
+            updateData={saveVideo}
+            editable={true}
+            formName="Pluto"
+          >
+            <ManagedField
+              fieldLocation="plutoData.commissionId"
+              name="Commission"
+              isRequired={false}
+              updateSideEffects={() => this.onCommissionSelection()}
```
Reettaphant: Why is the call to `onCommissionSelection` wrapped inside the function?

public/video-ui/src/components/ManagedForm/ManagedField.js
```diff
@@ -100,6 +100,10 @@ export class ManagedField extends React.Component {
         _set(this.props.fieldLocation, null, this.props.data)
       );
     }
+
+    if (this.props.updateSideEffects) {
```
Reettaphant: This function is used in the [tagpicker](https://github.com/guardian/media-atom-maker/blob/0d2a975d6ffde2ab6fd8a83a3978361271c02d0c/public/video-ui/src/components/FormFields/TagPicker.js#L346) as well but inside the component to copy tags. Does the function call here cause errors from the tagpicker, especially since the function in the tagpicker does not take any arguments.

public/video-ui/src/components/Pluto/PlutoProjectPicker.js
akash1810: I need access to `this.props.video`

public/video-ui/src/components/ManagedForm/ManagedField.js
akash1810: good call, will check!

akash1810: tag picker is working as usual :)

public/video-ui/src/components/Pluto/PlutoProjectPicker.js
```diff
@@ -0,0 +1,99 @@
...
+        if (plutoData.commissionId) {
+          this.props.plutoActions.getProjects(plutoData);
+        }
+      });
+    }
+  }
+
+  onCommissionSelection() {
+    this.props.saveVideo(
+      _set('plutoData.projectId', null, this.props.video)
```
akash1810: we've changed the commission, so we need to erase the `projectId` and save the atom

PR #704, opened 2017-12-01
Av improve expirer

7091lapS:
- Change the Expirer lambda to query CAPI for a day's worth of expired atoms and set their youtube status to private
- Change Cloudformation to run lambda every minute


akash1810:
we need to reindex before deploying this in order to get the new field populated

PR #708, opened 2017-12-04
on update should always return a promise

Reettaphant:
`onUpdate` sometimes only updates the video and does not save it so it wasn't always returning a promise and was throwing an error when we were assuming it was in `managedForm`. This pr makes sure that all `onUpdate` functions return promises.

### DIFF
--- a/public/video-ui/src/components/Video/Display.js
+++ b/public/video-ui/src/components/Video/Display.js
@@ -41,6 +41,7 @@ class VideoDisplay extends React.Component {
 
   updateVideo = video => {
     this.props.videoActions.updateVideo(video);
+    return Promise.resolve();
   };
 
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/Video/Display.js
```diff
@@ -41,6 +41,7 @@ class VideoDisplay extends React.Component {
 
   updateVideo = video => {
     this.props.videoActions.updateVideo(video);
+    return Promise.resolve();
```
akash1810: why not return `this.props.videoActions.updateVideo(video);`?

Reettaphant: Because it doesn't return a promise - it's not actually saving anything to the database, only to the redux store.

PR #710, opened 2017-12-06
Add new endpoints to reindex atoms with new expiry date field

7091lapS:


app/model/commands/ExpiryFieldPreviewCommand.scala
```diff
@@ -0,0 +1,67 @@
...
+        atoms => atoms.map(atom => reIndex(MediaAtom.fromThrift(atom)))
+      }
+    )
+  }
+
+  def reIndex(atom: MediaAtom) = {
+    log.info(s"Request to update atom ${atom.id}")
+    val expiry: Option[DateTime] = atom.expiryDate.map(expiry => new DateTime(expiry))
+
+    val details = atom.contentChangeDetails.copy(
```
akash1810: We're updating all atoms here, I think it'll be better if we only update the ones that have the old field set:


```scala
expiry match {
  case expDate: Some(DateTime) => {
    // update store and send message
  }
  case None => {
    // don't do anything, just return the same atom
    atom
  }
}

```

PR #713, opened 2017-12-08
refactor `/videos/pluto-list` to use real data

akash1810:
- add commission/project selection to pluto-list page, re-using component on the upload page
- re-use the saveVideo action
- delete unused endpoints
- send message on kinesis to notify Pluto when an atom has a project set
- update `/missing` to include atoms without a project
- re-enable the mailer

image


app/controllers/Api2.scala
```diff
@@ -162,16 +162,18 @@ class Api2 (override val stores: DataStores, conf: Configuration, override val a
...
+    val unprocessedAssetResponses: List[PlutoSyncMetadataMessage] = stores.pluto.list()
 
     val uploadsWithoutPlutoId = unprocessedAssetResponses.foldRight(Map[String, MediaAtom]())((upload, acc) => {
       if (!acc.contains(upload.atomId)) {
         previewDataStore.getAtom(upload.atomId) match {
           case Right(atom) => {
             val mediaAtom = MediaAtom.fromThrift(atom)
+
             mediaAtom.plutoData match {
+              case Some(plutoData) if plutoData.projectId.isEmpty => acc ++ Map(upload.atomId -> mediaAtom)
```
Reettaphant: Could you write 
```
case Some(plutoData) ifplutoData.projectId.exists => acc 
case _ => acc ++ Map(upload.atomId -> mediaAtom)
```
to simplify pattern matching?

app/model/commands/UpdateAtomCommand.scala
```diff
@@ -69,6 +75,20 @@ case class UpdateAtomCommand(id: String, atom: MediaAtom, override val stores: D
     )
   }
 
+  private def processPlutoData(oldAtom: MediaAtom, newAtom: MediaAtom) = {
+    (oldAtom.plutoData.flatMap(_.projectId), newAtom.plutoData.flatMap(_.projectId)) match {
+      case (Some(oldProject), Some(newProject)) if oldProject != newProject => notifyPluto(newAtom)
+      case (None, Some(_)) => notifyPluto(newAtom)
+      case (_, _) => None
```
Reettaphant: If we don't do anything in the case of missing project ids here, are we still saving projects without pluto project ids to be looked at later?

Reettaphant:
If I've understood this correctly, we are only sending data to pluto once we publish an atom. So if someone adds a pluto project to an atom, does the project only get sent to pluto once we publish the atom? This could be confusing, especially since I think the atom itself will disappear from the list when a pluto id is added to it.

app/model/commands/UpdateAtomCommand.scala
akash1810: This serves to send a message to Pluto if the project property of an Atom has changed.

> are we still saving projects without pluto project ids to be looked at later?

Yes, however we do this on video upload as we talk to Pluto to tell it to ingest the video. During upload:

```
if !projectId set:
  [write to table](https://github.com/guardian/media-atom-maker/blob/master/common/src/main/scala/com/gu/media/upload/PlutoUploadActions.scala#L18-L19)
  send email to user
else:
  [send message](https://github.com/guardian/media-atom-maker/blob/master/common/src/main/scala/com/gu/media/upload/PlutoUploadActions.scala#L17) to pluto to ingest newly uploaded video
```

Doing so here would be a little noisy as its the UpdateCommand, so an edit to the Atom title, description, etc would mean also writing to that table...

`processPlutoData ` isn't the clearest naming (suggestions welcomed) and maybe something like this would be clearer:

```scala
(oldAtom.plutoData.flatMap(_.projectId), newAtom.plutoData.flatMap(_.projectId)) match {
  case (Some(oldProject), Some(newProject)) if oldProject == newProject => None
  case (_, Some(_)) => notifyPluto(newAtom)
  case (_, _) => None
}
```

app/controllers/Api2.scala
akash1810: 👍

akash1810:
We send a message to Pluto in a couple of different places:

- In the UpdateCommand, if the project id has changed, we send [this message](https://github.com/guardian/media-atom-maker/blob/aa-pluto-list-selections/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala#L15)
- During a video upload, if the atom has a project assigned, we send [this message](https://github.com/guardian/media-atom-maker/blob/aa-pluto-list-selections/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala#L50)

The latter also writes to the missing table, which drives `/pluto-list`.

When we assign a project on that page, we save the atom via the UpdateCommand, so the former gets triggered.

There's also a third type of message [here](https://github.com/guardian/media-atom-maker/blob/aa-pluto-list-selections/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala#L33) that is send during a pac xml upload.

akash1810:
Rebased master to resolve the conflicts. @Reettaphant could you take a look when you've got a second?

app/model/commands/UpdateAtomCommand.scala
Reettaphant: Ok, that makes sense! I guess the function could be called something like `processChangesToPlutoData` but not sure it makes much of a difference.

PR #714, opened 2017-12-08
Add scheduler lambda

7091lapS:
- Add lambda to launch scheduled videos at the correct time
- If an embargo date exists, videos will only launch if the scheduled launch date is later than the embargo date
- Add tests
- scheduledLaunch filter added to CAPI here: https://github.com/guardian/content-api/pull/2075


scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
```diff
@@ -0,0 +1,79 @@
...
+    val scheduledLaunchDate = (atom \ "contentChangeDetails" \ "scheduledLaunch" \ "date").as[Long]
+
+    (atom \ "contentChangeDetails" \ "embargo" \ "date").asOpt[Long] match {
+      case Some(embargoDate) if embargoDate < scheduledLaunchDate =>
+        val assets = (atom \ "data" \ "media" \ "assets").as[JsArray]
+        val videos = assets.value.filter { asset => (asset \ "platform").as[String] == "youtube" }
+        val ids = videos.map { asset => (asset \ "id").as[String] }
+
+        ids.toSet
+      case None =>
```
7091lapS: I am not sure how to refactor this to avoid repetition, as Scala doesn't allow to do something like ```case None | Some(embargoDate) if embargoDate < scheduledLaunchDate```.  Should I just extract the logic in a separate method?

regiskuckaertz: `val embargoDate = (atom \ "contentChangeDetails" \ "embargo" \ "date").asOpt[Long].getOrElse(Long.MAX_VALUE)`

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
```diff
@@ -0,0 +1,79 @@
...
+      } catch {
+        case NonFatal(err) =>
+          log.error(s"Unable to launch $video", err)
+      }
+    }
+  }
+
+  @tailrec
+  private def getVideosFromScheduledAtoms(page: Int, pageSize: Int, oneDayAgo: Instant, now: Instant, before: Set[String]): Set[String] = {
+    val url = s"atoms?types=media&page-size=$pageSize&from-date=$oneDayAgo&to-date=$now&use-date=scheduled-launch"
```
regiskuckaertz: You don't pass the `page` parameter, so this will loop forever.

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
```diff
@@ -0,0 +1,79 @@
...
+  with YouTubeVideos {
+
+  def scheduleInParallel = true // disabled in unit tests
+
+  override def handleRequest(input: Unit, context: Context): Unit = {
+    val now = Instant.now()
+    val oneDayAgo = Instant.now().minus(1, ChronoUnit.DAYS)
+    val assets = getVideosFromScheduledAtoms(1, 100, oneDayAgo, now, Set.empty).filter(isManagedVideo)
+
+    val toLaunch = if(scheduleInParallel) { assets.par } else { assets }
```
regiskuckaertz: How many videos are we talking about? Just tried a couple requests and got 0 results all the time. This may be a premature optimization.

regiskuckaertz:
(This is just an FYI) Seems like what you're doing is not a fold but an unfold, i.e. you are reading from a data structure (CAPI) that is potentially infinite, and at each iteration you have the opportunity to decide if you are done or not. Its signature is
```
def unfold[A, B](seed: B)(iter: B => Option[(A, B)]): List[A]
```
Each iteration can generate results and a new seed to run `iter` against. You can use the first page of results as the seed. Simplifying:
```
def queryCapi[A](from: Instant, until: Instant)(pageNo: Int): List[A] = ???
def getVideosFromScheduledAtoms(until: Instant): Set[String] = {
  val from = until.minus(1, ChronoUnit.DAYS)
  val queryPage = queryCapi(from, until)
  unfold(queryPage(1)) {
    case Nil => 
      None
    case results if results.currentPage < results.pages => 
      (results.map(getScheduledVideos), queryPage(results.currentPage + 1))
    case results =>
      (results.map(getSchedulesVideos), Nil)
  }.flatten.toSet
```
Then the definition of unfold is straightforward:
```
@tailrec
def unfold[A, B](seed: B)(iter: B => Option[(A, B)]): List[B] = iter(seed) match {
  case None => Nil
  case (result, next) => result :: unfold(next)(iter)
}
```
It nicely decouples your code from the recursion, which is great for readability.

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
7091lapS: Ok, this is a leftover from my refactor of a previous version of a similar lambda. The lambda used to run for all the existing atoms (before filtering by scheduledLaunch was possible in CAPI), so that's probably why the optimisation was there? Now that we are filtering and getting fewer results this shouldn't be needed anymore.

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
7091lapS: nice!

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
```diff
@@ -0,0 +1,70 @@
...
+    val after = results.foldLeft(before) { (acc, atom) => acc ++ getScheduledVideos(atom) }
+
+    if(currentPage < pages)
+      getVideosFromScheduledAtoms(page + 1, pageSize, oneDayAgo, now, after)
+    else
+      after
+  }
+
+  private def getScheduledVideos(atom: JsValue): Set[String] = {
+    val scheduledLaunchDate = (atom \ "contentChangeDetails" \ "scheduledLaunch" \ "date").as[Long]
```
akash1810: using the model should make things a bit easier:

```scala
import com.gu.contentatom.thrift.atom.media.{MediaAtom => ThriftMediaAtom}
import com.gu.media.model MediaAtom

val thriftAtom: ThriftMediaAtom = atom.validate[ThriftMediaAtom]
val mediaAtom: MediaAtom = MediaAtom.fromThrift(thriftAtom)
val scheduledLaunchDate: Option[ChangeRecord] = mediaAtom.contentChangeDetails.scheduledLaunch
...
```

However, I realise we're using `JsValue` elsewhere, so feel free to do this in a separate PR.

scheduler/src/main/scala/com.gu.media.scheduler/SchedulerLambda.scala
```diff
@@ -0,0 +1,70 @@
...
+  with YouTubeVideos {
+
+  override def handleRequest(input: Unit, context: Context): Unit = {
+    val now = Instant.now()
+    val oneDayAgo = Instant.now().minus(1, ChronoUnit.DAYS)
+    val assets = getVideosFromScheduledAtoms(1, 100, oneDayAgo, now, Set.empty).filter(isManagedVideo)
+
+    assets.foreach { video =>
+      try {
+        setStatus(video, PrivacyStatus.Public)
```
akash1810: I think this should be the atom's status, as `Unlisted` is a valid option.

7091lapS: Looking at this with Retta we realised that the work that this lambda needs to do might be slightly more complex than what it currently does: not only set the privacy of the active asset to public/unlisted, but also to push the updates to the video data yo youtube, as when you schedule it, it is not published. Would it make sense to get the lambda to call the `PublishAtomCommand` endpoint on MAM?
ps I'm on health this week so Retta might pick this up

PR #716, opened 2017-12-12
stricter capi query

akash1810:
use stricter types in the capi query interface

`Map[String, String]` is more comforting than `String`

### DIFF
--- a/app/data/AtomListStore.scala
+++ b/app/data/AtomListStore.scala
@@ -17,14 +17,29 @@ trait AtomListStore {
 class CapiBackedAtomListStore(capi: CapiAccess) extends AtomListStore {
   override def getAtoms(search: Option[String], limit: Option[Int]): MediaAtomList = {
     // CAPI max page size is 200
-    val cappedLimit = limit.map(Math.min(200, _))
+    val cappedLimit: Option[Int] = limit.map(Math.min(200, _))
 
-    val base = "atoms?types=media&order-by=newest"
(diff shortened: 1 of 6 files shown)


app/data/AtomListStore.scala
```diff
@@ -17,14 +17,29 @@ trait AtomListStore {
...
+        "q" -> q,
+        "searchFields" -> "data.title"
+      )
+      case None => base
+    }
 
-    val base = "atoms?types=media&order-by=newest"
-    val searchPart = search.map { q => s"&searchFields=data.title&q=$q" }.getOrElse("")
-    val pageSizePart = cappedLimit.map { l => s"&page-size=$l" }.getOrElse("")
+    val baseWithSearchAndLimit = cappedLimit match {
```
akash1810: I'm not overly impressed with this, but it felt less verbose than:

```scala
val qs = (search, cappedLimit) match {
  case (Some(q), Some(limit)) => base ++ Map(
    "q" -> q,
    "searchFields" -> "data.title",
    "page-size" -> limit.toString
  )
  case (Some(q), None) => base ++ Map(
    "q" -> q,
    "searchFields" -> "data.title"
  )
  case (None, Some(limit)) => base ++ Map(
    "page-size" -> limit.toString
  )
  case (None, None) => base
}

val response = capi.capiQuery("atoms", qs)
```

Reettaphant: I've been struggling with writing similar logic elsewhere, but I don't have any pretty solutions. Might be a little bit tidier to if you define each map (base, search, pageSize) separately and then add them together.

PR #717, opened 2017-12-13
observe schedule and embargo property on publish

akash1810:
- if embargo set, don't publish
- if schedule set for future, don't publish
- if schedule < embargo, don't publish
- else, publish

also extracted some code into functions for maintainability

easier to review w/out whitespace - https://github.com/guardian/media-atom-maker/pull/717/files?w=1

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -45,44 +45,66 @@ case class PublishAtomCommand(
       AtomPublishFailed("Atom status set to private")
     }
 
-    getActiveAsset(previewAtom) match {
-      case Some(asset) if asset.platform == Youtube =>
-        val duration = youtube.getDuration(asset.id)
+    val contentChangeDetails = thriftPreviewAtom.contentChangeDetails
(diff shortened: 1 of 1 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -45,44 +45,66 @@ case class PublishAtomCommand(
...
-              val hasMakePublicPermission = permissions.setVideosOnAllChannelsPublic
-
-              if (hasMakePublicPermission){
-                PrivacyStatus.Public
-              } else {
-                log.info(s"User ${user.email} does not have permission to publish atom ${previewAtom.id} as Public, setting as Unlisted")
-                PrivacyStatus.Unlisted
-              }
+    (contentChangeDetails.scheduledLaunch, contentChangeDetails.embargo) match {
+      case (_, Some(embargo)) => {
```
7091lapS: Shouldn't this also check ` if embargo.date > now`? Currently the embargo date value/label do not get automatically removed when the embargo date is reached

akash1810: good spot! changed.

PR #719, opened 2017-12-13
Release scheduling ui

Reettaphant:
Remove query string hiding the scheduler and send scheduling and embargo dates to workflow. Indefinite embargoes won't work yet in workflow but we will fix this!


akash1810:
rebased and deployed to CODE to demo to Editorial, they're happy with it 🎉 . They said it would be best to release on Monday as the comms are more likely to be read then 😄

Reettaphant:
I've added the logic to add indefiniteEmbargo on workflow track as well.

PR #722, opened 2017-12-14
refactor duration and the PublishAtomCommand

akash1810:
Calculating the duration requires a call to YouTube. Whilst this isn't a problem in general, making this call during publishing feels unnecessary as we can do it during asset activation. So, refactor ActiveAssetCommand too, simplifying the routes of entry; you only need an atom id and a version number to activate an asset.

The side-effect of this, is one less call to YouTube, reducing our API quota usage and also a slightly faster publishing route as there's one less call to a third party API.


app/model/commands/ActiveAssetCommand.scala
```diff
@@ -9,51 +9,44 @@ import data.DataStores
...
+) extends Command with MediaAtomImplicits with Logging {
 
   type T = MediaAtom
 
   def process(): T = {
-    log.info(s"Request to $params in $atomId")
+    log.info(s"Request to set active asset atom=$atomId version=${activateAssetRequest.version}")
+
+    if (atomId != activateAssetRequest.atomId) {
+      AtomIdConflict
```
Reettaphant: Should we be using `AtomIdConflict` here since ending up here would be a symptom of of something else having gone wrong?

akash1810: I was thinking we should add this check to all the endpoints, otherwise we're trusting the id in the url path matches the id in the request body. We shouldn't reach here, and if we do its probably safer to fail early. What do you think?

Reettaphant: Makes sense I guess. What do you think could cause for the two not to match? It might still save confusion to reserve the different error name for this case, sth like `nonMatchingAtomIds` maybe because `AtomIdConflict` is used when creating atoms if the new random id generated is already in use.

PR #723, opened 2017-12-14
clearer warning message

akash1810:
based on user feedback


akash1810:
pretty small, merging w/out a 👍

PR #724, opened 2017-12-15
observe expiry date on publish

akash1810:
Don't allow an atom to be published if it has expired. We already do this check client-side, so this just tighten things up a bit.

if the expiry date is <= now, fail


akash1810:
giphy isn't that great at `templating` results... 😞

PR #726, opened 2017-12-15
allow s3 uploder urls to be added as assets

akash1810:
this is to support an article going live next week that can't go to YouTube, so is kind of a hack as the better solution would be to use the self hosted route...

### DIFF
--- a/app/model/commands/AddAssetCommand.scala
+++ b/app/model/commands/AddAssetCommand.scala
@@ -1,7 +1,9 @@
 package model.commands
 
+import java.net.URI
+
 import com.gu.contentatom.thrift.Atom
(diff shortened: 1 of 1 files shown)


akash1810:
Added do not merge label as I think it'd be better if we fix the self hosted client side issues. That way we can benefit from permissions too.

mbarton:
@akash1810 looks fine to me. Is the plan to deploy this branch to PROD to make the atoms and then deploy master again?

akash1810:
So I had a look at the code again on Friday evening and the only restriction to uploading as self-hosted is permissions. I've raised this with Yusuf and he's happy to try that in the first instance.

I'll keep this open whilst the story is in progress.

PR #731, opened 2017-12-19
disable scheduler when info missing

Reettaphant:
There were already checks hiding the scheduler button if there are published video page usages. I thought it might be clearer to disable the button instead and display a message to the user, the same applies if the user erases the title from the atom.

### DIFF
--- a/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
+++ b/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
@@ -17,7 +17,8 @@ export default class ScheduledLaunch extends React.Component {
   static propTypes = {
     video: PropTypes.object.isRequired,
     saveVideo: PropTypes.func.isRequired,
-    videoEditOpen: PropTypes.bool.isRequired
+    videoEditOpen: PropTypes.bool.isRequired,
+    hasPublishedVideoPageUsages: PropTypes.func.isRequired
   };
 
   state = {
--- a/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
+++ b/public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
@@ -11,6 +11,7 @@ import {
 import { impossiblyDistantDate }  from '../../constants/dates';
 import datesProperties from '../../constants/datesProperties';
 import VideoUtils from '../../util/video';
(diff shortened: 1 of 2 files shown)


public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -250,26 +251,45 @@ export default class ScheduledLaunch extends React.Component {
...
-    </button>
-  )
+
+      if (this.props.hasPublishedVideoPageUsages()) {
+        return 'The atom cannot be scheduled because it has a published video page';
+      }
+
+      return null;
+
+    }.bind(this)());
```
akash1810: Why do we need to `.bind`? 😱

Reettaphant: Because I couldn't figure out the syntax to call an arrow function as I was defining it and then I decided to use bind to have access to `this.props`...

akash1810: we could move this function to a higher scope on the component itself?

```js
export default class ScheduledLaunch extends React.Component {
  getNoScheduleReason() {
    if (!this.props.video.title) {
      return 'You must add a title before scheduling';
    }
    ...
  }
  renderSchedulerButton(showScheduleOptions) {
    const notAbleToScheduleReason = this.getNoScheduleReason();
    const disabled = !!notAbleToScheduleReason;
    return (
      <div data-tip={notAbleToScheduleReason}>
        ...
      </div>
    );
  }
}
```

public/video-ui/src/components/ScheduledLaunch/ScheduledLaunch.js
```diff
@@ -130,6 +131,18 @@ export default class ScheduledLaunch extends React.Component {
     this.setState({ showDatePicker: false, [propertyName]: null });
   };
 
+  getNoScheduleReason = () => {
+    if (!this.props.video.title) {
+      return 'You must add a title before scheduling';
+    }
+
+    if (this.props.hasPublishedVideoPageUsages()) {
```
akash1810: can we add `hasPublishedVideoPageUsages` to the `propTypes` [above](https://github.com/guardian/media-atom-maker/pull/731/files#diff-07b21fff0b6992387e6c2e1c96395ef5R17)?

PR #733, opened 2017-12-19
add utility functions to get embargo and scheduled launch dates

Reettaphant:


public/video-ui/src/util/video.js
```diff
@@ -99,17 +99,29 @@ export default class VideoUtils {
...
-  }
-
   static canUploadToYouTube({ youtubeCategoryId, channelId, privacyStatus }) {
     return !!youtubeCategoryId && !!channelId && !!privacyStatus;
   }
 
-  static getEmbargo({ contentChangeDetails }) {
-    const embargo = contentChangeDetails.embargo;
-    return embargo ? moment(embargo.date) : null;
+  static getScheduledLaunch(contentChangeDetails) {
```
akash1810: minor: we could use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) here, then we can call this function with the video, e.g. `const scheduledLaunch = VideoUtils.getScheduledLaunch(this.props.video)`.

PR #734, opened 2017-12-19
explicitly mention "save"

akash1810:
James C and James D have mentioned this would be helpful in the last two days.

Before
before

After
after

### DIFF
--- a/public/video-ui/styles/layout/_icons.scss
+++ b/public/video-ui/styles/layout/_icons.scss
@@ -28,7 +28,7 @@
   &--text {
     margin-left: 5px;
   }
-  
+
   &:link {
     text-decoration: none;
   }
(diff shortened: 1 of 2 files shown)


public/video-ui/styles/layout/_icons.scss
```diff
@@ -38,13 +38,16 @@
...
+  min-width: 33px;
   height: 33px;
-  font-size: $iconSize;
   cursor: pointer;
   display:flex;
   justify-content:center;
   align-items: center;
+
+  i.icon {
+    font-size: $iconSize;
```
akash1810: only set font size on the icon, not the text

mchv:
💎 💯  one of those little thing that help users!

PR #737, opened 2017-12-20
only allow tracking in workflow if a section has been chosen

akash1810:


akash1810:
[When tracking an atom in workflow, the section must be present but there is no warning/disabled button](https://trello.com/c/9ASrkIME/137-when-tracking-an-atom-in-workflow-the-section-must-be-present-but-there-is-no-warning-disabled-button)

PR #739, opened 2017-12-20
allow hosted atoms to block ads

akash1810:
Labs want to be able to toggle blockAds on Composer pages, so enable the checkbox to allow them to. The `fieldDetails` changes if the atom is a commercial type as we always block ads on the video server side.


akash1810:
[Commercial ad-blocking](https://trello.com/c/8lUID73p/158-commercial-ad-blocking)

akash1810:
tested in CODE

PR #742, opened 2018-01-02
don't send Pluto warning emails in integration tests

akash1810:
Currently, we send an email asking you to set pluto data on an atom if its missing. The email address we use during integration tests doesn't accept incoming mail. This has resulted in a high bounce rate in SES and a temporary suspension...


mbarton:
It may be less brittle to allow sync to be turned off as a flag in the `create` HTTP request itself rather than rely on the integration test user.

Not averse to merging this to get the probation lifted however.

akash1810:
> It may be less brittle to allow sync to be turned off as a flag in the create HTTP request itself rather than rely on the integration test user.

It feels a bit odd to allow this to be enabled/disabled via the client...

PR #746, opened 2018-01-04
allow a note to be sent to Workflow

akash1810:
a note can be set when initially tracking in Workflow and is readonly in MAM once in Workflow

wf-note

requested by Katie, it was quite a quick change and saves time so just did it....

### DIFF
--- a/public/video-ui/styles/components/_forms.scss
+++ b/public/video-ui/styles/components/_forms.scss
@@ -137,6 +137,9 @@
 
   }
 
+  &--textarea {
+    height: inherit;
+  }
 }
 
 .form__label {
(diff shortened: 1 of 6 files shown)


public/video-ui/styles/components/_forms.scss
```diff
@@ -137,6 +137,9 @@
 
   }
 
+  &--textarea {
```
akash1810: we set height to 36px [above](https://github.com/guardian/media-atom-maker/pull/746/files#diff-b745c7e774cc24d5f1eb2621797259efR73)... rather than creating a new class for the new form field, just use css specificity 😄

PR #747, opened 2018-01-05
add GA Autotrack

akash1810:
for cheap analytics of app usage, [autotrack](https://github.com/googleanalytics/autotrack/) should just work out of the box.

Note, I've not added this to the main js bundle as, as noted in the repo:

> Note that it's generally not a good idea to include any analytics as part of your site's main JavaScript bundle since analytics are not usually critical application functionality.

and we're not using code splitting either, so we'd be adding ~8kb to the main js bundle otherwise

### DIFF
--- a/app/views/VideoUIApp/layout.scala.html
+++ b/app/views/VideoUIApp/layout.scala.html
@@ -1,4 +1,10 @@
-@(title: String, isHotReloading: Boolean)(content: Html)
+@(
+  title: String,
+  isHotReloading: Boolean,
+  autotrackJsLocation: String,
+  maybeGaPropertyId: Option[String]
+)(content: Html)
+
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "media-atom-maker",
   "description": "media atom editor",
   "scripts": {
+    "build-autotrack": "autotrack -o public/video-ui/build/autotrack.js -p cleanUrlTracker,eventTracker,impressionTracker,maxScrollTracker,mediaQueryTracker,outboundFormTracker,outboundLinkTracker,pageVisibilityTracker,socialWidgetTracker,urlChangeTracker",
```
akash1810: a custom build - https://github.com/googleanalytics/autotrack/#custom-builds

akash1810:
cc @rcrphillips

app/views/VideoUIApp/layout.scala.html
```diff
@@ -12,6 +18,34 @@
...
+
+          ga('require', 'cleanUrlTracker');
+          ga('require', 'eventTracker');
+          ga('require', 'impressionTracker');
+          ga('require', 'maxScrollTracker');
+          ga('require', 'mediaQueryTracker');
+          ga('require', 'outboundFormTracker');
+          ga('require', 'outboundLinkTracker');
+          ga('require', 'pageVisibilityTracker');
+          ga('require', 'socialWidgetTracker');
```
gtrufitt: Did you need this one!?

PR #748, opened 2018-01-15
Pluto resync endpoint

fredex42:
Adding an endpoint to request manual resync of data to pluto

### DIFF
--- a/app/controllers/PlutoController.scala
+++ b/app/controllers/PlutoController.scala
@@ -62,8 +62,7 @@ class PlutoController(
           pluto.sendToPluto(PlutoResyncMetadataMessage.build(
             versionWithId,
             atom,
-            awsConfig,
-            "system@video.gutools.co.uk"
+            awsConfig
           ))
           Ok(Json.toJson(atom))
         case None=>
--- a/app/controllers/PlutoController.scala
+++ b/app/controllers/PlutoController.scala
@@ -11,7 +11,7 @@ import com.gu.media.util.MediaAtomHelpers
 import data.{DataStores, UnpackedDataStores}
 import play.api.libs.json.Json
 import play.api.mvc.Controller
-import com.gu.media.model.{MediaAtom, MediaAtomBeforeCreation, PlutoSyncMetadataMessage}
+import com.gu.media.model.{MediaAtom, MediaAtomBeforeCreation, PlutoResyncMetadataMessage}
 import com.typesafe.config.{Config, ConfigFactory}
 import util.AWSConfig
 
@@ -59,12 +59,12 @@ class PlutoController(
 
       atom.plutoData match {
         case Some(data)=>
-          pluto.sendToPluto(PlutoSyncMetadataMessage.build(
+          pluto.sendToPluto(PlutoResyncMetadataMessage.build(
             versionWithId,
             atom,
             awsConfig,
             "system@video.gutools.co.uk"
-          ),shouldSendEmail=false)
+          ))
           Ok(Json.toJson(atom))
         case None=>
           BadRequest(s"$id does not have pluto data attached")
--- a/app/controllers/PlutoController.scala
+++ b/app/controllers/PlutoController.scala
@@ -1,19 +1,38 @@
 package controllers
 
+import java.util.Locale
+import com.gu.media.aws._
+import com.gu.media.Settings
+import com.gu.media.logging.Logging
 import com.gu.media.pluto.PlutoUpsertRequest
+import com.gu.media.upload.PlutoUploadActions
 import com.gu.pandahmac.HMACAuthActions
 import data.{DataStores, UnpackedDataStores}
 import play.api.libs.json.Json
 import play.api.mvc.Controller
+import com.gu.media.model.{MediaAtom, MediaAtomBeforeCreation, PlutoSyncMetadataMessage}
+import com.typesafe.config.{Config, ConfigFactory}
 
 class PlutoController(
+  val config:Config,
+  val credentials:AwsCredentials,
   val authActions: HMACAuthActions,
   override val stores: DataStores
 ) extends Controller
   with UnpackedDataStores
-  with JsonRequestParsing {
+  with JsonRequestParsing
+  with Settings
+  with AwsAccess
+  with DynamoAccess
+  with KinesisAccess
+  with SESSettings
+with UploadAccess
+  with Logging {
 
   import authActions.{APIAuthAction, APIHMACAuthAction}
+
+  final override def regionName = sys.env.get("REGION")
+  final override def readTag(tag: String) = sys.env.get(tag.toUpperCase(Locale.ENGLISH))
 
   def getCommissions() = APIAuthAction {
     val plutoCommissions = stores.plutoCommissionStore.list()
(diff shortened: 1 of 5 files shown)


fredex42:
@akash1810 confirmed working for media atom code -> pluto code

app/controllers/PlutoController.scala
```diff
@@ -1,17 +1,29 @@
 package controllers
 
+import java.util.Locale
```
akash1810: think this can be removed as we're not using it

app/controllers/PlutoController.scala
```diff
@@ -31,4 +43,34 @@ class PlutoController(
       Ok(Json.toJson(project))
     }}
   }
+
+  def resendAtomMessage(id: String) = APIHMACAuthAction {
+    val settings = com.gu.media.Settings(this.config)
```
akash1810: we don't seem to be using this, so can be removed

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -79,6 +79,38 @@ object PlutoSyncMetadataMessage {
...
 }
 
+case class PlutoResyncMetadataMessage(
+   `type`: String,
+    enabled: Boolean,
+    projectId: Option[String],
+    s3Key: String,
+    atomId: String,
+    title: String,
+    user: String,
```
akash1810: Do we need this property in this case class? Setting it to `system@video.gutools.co.uk` doesn't seem to have much value as we can switch on `type`. Removing `user` also makes the signature of `build` simpler as `email` can be removed.

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -79,6 +79,38 @@ object PlutoSyncMetadataMessage {
   }
 }
 
+case class PlutoResyncMetadataMessage(
+   `type`: String,
+    enabled: Boolean,
```
akash1810: Don't think we need this property.

app/controllers/PlutoController.scala
```diff
@@ -31,4 +43,34 @@ class PlutoController(
...
+    val pluto = new PlutoUploadActions(awsConfig)
```
akash1810: we can create this variable a bit later within the `atom.plutoData` pattern match

app/controllers/PlutoController.scala
```diff
@@ -31,4 +43,34 @@ class PlutoController(
...
+            atom,
+            awsConfig,
+            "system@video.gutools.co.uk"
+          ))
+          Ok(Json.toJson(atom))
+        case None=>
+          BadRequest(s"$id does not have pluto data attached")
+      }
+    } catch {
+      case excep:Throwable=>InternalServerError(excep.toString)
```
akash1810: could we follow the pattern of other actions in returning `commandExceptionAsResult`, e.g here - https://github.com/guardian/media-atom-maker/blob/master/app/controllers/Api2.scala#L51-L58

app/controllers/PlutoController.scala
```diff
@@ -1,17 +1,29 @@
...
 
 class PlutoController(
+  val config:Config,
+  val awsConfig:AWSConfig,
   val authActions: HMACAuthActions,
   override val stores: DataStores
 ) extends Controller
   with UnpackedDataStores
-  with JsonRequestParsing {
+  with JsonRequestParsing
```
akash1810: do we need to extend with `JsonRequestParsing`?

conf/routes
```diff
@@ -42,6 +42,8 @@ POST    /reindex-publish                com.gu.atom.play.ReindexController.newPu
 GET     /reindex-preview                com.gu.atom.play.ReindexController.previewReindexJobStatus()
 GET     /reindex-publish                com.gu.atom.play.ReindexController.publishedReindexJobStatus()
 
+POST    /pluto/resend/:id               controllers.PlutoController.resendAtomMessage(id)
```
akash1810: can this be `/api2/pluto/resend/:id` to keep [the namespace](https://github.com/guardian/media-atom-maker/blob/master/conf/routes#L17-L19)

app/controllers/PlutoController.scala
fredex42: in a word.... yes.

```
[error] /Users/localhome/workdev/media-atom-maker/app/controllers/PlutoController.scala:39: object parse does not take type parameters.
[error]     parse[PlutoUpsertRequest](req) { data: PlutoUpsertRequest => {
[error]          ^
[error] one error found
[error] (app/compile:compileIncremental) Compilation failed
[info] Compiling 1 Scala source to /Users/localhome/workdev/media-atom-maker/target/scala-2.11/classes...
[error] /Users/localhome/workdev/media-atom-maker/app/controllers/PlutoController.scala:39: object parse does not take type parameters.
[error]     parse[PlutoUpsertRequest](req) { data: PlutoUpsertRequest => {
[error]          ^
[error] one error found
[error] (app/compile:compileIncremental) Compilation failed
```

app/controllers/PlutoController.scala
fredex42: Sure. Out of interest, why does this matter?

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
fredex42: I have kept the properties that are not being used to avoid the need to do any special processing on the other side of the transaction. pluto may be expecting these (it definitely is in the case of userEmail, it uses it to map the user into a user ID for tracking) so if they are not present that may break things downstream.
I am keen not to add any more complexity in processing than necessary, since this is simply a replay of an existing action I don't really want to be having alternate code paths in the handling of the message

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
fredex42: See above, pluto uses this to track which user put the thing in.  I'd rather have an explicit entry, so nothing needs to change on the other side, and that way these are conveniently logged with no code changes

fredex42:
@akash1810 I have updated the code and responded in-line, please let me know

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
akash1810: `enabled` is a legacy feature and should/will be removed as its now controlled by configuration.

> it definitely is in the case of userEmail, it uses it to map the user into a user ID for tracking

The issue here though is we're setting the email to `system@video.gutools.co.uk` which does not relate to an Editorial user...

app/controllers/PlutoController.scala
akash1810: clarity really - there's no point in instantiating it if it won't be used, which is true if there is no `plutoData`

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
fredex42: hmmm, that is true. in that case Pluto just logs the email address.  More importantly, I don't want it falling over because it is expecting to have that key in the sent message

fredex42: tbh, I have checked through the pluto code that handles this and it looks like nothing reads that property.  From what you're saying it should be deprecated on _all_ the messages, not just this one, right?  Would you prefer that in a seperate PR or on the end of this one?

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
fredex42: I've been through the code and it looks like a default value is used at the pluto end if this is not present so it should be safe to remove, i.e. it won't cause a crash.  However it is used in records of what has been imported, for tracking purposes, so `system@video.gutools.co.uk` makes sense at the pluto end as it's obvious from this record what triggered it compared to `unknown user`. I could hard-set it at the pluto end i suppose, but this would mean rejigging some of the logic because right now i can run exactly the same behaviour whether it's a new atom or a forced update.  Seems a bit bloaty to have to add an extra branch of code to fill something extra in at one end while changing the protocol when the same can be achieved just by setting a key at the other end

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
akash1810: related PR https://github.com/guardian/media-atom-maker/pull/749

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
akash1810: Ah ok, yeah probably easier to send it in the message then. Given its just a string, I wonder if `media-atom-pluto-reindex` would be more descriptive and unique for logging/debugging purposes?

fredex42:
confirmed working on dev by the way

PR #749, opened 2018-01-16
remove deprecated field

akash1810:
The `enabled` field has been moved to config and doesn't need to be on the model anymore.

NB the config is pretty painful because of all the traits. I've put:
- `syncWithPluto` in `KinesisAccess` because if false, we don't write to kinesis
- `integrationTestUser` in `SESSettings` because if the user matches the integration test user, we won't send an email


akash1810:
related to #748 and #645

PR #752, opened 2018-01-25
Fix url encoding for capi preview

tomrf1:
api-gateway requires the query param encoding to use `%20`, not `+`
I've updated the library to provide a `IAMEncoder` helper - https://github.com/guardian/content-api-scala-client/pull/250

This hasn't yet been a problem elsewhere because we generally use the correctly-encoded raw query params as they were sent by the client.


mchv:
👍

PR #755, opened 2018-01-26
Add alarms to scheduler and expirer lambdas

Reettaphant:
- Add alarms for when scheduler and expirer lambdas do not run or error.


akash1810:
this is a really nice solution, should we update https://github.com/guardian/recommendations?

PR #757, opened 2018-01-30
Update VideoTrail.js

akash1810:
Editorial feedback suggested S3 was a confusing term as they did not know what it is. This is a fair criticism as its a technical concern rather an a user concern. This new wording should be more approachable.


akash1810:
[WTF is S3 anyway?!](https://trello.com/c/70aXnOh5/127-wtf-is-s3-anyway)

mbarton:
Not to bikeshed about wording but I wonder if we should indicate in the wording that the user should not refresh the page whilst uploading to S3 but can safely do so during YouTube processing?

akash1810:
IIRC we already do this when they navigate away?

PR #758, opened 2018-01-31
Error messages

Reettaphant:
- Use markers to log more structures messages when we update youtube 
- Include the atom id in all the messages
- Display a more descriptive message to the user if atom publishing fails - if it is a known issue such as too many tags, they will have enough information to solve it themselves. If not, they will at least have more information to give us about a problem.

### DIFF
--- a/app/model/YouTubeMessage.scala
+++ b/app/model/YouTubeMessage.scala
@@ -8,7 +8,7 @@ import scala.collection.JavaConverters._
 
 case class YouTubeMessage(atomId: String, videoId: String, reason: String, message: String, isError: Boolean = false) {
 
-  def logMessage() =
+  def logMessage()   =
     if (isError) Logger.logger.error(createMarkers(), "YouTube Video update")
     else Logger.logger.info(createMarkers(), "YouTube Video update")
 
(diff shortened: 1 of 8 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -192,39 +192,36 @@ case class PublishAtomCommand(
...
 
         val blockAds = previewAtom.category match {
           case Category.Hosted | Category.Paid => true
           case _ => previewAtom.blockAds
         }
 
-        youtube.createOrUpdateClaim(previewAtom.id, asset.id, blockAds)
-        previewAtom
+        val claimUpdate = youtube.createOrUpdateClaim(previewAtom.id, asset.id, blockAds)
+        handleYouTubeMessages(claimUpdate, "YouTube Claim Update: creatign a claim", previewAtom, asset.id)
```
mbarton: Typo

app/model/YouTubeMessage.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import net.logstash.logback.marker.Markers
+import org.joda.time.DateTime
+import play.api.Logger
+
+import scala.collection.JavaConverters._
+
+case class YouTubeMessage(atomId: String, videoId: String, reason: String, message: String, isError: Boolean = false) {
+
+  def logMessage =
```
mbarton: Nitpick, I'd define this `def logMessage()` so it looks like a side-effecty function rather than a getter

PR #759, opened 2018-01-31
Add created by

Reettaphant:
- Add created by field on created atoms and remove audit trail button as theory is it will no longer be used.

screen shot 2018-01-31 at 15 23 00

### DIFF
--- a/public/video-ui/src/components/ManagedForm/ManagedSection.js
+++ b/public/video-ui/src/components/ManagedForm/ManagedSection.js
@@ -6,8 +6,8 @@ export class ManagedSection extends React.Component {
     data: PropTypes.object,
     updateData: PropTypes.func,
     children: PropTypes.oneOfType([
-      PropTypes.element,
-      PropTypes.arrayOf(PropTypes.element)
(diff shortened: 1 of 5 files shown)


public/video-ui/src/components/ManagedForm/ManagedSection.js
```diff
@@ -19,13 +19,16 @@ export class ManagedSection extends React.Component {
...
     const hydratedChildren = React.Children.map(this.props.children, child => {
-      return React.cloneElement(child, {
-        data: this.props.data,
-        updateData: this.props.updateData,
-        updateFormErrors: this.props.updateFormErrors,
-        updateWarnings: this.props.updateWarnings,
-        editable: this.props.editable
-      });
+
+      if (child) {
```
akash1810: is this check needed? `child` should be defined within the `.map`?

Reettaphant: If we are creating an atom, then I'm passing in null rather than the created by field so we should just ignore that element.

PR #764, opened 2018-02-05
Targeting UX - first pass

RichieAHB:
This removes the editable title for targeting (which appears to not be used and is confusing) and also add some UI descriptions around what targets are and how they work. Specifically with an aim to be more clear about where the tags are `AND` and where they are `OR` for showing in proposer. Thoughts?!

targeting

### DIFF
--- a/public/video-ui/src/components/Targeting/Targeting.js
+++ b/public/video-ui/src/components/Targeting/Targeting.js
@@ -43,7 +43,7 @@ class Targeting extends React.Component {
       <div>
         {this.props.targetsLoaded && (
           <div>
-            <h3>Targeting rules</h3>
+            <p>A targeting rule will allow this video to be suggested for certain articles.</p>
             {this.props.targets.map((target, index) => (
               <div key={target.id} className="targeting__form">
                 {!isDeleting(target, this.props.deleting) && (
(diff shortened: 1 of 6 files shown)


public/video-ui/src/components/Targeting/Targeting.js
```diff
@@ -44,25 +43,34 @@ class Targeting extends React.Component {
...
+                    data={target}
+                    updateData={this.updateTarget}
+                    editable={true}
+                    formName="TargetingForm"
                   >
-                    <TextInput />
-                  </ManagedField>
-                  {!isDeleting(target, this.props.deleting) && (
+                    <p>
+                      {index === 0 ? (
```
akash1810: lovely...

paperboyo:
👏 
ActiveUntil > Active Until

------------TO IGNORE-------------

Personally, I’m not sure about the wording, but I am Polish, mind you. It may suggest the article is doing the suggesting and not that it’s the one being suggested to.

”Targeting tags” > ”Target**ed** tags” (or ”Tags to target”)
”Search commissioning info (type ‘\*’ to show all)" > "Target these tags (type ’\*’ for wildcard)”?
”atom” > ”video”

There may be some confusion arising from the fact that what you _target_ end up being _suggested_ by a thing called _proposer_, but that’s not really related to this PR. And dropping one (two?!) of these terms... would be serious business.

Plus wording’s bit wordy?

RichieAHB:
@paperboyo completely agree with all of that. I've made some edits to reflect some of your points. I think some of the larger things around what we mean by targeted / targeting will feed into a wider chat around targeting. As I think there are quite a few misconceptions about it!

Also it is a bit wordy but I can't think of a way of being less wordy and I think it probably is clearer than no words at all??

PR #769, opened 2018-02-26
change atom channel on add asset

akash1810:
There have been a couple of times where Adam has wanted to change the channel of an Atom because of incorrect selection but was unable to. In these scenarios, we either curl the update endpint, editing the channel, or create a new atom. Both aren't great.

This PR changes the behaviour of the add asset endpoint allowing you to paste in the URL of any video and if the channel is recognised, the Atom's channel gets updated to match.

### DIFF
--- a/app/model/commands/AddAssetCommand.scala
+++ b/app/model/commands/AddAssetCommand.scala
@@ -74,9 +74,14 @@ case class AddAssetCommand(atomId: String, videoUri: String, override val stores
         }
       }
       case (Some(channel), Some(video)) => {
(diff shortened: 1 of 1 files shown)


Reettaphant:
Is there potential for abuse or for accidents if we do this? Should this be behind permissions?

akash1810:
> Is there potential for abuse or for accidents if we do this? Should this be behind permissions?

Don't think so. You have to paste in a YouTube link for this route to be executed, so there's already a video on YouTube. We check the video is on a Guardian channel too.

PR #770, opened 2018-02-26
allow assets to be deleted

akash1810:
If the asset isn't active, allow it to be deleted.

asset-delete

### DIFF
--- a/public/video-ui/styles/components/_buttons.scss
+++ b/public/video-ui/styles/components/_buttons.scss
@@ -14,23 +14,6 @@ button {
   }
 }
 
-.button__secondary {
-  @include buttonClass($color800Grey, $color500Grey, $color300Grey, $color800Grey);
-
-  &__assets {
-    margin: 5px;
-  }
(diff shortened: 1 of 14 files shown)


mbarton:
I think this would be confusing since it does not also delete the YouTube video (or set it as private)

mbarton:
Also let's make it so it has to be clicked twice like composer deletions. It's precariously close to the Activate button and it would be difficult for a user to re-upload since YouTube would see it as a dupe

akash1810:
Updated with generic `DeleteButton` component.

update

public/video-ui/styles/components/_buttons.scss
```diff
@@ -46,13 +29,33 @@ button {
...
   }
-  
+
   &--list {
     width: 80px;
     padding: 7px;
   }
 }
 
+.button__secondary {
```
akash1810: ahh cascading style... `.button__secondary ` needs to come after `.btn` to win

PR #772, opened 2018-03-26
Project-assigned fields

fredex42:
adding "title" and "user" fields to the "project assigned" message; we have been receiving a lot of these at the pluto end which have not yet been ingested, and cannot start an ingest without these fields

### DIFF
--- a/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
+++ b/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
@@ -16,7 +16,9 @@ case class AtomAssignedProjectMessage(
   `type`: String,
   atomId: String,
   commissionId: String,
-  projectId: String
+  projectId: String,
+  title: String,
+  user: String
 ) extends PlutoIntegrationMessage {
   override def partitionKey: String = atomId
 }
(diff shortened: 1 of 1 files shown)


common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -16,7 +16,9 @@ case class AtomAssignedProjectMessage(
   `type`: String,
   atomId: String,
   commissionId: String,
-  projectId: String
+  projectId: String,
+  title: String,
+  user: String
```
akash1810: Can we make this `Option[String]` rather than sending `unknown_user`? As I understand, `unknown_user` is the default for Pluto at ingest point; sending it here feels like we're mixing domains.

fredex42: @akash1810 sure, done that now

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -26,7 +28,14 @@ object AtomAssignedProjectMessage {
 
   def build(atom: MediaAtom): AtomAssignedProjectMessage = {
     val plutoData = atom.plutoData.get
-    AtomAssignedProjectMessage("project-assigned", atom.id, plutoData.commissionId.get, plutoData.projectId.get)
+
+    val email = for {
```
akash1810: Do we really need a for comprehension here? `atom.contentChangeDetails.created.flatMap(_.user.map(_.email))` should be ok?

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -26,7 +28,14 @@ object AtomAssignedProjectMessage {
...
+      created <- atom.contentChangeDetails.created
+      user <- created.user
+      user_email <- user.email
```
akash1810: camelCase 🐫

fredex42: heh been doing too much python recently

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
fredex42: i was quite pleased with that, normally i can't make the damn things work 😄 thought it would be more readable than the flatmap/map

fredex42:
@akash1810 thanks Akash

PR #773, opened 2018-04-25
Workflow notes

Reettaphant:
I was lazy so I didn't change change the response coming back from workflow. https://trello.com/c/qVWVyE92/285-allow-notes-to-be-edited-for-workflow-in-mam


public/video-ui/src/actions/WorkflowActions/updateWorkflowData.js
```diff
@@ -0,0 +1,37 @@
...
+    message: 'Failed to update Atom data in Workflow',
+    error: error
+  };
+}
+
+export function updateWorkflowData({ workflowItem }) {
+  return dispatch => {
+    dispatch(requestDataUpdate());
+    return Promise.all([
+      WorkflowApi.updateStatus(workflowItem),
```
akash1810: minor: we should update the Workflow api to allow us to make a single request...

Reettaphant: I started looking into this and turns out it is not such a minor thing after all - because of the way workflow parses stubs this would take a bit more effort to accomplish so I'm just going to merge this and make another card for this.

PR #775, opened 2018-05-01
Private block ads

Reettaphant:
When videos are made private (on atom delete, publish or expiry), we should also block adds on these videos. 

https://trello.com/c/iP52mrB0/268-marking-video-as-private-should-also-block-ads

### DIFF
--- a/expirer/src/main/scala/com/gu/media/expirer/ExpirerLambda.scala
+++ b/expirer/src/main/scala/com/gu/media/expirer/ExpirerLambda.scala
@@ -8,7 +8,7 @@ import com.gu.contentatom.thrift.atom.media.PrivacyStatus
 import com.gu.media.CapiAccess
 import com.gu.media.lambda.LambdaBase
 import com.gu.media.logging.Logging
-import com.gu.media.youtube.{YouTubeAccess, YouTubeVideos}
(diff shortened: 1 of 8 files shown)


expirer/src/main/scala/com/gu/media/expirer/ExpirerLambda.scala
```diff
@@ -19,26 +19,34 @@ class ExpirerLambda extends RequestHandler[Unit, Unit]
...
-  with YouTubeVideos {
+  with YouTubeVideos
+  with YouTubeClaims {
 
 
   override def handleRequest(input: Unit, context: Context): Unit = {
     val now = Instant.now()
     val oneDayAgo = Instant.now().minus(1, ChronoUnit.DAYS)
-    val assets = getVideosFromExpiredAtoms(1, 100, oneDayAgo, now, Set.empty).filter(isManagedVideo)
+    val atomsWithAssets: Seq[AssetDetails] = getVideosFromExpiredAtoms(1, 100, oneDayAgo, now, Seq.empty)
```
akash1810: out of scope for this PR: this lambda should really hit and endpoint in MAM rather than do this logic itself

PR #778, opened 2018-05-23
don't add branding to videos on the commercial channels

akash1810:
GLabs have their own logo and it doesn't make sense to add the standard branding to these videos


common/src/main/scala/com/gu/media/model/MediaAtom.scala
```diff
@@ -34,6 +34,10 @@ abstract class MediaAtomBase {
   val optimisedForWeb: Option[Boolean]
   val composerCommentsEnabled: Option[Boolean]
   val suppressRelatedContent: Option[Boolean]
+
+  def isOnCommercialChannel(commercialChannels: Set[String]): Option[Boolean] = {
```
akash1810: opted to put this here rather than in the [helpers](https://github.com/guardian/media-atom-maker/blob/master/common/src/main/scala/com/gu/media/util/MediaAtomHelpers.scala) because I think the helpers should be methods on the companion objects rather than a standalone object (will refactor them soon).

PR #779, opened 2018-05-23
move cfn to platform repo

akash1810:
see https://github.com/guardian/editorial-tools-platform/pull/186


akash1810:
cc @guardian/digital-cms

PR #782, opened 2018-05-25
set the correct image write format for ImageIO

akash1810:
Previously, we were telling ImageIO to always write as `jpg`. However, this would fail if the image from Grid was a `png`.

Now, we use the mime-type in the Grid response.


akash1810:
Tested with [this image](https://media.test.dev-gutools.co.uk/images/d7f567627bfc5b25d3190e61ccd85c126f5f557f) that was failing in PROD. It now works!

PR #787, opened 2018-06-12
add delete endpoint for pluto commission deletion

akash1810:
deleting a commission will first delete all the projects associated with it

### DIFF
--- a/app/controllers/PlutoController.scala
+++ b/app/controllers/PlutoController.scala
@@ -37,6 +37,12 @@ class PlutoController(
     Ok(Json.toJson(plutoProjects))
   }
 
+  def deleteCommission(id: String) = APIHMACAuthAction {
+    stores.plutoProjectStore.deleteByCommissionId(id)
+    stores.plutoCommissionStore.delete(id)
+    Accepted
+  }
+
   def upsertProject() = APIHMACAuthAction { implicit req =>
     parse[PlutoUpsertRequest](req) { data: PlutoUpsertRequest => {
       val project = stores.plutoProjectStore.upsert(data)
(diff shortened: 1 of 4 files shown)


akash1810:
cc @fredex42 @yusufparkar

app/controllers/PlutoController.scala
```diff
@@ -37,6 +37,12 @@ class PlutoController(
     Ok(Json.toJson(plutoProjects))
   }
 
+  def deleteCommission(id: String) = APIHMACAuthAction {
+    stores.plutoProjectStore.deleteByCommissionId(id)
```
mbarton: Should we check the return result of these two to catch errors?

common/src/main/scala/com/gu/media/pluto/PlutoProjectDataStore.scala
```diff
@@ -43,4 +43,9 @@ class PlutoProjectDataStore(aws: DynamoAccess, plutoCommissionDataStore: PlutoCo
     Scanamo.exec(aws.dynamoDB)(op)
     project
   }
+
+  def deleteByCommissionId(commissionId: String) = {
+    log.info(s"deleting all pluto projects for commission $commissionId")
+    getByCommissionId(commissionId).map(project => Scanamo.delete(aws.dynamoDB)(table.name)('id -> project.id))
```
mbarton: as above, do we know if one of the deletes failed?

app/controllers/PlutoController.scala
akash1810: `Scanamo.delete` returns a `DeleteItemResult`, which, looking at [the docs](https://docs.aws.amazon.com/sdkfornet1/latest/apidocs/html/T_Amazon_DynamoDB_Model_DeleteItemResult.htm), returns on success or failure. I don't think we necessarily care if we try to delete a non-existing commission...

mbarton: Who re-sends the message?

akash1810: Pluto would.

mbarton: OK cool, so it checks to see if the deletion has actually worked and retries accordingly?

PR #788, opened 2018-06-12
update pluto message consumer to delete a commission when necessary

akash1810:
follows #787 

Pluto sends a message on kinesis with a commission title of `(DELETE)` for items that should be deleted. Lets obey that message!


akash1810:
cc @fredex42 @yusufparkar

akash1810:
fixes https://trello.com/c/7DzwGgPR/2651-atom-quick-fixes

PR #790, opened 2018-06-18
refactor for better organisation

akash1810:
push helper logic onto the case class for better organisation


LATaylor-guardian:
👍

PR #792, opened 2018-06-18
move calculation of youtube settings to UpdateAtomCommand

akash1810:
Move the calculation of `blockAds` and `privacyStatus` to `UpdateAtomCommand` so that they are calculated on save rather than on publish.

This'll make for a friendlier UI as the values will update at a more realistic time.


franziskas:
@akash1810 time to salvage or close this?

PR #793, opened 2018-06-20
send an email when a new YouTube video is uploaded with WC tags

akash1810:
We've got an ad campaign for the WC which will get served when a video has the tag `world cup 2018` or `2018 world cup`.

There have been a couple of instances where the YouTube video didn't make it to DfP for targeting. We're not sure why this is as the video appears to be tagged correctly on youtube.com.

Tom suggested sending him an email when a new video is uploaded to YouTube with these tags so he can then check DfP immediately rather than finding DfP is missing items when its too late.

Note: the email gets sent when an atom is published with a new YouTube asset, or the tags have been added to the existing asset.

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -8,6 +8,7 @@ import com.gu.contentatom.thrift.{ContentAtomEvent, EventType}
 import com.gu.media.logging.Logging
 import com.gu.media.model.Platform.Youtube
 import com.gu.media.model.{AuditMessage, _}
+import com.gu.media.ses.Mailer
 import com.gu.media.youtube.YouTubeMetadataUpdate
 import com.gu.media.{Capi, MediaAtomMakerPermissionsProvider}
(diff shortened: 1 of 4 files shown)


app/model/commands/PublishAtomCommand.scala
```diff
@@ -143,10 +146,23 @@ case class PublishAtomCommand(
...
     publishedAtom
   }
 
+  private def sendWorldCupNotification(previewAtom: MediaAtom, maybePublishedAtom: Option[MediaAtom]) = {
+    val previewAtomHasTags = previewAtom.tags.contains("2018 world cup") || previewAtom.tags.contains("world cup 2018")
+
+    if (previewAtomHasTags) {
+      maybePublishedAtom match {
+        case Some(publishedAtom) => {
+          val hasNewAssets = hasNewAssets(previewAtom, publishedAtom)
```
marialivia16: This won't compile, I believe, because it has the same name as the function. Also, it doesn't seem to be used?

akash1810: oops 😊! Updated!

PR #796, opened 2018-06-20
render drop down as a dropdown

akash1810:
Currently, there is some confusion around this button (some reports of not daring to click on it!). I think this is due to the fact it looks like a button rather than a drop down.

This is a quick attempt at making it more obvious. Design inspired by Composer's similar component.

# Before
image

# After
image


paperboyo:
> inspired by

Always a good idea. A dropdown looking like a button – not so much, but both tools are perpetrators here.

I would add a tooltip too!

akash1810:
> I would add a tooltip too!

👍 

image

akash1810:
merging w/out explicit approval as its a fairly innocent styling change...

PR #797, opened 2018-06-21
save and launch

akash1810:
In an effort to make "publish" more understandable, change this button to "save and launch" if there are unpublished changes. This is similar behaviour to Composer.

image


akash1810:
[Publish button on MAM misleading publishers](https://trello.com/c/YTY6pCR0/346-publish-button-on-mam-misleading-publishers)

PR #799, opened 2018-07-02
log revision number when logging successful publish

akash1810:
We've found some atoms where updates seemingly fail to reach capi. Log the revision number to make make debugging a bit easier.


LATaylor-guardian:
👍

PR #803, opened 2018-08-16
open Grid on preview page

akash1810:
rather than cropping page


akash1810:
Reverting this as its allowing the thumbnail to be cropped at 16:9 which is incorrect. The thumbnail should always be 5:3.

This is a problem within Grid, but lets revert here to be safe and then revert back once Grid is fixed.

PR #806, opened 2018-08-28
Reduce the number of calls to the YouTube Partner API when making a claim

akash1810:
YouTube's Partner API is rate limited on the number of calls that can be made in a day and also in a second. We've been seeing some API calls being rejected because `userRateLimitExceeded`.

This changes the way we find a previous claim on a video by calling `setOnBehalfOfContentOwner`, `setIncludeThirdPartyClaims` and `setPartnerUploaded` to only return first party claims. This should return 0 or 1 items. This significantly reduces the number of API calls we're doing - previously, we were getting *all* claims and then filtering in scala. This should result is us hitting `userRateLimitExceeded` less often.

See https://developers.google.com/youtube/partner/docs/v1/claimSearch/list for more information.

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
@@ -144,38 +144,29 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
     }
   }
 
-  private def getPagedVideoClaims(request: YouTubePartner#ClaimSearch#List, items: List[ClaimSnippet]):
-    List[ClaimSnippet] = {
+  private def getPartnerClaim(videoId: String): Option[ClaimSnippet] = {
+    val response = partnerClient
+      .claimSearch()
+      .list
+      .setVideoId(videoId)
+      .setOnBehalfOfContentOwner(contentOwner)
+      .setIncludeThirdPartyClaims(false)
+      .setPartnerUploaded(true)
+      .execute()
 
-    val response = request.execute()
-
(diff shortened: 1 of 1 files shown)


common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
```diff
@@ -144,38 +144,29 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
...
+  }
 
+  def createOrUpdateClaim(atomId: String, videoId: String, blockAds: Boolean): Either[VideoUpdateError, String] = {
     try {
-
-      val videoClaims = getPagedVideoClaims(request, List())
-
-      videoClaims.find(claimSnippet => !claimSnippet.getThirdPartyClaim) match {
+      getPartnerClaim(videoId) match {
         case Some(claimSnippet) => {
```
franziskas: Limited Scala knowledge alert: can this be done by directly mapping on the result of type Option?

akash1810: we're doing that here:

```scala
getPartnerClaim(videoId) match { //getPartnerClaim(videoId) is type Option[ClaimSnippet]
  case Some(claimSnippet) => ...
  case None => ...
}
```

unless I've misunderstood the question?

franziskas: got it, there's a different path on none :)

akash1810: discussed IRL: we want to do something on the unhappy path, so a `map` wouldn't work

PR #818, opened 2018-11-22
Handle YouTube failure when adding asset to atom that already has a channel

mbarton:
Making the workaround introduced in #817 work with existing atoms that already have channels (because they were set when the YouTube API was available).


mbarton:
Reviewed and approved by @Reettaphant and @jonathonherbert IRL

PR #820, opened 2019-01-25
introduce Livestream video category

akash1810:
this is to provide a hint to rendering clients that the atom is different from a normal video, so they can, for example, add a flashing red light next to the video


akash1810:
cc @gtrufitt

PR #821, opened 2019-01-27
accept more youtube url formats

akash1810:
YouTube URLs can come in many different shapes. Provide a kinder experience to user by accepting more formats other than the standard one.

This current limitation in the app has been reported a few times in the past and the error messaging we provide is pretty obscure [`Could not create asset. Asset is not a youtube video`](https://github.com/guardian/media-atom-maker/blob/11fcc50e721c6c4d76bb7fe01725342b90a5d42f/app/model/commands/AddAssetCommand.scala#L41).

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YoutubeUrl.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YoutubeUrl.scala
@@ -1,0 +1,14 @@
+package com.gu.media.youtube
+
+object YoutubeUrl {
+  private val re = "^https://(?:www.youtube.com/(?:watch\\?v=|embed/)|youtu.be/)([a-zA-Z0-9_-]{11}).*$".r
+
+  def unapply(url: String): Option[String] = parse(url)
+
+  def parse(url: String): Option[String] = {
+    url match {
+      case re(videoId) => Some(videoId)
+      case _ => None
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


akash1810:
[Media Atom Maker only accepts URLs in the form of https://www.youtube.com/watch?v=VIDEO_ID e.g. https://www.youtube.com/watch?v=PB9g6LJQLak.](https://trello.com/c/XahKJ0CG/622-media-atom-maker-only-accepts-urls-in-the-form-of-https-wwwyoutubecom-watchvvideoid-eg-https-wwwyoutubecom-watchvpb9g6ljqlak)

common/src/main/scala/com/gu/media/youtube/YoutubeUrl.scala
```diff
@@ -0,0 +1,14 @@
+package com.gu.media.youtube
+
+object YoutubeUrl {
+  private val re = "^https://(?:www.youtube.com/(?:watch\\?v=|embed/)|youtu.be/)([a-zA-Z0-9_-]{11}).*$".r
```
mbarton: `{11}` - nitpicking but are we sure YouTube will not extend their ids?

akash1810: I'm fairly confident it would change. Its been 11 chars long for as long as I've used the APIs (about 8 years).

akash1810:
Parsing the URL would possibly be neater, however more LOC I think. This is very much like our client-side pairing test lol.

I'll ship this for now.

PR #824, opened 2019-02-01
Refactor build.sbt to use inline dependencies 

Reettaphant:
Refactor build.sbt to use inline dependencies so that snyk can understand them and identify vulnerabilities correctly.


build.sbt
```diff
@@ -24,15 +70,81 @@ lazy val common = (project in file("common"))
...
   )
 
 lazy val app = (project in file("."))
   .dependsOn(common)
   .enablePlugins(PlayScala, SbtWeb, BuildInfoPlugin, JDebPackaging)
   .settings(commonSettings,
     name := "media-atom-maker",
-    libraryDependencies ++= Dependencies.appDependencies,
+    libraryDependencies ++= Seq(
+      "com.gu" %% "pan-domain-auth-play_2-5" % pandaVersion,
```
mbarton: Where we already have the dependency in `common` we don't need to re-state it here

franziskas:
@Reettaphant @mbarton is it worth revisiting this PR?

Reettaphant:
@franziskas @mbarton Yes, worth revisiting, I got the tests passing yesterday and this is now ready to go.

akash1810:
This seems to have broken the integration tests.

PR #826, opened 2019-02-05
show last modified by field

akash1810:
this is to make video producer's life a bit easier as they don't have to send group emails out

follows the pattern of the pre-existing created by field

image


RichieAHB:
👍

PR #832, opened 2019-03-12
Check for publication status before setting embargo or schedule dates

Reettaphant:
It is currently possible currently to set an embargo/schedule launch date on a media atom by editing the atom in two different tabs at the same time. We check for embargoes/scheduled launch on publish but not the published status when scheduling/embargoing atoms. This will result in an embargo/schedule date on the video page which is impossible to remove in the atom tool because the atom has already been published. 

After this fix we can still incorrectly set the embargo date on the video page because we send the data to composer from the client and we shouldn't do this but fixing this is out of the scope of this pr. This fix will still improve the situation because we can push the correct version of the video atom to composer by refreshing the video page and hitting publish because the inconsistent atom state never made it to dynamo.

### DIFF
--- a/app/model/commands/UpdateAtomCommand.scala
+++ b/app/model/commands/UpdateAtomCommand.scala
@@ -9,6 +9,7 @@ import com.gu.media.logging.Logging
 import data.DataStores
 import model.commands.CommandExceptions._
 import com.gu.media.model.{AtomAssignedProjectMessage, ChangeRecord, MediaAtom, AuditMessage}
+import com.gu.contentatom.thrift.{ChangeRecord => ThriftChangeRecord}
 import com.gu.media.upload.PlutoUploadActions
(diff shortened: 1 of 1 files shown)


app/model/commands/UpdateAtomCommand.scala
```diff
@@ -36,9 +36,18 @@ case class UpdateAtomCommand(id: String, atom: MediaAtom, override val stores: D
     log.info(s"Update atom changes ${atom.id}: $diffString")
 
     val changeRecord = ChangeRecord.now(user)
+    val atomIsPublished = existingAtom.contentChangeDetails.published
```
akash1810: could this be a boolean to make the lines below a bit simpler?

Reettaphant: I've abstracted it out to a separate function

PR #836, opened 2019-03-15
creates a content change detail component

akash1810:
smaller components make things easier to reason about

### DIFF
--- a/public/video-ui/src/components/ContentChangeDetails/index.js
+++ b/public/video-ui/src/components/ContentChangeDetails/index.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ManagedField, ManagedForm } from '../ManagedForm';
+import TextInput from '../FormFields/TextInput';
+
+class ContentChangeDetails extends React.Component {
+  static propTypes = {
+    video: PropTypes.object.isRequired
+  };
+
+  getCreatedByField = () => {
+    return (
+      <ManagedField
+        fieldLocation="contentChangeDetails.created.user.email"
+        name="Created by"
+        disabled={true}
+      >
+        <TextInput />
+      </ManagedField>
+    );
+  };
+
+  getModifiedByField = () => {
+    return (
+      <ManagedField
+        fieldLocation="contentChangeDetails.lastModified.user.email"
+        name="Last modified by"
+        disabled={true}
+      >
+        <TextInput />
+      </ManagedField>
+    );
+  };
+
+  render() {
+    const { video } = this.props;
+
+    if (!video || !video.id) {
+      return null;
+    }
+
+    return (
+      <ManagedForm data={video}>
+        {this.getCreatedByField()}
+        {this.getModifiedByField()}
+      </ManagedForm>
+    );
+  }
+}
+
+export default ContentChangeDetails;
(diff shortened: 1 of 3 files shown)


public/video-ui/src/components/ContentChangeDetails/index.js
```diff
@@ -0,0 +1,51 @@
...
+  render() {
+    const { video } = this.props;
+
+    if (!video || !video.id) {
+      return null;
+    }
+
+    return (
+      <ManagedForm>
+        {this.getCreatedByField(video)}
```
mbarton: The functions don't have any parameters in their definitions?

public/video-ui/src/components/ContentChangeDetails/index.js
```diff
@@ -0,0 +1,51 @@
...
+      >
+        <TextInput />
+      </ManagedField>
+    );
+  };
+
+  render() {
+    const { video } = this.props;
+
+    if (!video || !video.id) {
```
mbarton: `video` is marked as required in the props. Would we ever get a video without an id?

public/video-ui/src/components/ManagedForm/ManagedForm.js
```diff
@@ -55,7 +57,7 @@ export class ManagedForm extends React.Component {
         ManagedForm.managedTypes.indexOf(child.type) > -1
           ? React.cloneElement(child, {
               data: this.props.data,
-              updateData: this.props.updateData,
+              updateData: this.props.updateData || this.emptyUpdateData,
```
mbarton: What's the reason for the change here?

akash1810: there isn't one!

public/video-ui/src/components/ContentChangeDetails/index.js
akash1810: ah! great spot, thanks!

public/video-ui/src/components/ContentChangeDetails/index.js
akash1810: This is for atom creation, as the component is shared. An incoming PR will tidy this up a bit, pushing this further up the tree for a cleaner and clearer design.

public/video-ui/src/components/ManagedForm/ManagedForm.js
akash1810: removed.

public/video-ui/src/components/ContentChangeDetails/index.js
akash1810: fixed.

public/video-ui/src/components/ManagedForm/ManagedForm.js
```diff
@@ -6,7 +6,7 @@ import { ManagedSection } from './ManagedSection';
 export class ManagedForm extends React.Component {
   static propTypes = {
     data: PropTypes.object,
-    updateData: PropTypes.func.isRequired,
+    updateData: PropTypes.func,
```
mbarton: Is this still left over?

akash1810: Should have added a comment here, sorry.

This is a necessary change. We're using `ManagedForm` for the CCD component, however the CCD component is completely read-only, so doesn't need to specify an `updateData` function... even if it did, it would be `() => undefined` which is pointless. Much nicer to not make `updateData` required.

akash1810: Obviously we could not use `ManagedForm` for this component, but I cba to work out the styling atm 😄 .

PR #843, opened 2019-03-18
splitting furniture fields across tabs

akash1810:
https://github.com/guardian/media-atom-maker/pull/839 take 2.

Depends on #838 and #841 (those commits have been cherry picked onto this branch).

We want to be able to edit the YouTube title and description independently of the equivalent onsite metadata. This means we're introducing two new fields to the already busy form.

Switch to a tabbed interface to provide more clarity to the UI. This has been discussed with UX and a couple of video producers.

tabby

PR #844, opened 2019-03-18
move workflow to tab

akash1810:
depends on https://github.com/guardian/media-atom-maker/pull/843

image

### DIFF
--- a/public/video-ui/src/components/Workflow/Workflow.js
+++ b/public/video-ui/src/components/Workflow/Workflow.js
@@ -1,12 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import WorkflowForm from './WorkflowForm';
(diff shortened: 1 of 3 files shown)


public/video-ui/src/components/YoutubeFurniture/index.js
```diff
@@ -0,0 +1,149 @@
...
+      video,
+      editable,
+      updateVideo,
+      updateErrors,
+      updateWarnings,
+    } = this.props;
+
+    const { categories } = this.props.youtube;
+
+    const isYoutubeAtom = VideoUtils.isYoutube(video);
```
jonathonherbert: Are these calls, e.g. getAvailableChannels, triggering side-effects? Are we sure we want to make them every render?

PR #848, opened 2019-03-19
introduce youtubeTitle and youtubeDescription

akash1810:
Depends on #847 

We need to be able to edit the video title on youtube independently of the headline used on site, as what performs on theguardian.com doesn't necessarily perform on youtube.com.

Currently, we're achieving this by Adam editing the youtube metadata in the youtube cms after atom publish. This quickly gets annoying as any atom publish overrides the information on youtube.com.

yt

PR #855, opened 2019-03-26
remove html tags before using atom description as yt description

akash1810:
the atom's description is html, remove any html markup before using it as the youtube description as youtube displays as plain text

This was removed [here](https://github.com/guardian/media-atom-maker/pull/848/files#diff-7f574d10ef01ba109a3d2e022bea685cL239) and I clearly hadn't tested that branch with rich text! Moving it into `common` (and thus jsoup is a common dependency) so that it can be part of the model, rather calculated at publish time.

an example of the wrong behaviour:
image


common/src/main/scala/com/gu/media/model/MediaAtom.scala
```diff
@@ -227,15 +228,25 @@ case class MediaAtom(
 object MediaAtom extends MediaAtomImplicits {
   implicit val mediaAtomFormat = Jsonx.formatCaseClass[MediaAtom]
 
+  private def removeHtmlTagsForYouTube(description: String): String = {
```
akash1810: lifted from [here](https://github.com/guardian/media-atom-maker/pull/848/files#diff-7f574d10ef01ba109a3d2e022bea685cL239)

PR #857, opened 2019-03-26
comma separate tags

akash1810:
Its sometimes desired to copy and paste tags (in particular youtube tags) across atoms. We already accept a comma seperated string as input for youtube tags, this changes the rendering to include commas between tags to make copy pasting easier. Previously, we were copying the tags, pasting into a text editor, adding the commas, pasting into atom... long!

Also removes the underlining style as it was quite distracting.

# Before
image

# After
image


akash1810:
This is breaking the byline field - we should render `' '` only for the byline separator.

PR #858, opened 2019-03-26
dedupe youtube tags

akash1810:
YouTube doesn't treat `Foo` any differently from `foo` in terms of search/seo etc. We should dedupe input into the youtube tags field.

### DIFF
--- a/public/video-ui/src/components/FormFields/TextInputTagPicker.js
+++ b/public/video-ui/src/components/FormFields/TextInputTagPicker.js
@@ -5,7 +5,6 @@ import UserActions from '../../constants/UserActions';
 import TagTypes from '../../constants/TagTypes';
 import CapiSearch from '../CapiSearch/CapiSearch';
 import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
-import removeTagDuplicates from '../../util/removeTagDuplicates';
 
 export default class TextInputTagPicker extends React.Component {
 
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/FormFields/TextInputTagPicker.js
```diff
@@ -5,7 +5,6 @@ import UserActions from '../../constants/UserActions';
 import TagTypes from '../../constants/TagTypes';
 import CapiSearch from '../CapiSearch/CapiSearch';
 import removeStringTagDuplicates from '../../util/removeStringTagDuplicates';
-import removeTagDuplicates from '../../util/removeTagDuplicates';
```
akash1810: oh, intellij removed this as it wasn't being used... sounds similar to what we're now doing below 🤷‍♂️

public/video-ui/src/components/FormFields/TextInputTagPicker.js
```diff
@@ -40,33 +38,30 @@ export default class TextInputTagPicker extends React.Component {
...
+      .reduce((acc, candidate) => {
+        const lowerCaseCandidate = candidate.toLowerCase();
+        const isNewValue = !existingValueSet.has(lowerCaseCandidate) && !newValueSet.has(lowerCaseCandidate);
+
+        if (isNewValue) {
+          acc.push({ id: candidate, webTitle: candidate});
         }
 
-        return keywordsAsObjects;
+        newValueSet.add(candidate);
```
Reettaphant: What is `newValueSet` being used for and why are we updating it from inside the reduce?

akash1810: `newValueSet` represents the new tags being added to the field. This field accepts a comma separated string, so we add to `newValueSet` to ensure input of `Foo, bar, baz, foo` results in `Foo, bar, baz` as `newValueSet` will be `Set('foo', 'bar', 'baz')` and the second `foo` is ignored. Its not especially obvious, I agree.

PR #865, opened 2019-03-28
Check for empty description

Reettaphant:
To avoid sentry errors

### DIFF
--- a/public/video-ui/src/components/YoutubeFurniture/index.js
+++ b/public/video-ui/src/components/YoutubeFurniture/index.js
@@ -37,7 +37,7 @@ class YoutubeFurniture extends React.Component {
   }
 
   validateYouTubeDescription = description => {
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/YoutubeFurniture/index.js
```diff
@@ -37,7 +37,7 @@ class YoutubeFurniture extends React.Component {
   }
 
   validateYouTubeDescription = description => {
-    return description.match(/<|>/)
+    return description && description.match(/<|>/)
```
franziskas: would the error message be correct in the case of an empty description?

Reettaphant: Yes, if there is an empty description we don't return an error and don't get an error message which is the desired behaviour.

PR #866, opened 2019-03-29
clean youtube description at atom creation time

akash1810:
On creation, we set the YouTube description using the standfirst. The standfirst can be HTML, however the YouTube description cannot be and so we need to clean it first.

Extract out the HTML cleaning so it can be used at create and update time and also include more tests.


common/src/main/scala/com/gu/media/model/MediaAtom.scala
```diff
@@ -123,7 +123,7 @@ case class MediaAtomBeforeCreation(
 
   // when creating an atom, use the `title` (headline) and `description` (standfirst) as the initial `youtubeTitle` and `youtubeDescription`
   override val youtubeTitle: String = title
-  override val youtubeDescription: Option[String] = description
+  override val youtubeDescription: Option[String] = YoutubeDescription.clean(description)
```
akash1810: this is the main change: clean `description` (aka standfirst) before setting it as `youtubeDescription` on atom creation

akash1810:
Sanity tested on CODE now too.

PR #867, opened 2019-04-01
Remove embargo dates

marialivia16:
Because the scheduled lambda doesn't remove embargo dates once they've expired, we've decided to temporarily remove the embargo dates from the ui. That means users will only be able to prevent publication indefinitely.


marialivia16:
@akash1810 Reetta added a trello card for this, I think that'll do.

PR #868, opened 2019-04-04
show more content change details

akash1810:
Specifically dates and publish information (if available).

This is to help diagnose issues.

image

### DIFF
--- a/public/video-ui/styles/components/_forms.scss
+++ b/public/video-ui/styles/components/_forms.scss
@@ -7,7 +7,6 @@
 }
 
 .form__section {
-  padding: 20px;
   width: 50%;
 }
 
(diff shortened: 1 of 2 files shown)


paperboyo:
Great! Couldn’t we revive full audit here or am I dreaming that it was there but removed for lack of space (which now came back)?

Irregardless, wouldn’t it be more readable like:
```Created at
19 Mar 2019 09:09 by akash.askoolum@guardian.co.uk```
etc.? Or am I reading it wrong?

@harpreetpurewal @akemitakagi

**Comment by akash1810** [comment]
> Couldn’t we revive full audit here or am I dreaming that it was there

Audit was [removed](https://github.com/guardian/media-atom-maker/pull/767) a while ago as part of the GDPR work.

> wouldn’t it be more readable like [...]

😍 Yes! However I'm fundamentally lazy and this was the fastest way to get something out whilst reusing existing code and components.

Ideally, we'd take inspiration from Composer's management tab wrt formatting:

image

File: public/video-ui/styles/components/_forms.scss [hunk 1]
```diff
@@ -7,7 +7,6 @@
 }
 
 .form__section {
-  padding: 20px;
```
mbarton: Seems a bit general?

akash1810: nothing else is using it

PR #869, opened 2019-04-15
Send CAPI notifications as well as updates

mbarton:
This will allow us to run pubflow against video atoms (https://github.com/guardian/pubflow/pull/193)

### DIFF
--- a/app/util/NotifyingAtomPublisher.scala
+++ b/app/util/NotifyingAtomPublisher.scala
@@ -1,0 +1,48 @@
+package util
+
+import com.amazonaws.services.sns.AmazonSNS
+import com.amazonaws.services.sns.model.PublishRequest
+import com.gu.atom.publish.AtomPublisher
+import com.gu.contentatom.thrift.{ContentAtomEvent, EventType}
+import org.joda.time.{DateTime, DateTimeZone}
+import play.api.libs.json.{Json, Writes}
+
+import scala.util.Try
+
+class NotifyingAtomPublisher(isLive: Boolean, topicArn: String, underlying: AtomPublisher, sns: AmazonSNS) extends AtomPublisher {
+  override def publishAtomEvent(event: ContentAtomEvent): Try[Unit] = {
+    underlying.publishAtomEvent(event).map { _ =>
+      val notification = SimpleContentUpdate.fromEvent(event, isLive)
+      val json = Json.stringify(Json.toJson(notification))
+
+      // TODO MRB: should we have a new message-type called atom update?
+      val request = new PublishRequest(topicArn, json, "composer-update")
+      Try(sns.publish(request))
+    }
+  }
+}
+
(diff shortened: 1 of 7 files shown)


app/util/NotifyingAtomPublisher.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.gu.atom.publish.AtomPublisher
+import com.gu.contentatom.thrift.{ContentAtomEvent, EventType}
+import org.joda.time.{DateTime, DateTimeZone}
+import play.api.libs.json.{Json, Writes}
+
+import scala.util.Try
+
+class NotifyingAtomPublisher(isLive: Boolean, topicArn: String, underlying: AtomPublisher, sns: AmazonSNS) extends AtomPublisher {
+  override def publishAtomEvent(event: ContentAtomEvent): Try[Unit] = {
+    underlying.publishAtomEvent(event).map { _ =>
```
regiskuckaertz: `flatMap`? (since `sns.publish` may fail)

app/util/NotifyingAtomPublisher.scala
```diff
@@ -0,0 +1,48 @@
...
+      val notification = SimpleContentUpdate.fromEvent(event, isLive)
+      val json = Json.stringify(Json.toJson(notification))
+
+      // TODO MRB: should we have a new message-type called atom update?
```
regiskuckaertz: It can't hurt 😅 though at this moment it's redundant with the topic. I suppose the question is if instead of two topic+queue there could be only one?

app/util/NotifyingAtomPublisher.scala
```diff
@@ -0,0 +1,48 @@
...
+                                isMigration: Boolean = false,
+                                isProdmon: Boolean,
+                                isExpired: Boolean
+                              )
+
+object SimpleContentUpdate {
+  implicit val writes: Writes[SimpleContentUpdate] = Json.writes[SimpleContentUpdate]
+
+  def fromEvent(event: ContentAtomEvent, isLive: Boolean): SimpleContentUpdate = SimpleContentUpdate(
+    composerId = s"${event.atom.atomType.toString}-${event.atom.id}",
```
regiskuckaertz: can we align with CAPI and do `atomType/atomId`?

app/util/NotifyingAtomPublisher.scala
mbarton: For the price of an extra SQS I think it's best to keep them separated for now whilst we experiment. Doesn't look like we filter on subject but good to know the feature is there

mbarton: I think I will change the serialisation reader in Pubflow to allow us to use `id` as well as `composerId` in `SimpleContentUpdate` just to avoid it getting confusing

regiskuckaertz: yes, makes sense.

PR #870, opened 2019-04-29
sticky toolbar

akash1810:
# Before 😢 
image

# After 🎉 
image


akash1810:
going rogue and merging w/out a 👍

PR #880, opened 2019-07-18
Rewrite Youtube embed URLs to use the cookie-less domain

regiskuckaertz:
This is going to supersede [this PR](https://github.com/guardian/content-api/pull/2285) in CAPI as MAM is the source of truth.


regiskuckaertz:
Question is: if we launch a reindex, will all media atoms be generated from scratch and use that new URL?

akash1810:
> Question is: if we launch a reindex, will all media atoms be generated from scratch and use that new URL?

Sadly the `defaultHtml` is saved in dynamo. That is it's not dynamically generated and we'd need to update records.

regiskuckaertz:
Concierge has been doing the rewrite on the fly for a few days now, and nobody complained, so we can proceed with this one. Can I just merge @akash1810 ? Or is there a preliminary step to take first with the DB?

regiskuckaertz:
ping 🙏

akash1810:
Let's merge!

PR #881, opened 2019-08-16
use an api key to talk to CAPI

akash1810:
migrating off the internal capi endpoint as it looks to be removed from DNS


akash1810:
cc @guardian/digital-cms

akash1810:
Updated config on CODE and PROD and tested on CODE.

PR #882, opened 2019-08-16
on first publish, set YouTube title and description to that of the Atom

akash1810:
This is because there's no guarantee that the YouTube furniture gets subbed before publication and can result in draft furniture being used.

When we create an atom, we set the youtube furniture to the same as the Atom's. This is because the `YoutubeTitle` is not optional. We've seen that after initial creation, the YT furniture doesn't get looked at 😢 and so if we create an Atom with title [`Wonder goal`](https://video.gutools.co.uk/videos/d9ace151-0496-4fa0-acda-150e4ebff29d), that's what gets sent to YouTube.

Extends the behaviour of https://github.com/guardian/media-atom-maker/pull/848


akash1810:
tested on CODE

PR #883, opened 2019-08-27
FIX: make prodoffice field editable in Workflow form 

ajwl:
The field `Production Office` should be editable in Media Atom Maker, just like it is in Composer. This  adds a select form with 3 hardcoded values - US, UK, AU.

These values are hardcoded - like they are in Composer - because it is unlikely a new production office will be setup. 

1) The first part of this code hardcodes the above values and passes them down to `WorkflowForm` to be displayed on a select menu. The select menu uses the existing function called `getProductionOffice to guess the user's production office from their timezone and set it as the default in the select menu. 
It is  required to  add: 'productionOffice' before you can save the workflow form.  

2) The second part makes sure that the POST request sent when a workflow form is saved (for the first time) includes the prodOffice property taken from the form.

3) Thirdly, adds a PUT request that just updates the ProdOffice property on the Workflowform, if that's the only bit that's been updated. 

Deployed on CODE here: https://video.code.dev-gutools.co.uk

### DIFF
--- a/public/video-ui/src/components/Workflow/Workflow.js
+++ b/public/video-ui/src/components/Workflow/Workflow.js
@@ -48,6 +48,11 @@ class Workflow extends React.Component {
         workflowStatuses={this.props.workflow.statuses || []}
         workflowPriorities={this.props.workflow.priorities}
         workflowStatus={this.props.workflow.status}
+        workflowProductionOffices={[
+          { id: 'UK', title: 'UK' },
+          { id: 'US', title: 'US' },
+          { id: 'AU', title: 'AU' }
+        ]}
         updateData={this.updateLocalData}
       />
     );
--- a/public/video-ui/src/pages/Video/index.js
+++ b/public/video-ui/src/pages/Video/index.js
@@ -293,7 +293,7 @@ class VideoDisplay extends React.Component {
             this.updateEditingState({ key: 'editingWorkflow', editing: false });
             this.saveInWorkflow();
           }}
-          canSave={() => workflow.status.section && workflow.status.status && workflow.status.prodOffice}
+          canSave={() => workflow.status.section && workflow.status.status}
           video={video}
           isTrackedInWorkflow={workflow.status.isTrackedInWorkflow || false}
         />
--- a/public/video-ui/src/pages/Video/index.js
+++ b/public/video-ui/src/pages/Video/index.js
@@ -149,7 +149,7 @@ class VideoDisplay extends React.Component {
 
     const {
       sections,
-      status: { status, section, note, isTrackedInWorkflow }
+      status: { status, section, note, isTrackedInWorkflow, prodOffice }
     } = this.props.workflow;
 
     const {
@@ -163,7 +163,8 @@ class VideoDisplay extends React.Component {
         video: video,
         section: sections.find(_ => _.id === section),
         status: status,
-        note: note
+        note: note,
+        prodOffice: prodOffice
       });
 
     const updateWorkflowItem = () =>
@@ -292,7 +293,7 @@ class VideoDisplay extends React.Component {
             this.updateEditingState({ key: 'editingWorkflow', editing: false });
             this.saveInWorkflow();
           }}
-          canSave={() => workflow.status.section && workflow.status.status}
+          canSave={() => workflow.status.section && workflow.status.status && workflow.status.prodOffice}
           video={video}
           isTrackedInWorkflow={workflow.status.isTrackedInWorkflow || false}
         />
--- a/public/video-ui/src/services/WorkflowApi.js
+++ b/public/video-ui/src/services/WorkflowApi.js
@@ -1,9 +1,8 @@
 import { pandaReqwest } from './pandaReqwest';
 import { getStore } from '../util/storeAccessor';
-import getProductionOffice from '../util/getProductionOffice';
 import VideoUtils from '../util/video';
 import moment from 'moment';
-import { impossiblyDistantDate }  from '../constants/dates';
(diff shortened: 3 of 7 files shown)


public/video-ui/src/actions/WorkflowActions/getProdOffice.js
```diff
@@ -0,0 +1,26 @@
...
+    type: 'SHOW_ERROR',
+    message: `Could not get Workflow production office. <a href="${WorkflowApi.workflowUrl}" target="_blank" rel="noopener">Open Workflow to get a cookie.</a>`,
+    error: error,
+    receivedAt: Date.now()
+  };
+}
+
+export function getProdOffice() {
+  return dispatch => {
+    return WorkflowApi.getProdOffice()
```
joelochlann: did we actually create this function? seems like this file is left over from when we thought we were fetching the prod office stuff from the workflow API (which doesn't actually supply it)

ajwl: oh thanks joe! i should have put a message on this. Didn't get to finish it and I DO need to add this function

public/video-ui/src/pages/Video/index.js
```diff
@@ -158,18 +158,21 @@ class VideoDisplay extends React.Component {
       updateWorkflowData
     } = this.props.workflowActions;
 
-    const createWorkflowItem = () =>
-      trackInWorkflow({
+    const createWorkflowItem = () => {
```
ajwl: can remove these, just added for logging

public/video-ui/src/reducers/workflowReducer.js
```diff
@@ -1,4 +1,4 @@
-export default function workflow(state = { sections: [], statuses: [], status: {}, priorities: [] }, action) {
+export default function workflow(state = { sections: [], statuses: [], status: {}, priorities: [], prodOffice: [] }, action) {
```
ajwl: this may be redundant.. check

public/video-ui/src/components/Workflow/Workflow.js
```diff
@@ -48,6 +48,11 @@ class Workflow extends React.Component {
         workflowStatuses={this.props.workflow.statuses || []}
         workflowPriorities={this.props.workflow.priorities}
         workflowStatus={this.props.workflow.status}
+        workflowProductionOffices={[
+          { id: 'UK', title: 'UK' },
+          { id: 'US', title: 'US' },
+          { id: 'AU', title: 'AU' }
+        ]}
```
ajwl: Setting this in here as it's where the other properties get set. Hardcoding because that's what happens in Composer and workflow itself

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -86,9 +85,9 @@ export default class WorkflowApi {
     video,
     status,
     section,
-    note
+    note,
+    prodOffice
   }) {
-    const prodOffice = getProductionOffice();
```
ajwl: This `getProductionOffice()` no longer required now the value is set manually.  It used to automatically set the production office based on timezone.

ajwl: See also deleted file below

akash1810: Might be worth checking this with Worth checking with Kirsten Broomhall as it means when we first track the atom in Workflow we have to explicitly set the production office rather than estimate it.

ajwl: Good point. This change will affect users..

ajwl: have emailled

ajwl: update: Kirsten wd like the default value of production office to be set to the office that we think they are in. Will reinstate the function and add a default in the select

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -149,7 +149,21 @@ export default class WorkflowApi {
     });
   }
 
-  static updateStatus({id, status}) {
+  static updateProdOffice({ id, prodOffice }) {
+    const payload = {
+      data: prodOffice
+    };
+
+    return pandaReqwest({
```
jonathonherbert: Typo? 😁

jonathonherbert: (oh, wait, it's everywhere. NVM!)

ajwl: a kewl way of spelling I guess.. caught me out first time too

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -149,7 +149,21 @@ export default class WorkflowApi {
...
+      method: 'PUT',
+      url: `${WorkflowApi.workflowUrl}/api/stubs/${id}/prodOffice`,
```
akash1810: This doesn't appear to be a valid route in Workflow.

akash1810: Ah, it is, it's just blocking the request due to CORS. We'd want to enable it first (e.g https://github.com/guardian/workflow-frontend/pull/170)

ajwl: ah great point thank you. I assumed the CORS flagged up because of an environment issue. Will confirm on code and merge

ajwl:
PR to change Workflow endpoints: https://github.com/guardian/workflow-frontend/pull/175

ajwl:
Merging - and will monitor this

PR #886, opened 2019-09-02
BUG: Workflow form does not correctly set default 

ajwl:
The default on the form for the `Production Office` field was not being saved to the state, unless the user specifically interacted with it. 

This caused a lot of form saves to fail as the prodOffice value was not present

1. I have moved the default prodOffice value from where the form fields get created to a defaultState object

### DIFF
--- a/public/video-ui/src/actions/WorkflowActions/getStatus.js
+++ b/public/video-ui/src/actions/WorkflowActions/getStatus.js
@@ -1,5 +1,5 @@
 import WorkflowApi from '../../services/WorkflowApi';
-import { blankWorkflowStatusData } from '../../constants/blankWorkflowStatusData';
+import { defaultWorkflowStatusData } from '../../constants/defaultWorkflowStatusData';
 
 function requestStatus() {
   return {
@@ -20,7 +20,7 @@ function receiveStatus404() {
   return {
     type: 'WORKFLOW_STATUS_NOT_FOUND',
     receivedAt: Date.now(),
-    status: blankWorkflowStatusData
+    status: defaultWorkflowStatusData()
   };
 }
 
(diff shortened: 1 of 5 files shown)


ajwl:
Meant to fix the priority thing. Still not working. But this fixes the main production office bug

ajwl:
@akash1810 if you can figure out why the priority is not set in the POST that would be helpful

ajwl:
Removed the priority work - this is just a fix for the prodoffice bug now

public/video-ui/src/components/FormFields/SelectBox.js
```diff
@@ -10,19 +10,11 @@ export default class SelectBox extends React.Component {
...
-          </option>
-        );
-      } else {
-        return (
-          <option value={null}>
-            {'Please select...'}
-          </option>
-        );
-      }
+      return (
```
jonathonherbert: Is there a reason we want to remove this code?

jonathonherbert: Discussed IRL: introduced in the last PR for this fix, so effectively a revert 👍

ajwl: ftr - because i added it last time, but is now no longer needed

public/video-ui/src/actions/WorkflowActions/getStatus.js
```diff
@@ -20,7 +20,7 @@ function receiveStatus404() {
   return {
     type: 'WORKFLOW_STATUS_NOT_FOUND',
     receivedAt: Date.now(),
-    status: blankWorkflowStatusData
+    status: defaultWorkflowStatusData()
```
akash1810: I think this makes the UX confusing when an atom is not found to be tracked in Workflow as the form looks valid with a production office set.

ajwl: Fair enough. What would you call it? Technically it is no longer blank

PR #887, opened 2019-09-03
Workflow Form Improvement: Save priority first time

ajwl:
When users save a form for the first time, the PRIORITY is not set. ie the POST request doesn't add it.

It wasn't added as a property to the functions making the create request. 

The update PUT request does work.


ajwl:
cheers

PR #888, opened 2019-09-04
WORKFLOW FORM: Fix 0 == false error on saving 'Normal Priority'

ajwl:
An annoying bug came up when saving the "Priority" section on the Workflow form. 

# User problem
From a user perspective, all priorities can be saved except for "Normal". See video below. 

ezgif com-optimize

# The reason
The priorities come in from the Workflow API with a value property that gets read as a number. The problem is that if 'Normal' is selected, the value of the field is assigned `0`. Because `0` evaluates as false in js - this means it fails a bunch of checks throughout the code - along the lines of `fieldValue && xxxxx`. 

The priority object.
```
[
  { "name": "Very-Low", "value": -2 },
  { "name": "Low", "value": -1},
  { "name": "Normal", "value": 0},
  {"name": "Urgent","value": 1},
  {"name": "Very-Urgent","value": 2}
]
```

# Solution
I opted to run an extra cleaning step on atom data loaded in from workflow - this means that `priority: 0` gets converted to a string `priority: "0"` which means it doesn't evaluate to false.

### DIFF
--- a/public/video-ui/src/components/Workflow/Workflow.js
+++ b/public/video-ui/src/components/Workflow/Workflow.js
@@ -46,7 +46,9 @@ class Workflow extends React.Component {
         video={video}
         workflowSections={this.props.workflow.sections || []}
         workflowStatuses={this.props.workflow.statuses || []}
-        workflowPriorities={this.props.workflow.priorities}
+        workflowPriorities={
+          this.props.workflow.priorities.map(({ name, value }) => ({ id: value, title: name })) || []
+        }
         workflowStatus={this.props.workflow.status}
         workflowProductionOffices={[
           { id: 'UK', title: 'UK' },
--- a/public/video-ui/src/services/WorkflowApi.js
+++ b/public/video-ui/src/services/WorkflowApi.js
@@ -73,12 +73,18 @@ export default class WorkflowApi {
     });
   }
 
+  //clean up the workflow data so that the priority field number, which can be 0, is converted to a string
+  static cleanUpWorkflowData(workflowData) {
+    return { ...workflowData, priority: workflowData.priority.toString() };
+  }
+
   static getAtomInWorkflow({ id }) {
     return pandaReqwest({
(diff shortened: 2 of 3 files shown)


public/video-ui/src/components/Workflow/Workflow.js
```diff
@@ -46,13 +46,18 @@ class Workflow extends React.Component {
         video={video}
         workflowSections={this.props.workflow.sections || []}
         workflowStatuses={this.props.workflow.statuses || []}
-        workflowPriorities={this.props.workflow.priorities}
+        workflowPriorities={
+          this.props.workflow.priorities.map(({ name, value }) => {
+            return { id: value.toString(), title: name };
+          }) || []
+        }
```
ajwl: Additionally I moved this transformation code to here and out of `WorkflowForm` so that it's all happening at the same level

public/video-ui/src/services/WorkflowApi.js
```diff
@@ -73,12 +73,18 @@ export default class WorkflowApi {
...
+
   static getAtomInWorkflow({ id }) {
     return pandaReqwest({
       url: `${WorkflowApi.workflowUrl}/api/atom/${id}`,
       crossOrigin: true,
       withCredentials: true
-    }).then(response => WorkflowApi._getResponseAsJson(response).data);
+    }).then(response => WorkflowApi._getResponseAsJson(response).data)
+      .then(jsonRes => this.cleanUpWorkflowData(jsonRes));
   }
```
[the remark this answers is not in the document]
ajwl: Good point! It is a private function that won't be used outside this class. Will amend

ajwl:
Thanks @mkuzdowicz - okay to merge @akash1810 @jonathonherbert ?

public/video-ui/src/components/Workflow/Workflow.js
jonathonherbert: 👍

ajwl:
cheers

PR #927, opened 2020-05-06
add a border to the YouTube thumbnail

akash1810:
Add a 40px red border to the YouTube thumbnail if it's in the `youtube.channels.bordered` configuration.

Example image:
image

### DIFF
--- a/app/util/ThumbnailGenerator.scala
+++ b/app/util/ThumbnailGenerator.scala
@@ -45,18 +45,20 @@ case class ThumbnailGenerator(logoFile: File) {
 
     if(withBorder) {
       val borderThickness = 40
+      val borderOffset = borderThickness / 2
+
       val hexColour = "#C70000"
       val colour = Color.decode(hexColour)
-
-      Logger.info(s"Adding a ${borderThickness}px $hexColour border to thumbnail for atom $atomId")
 
       val currentStroke = graphics.getStroke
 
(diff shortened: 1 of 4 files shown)


app/util/ThumbnailGenerator.scala
```diff
@@ -42,6 +42,23 @@ case class ThumbnailGenerator(logoFile: File) {
...
+      val hexColour = "#C70000"
+      val colour = Color.decode(hexColour)
+
+      Logger.info(s"Adding a ${borderThickness}px $hexColour border to thumbnail for atom $atomId")
+
+      val currentStroke = graphics.getStroke
+
+      graphics.setStroke(new BasicStroke(borderThickness))
+      graphics.setColor(colour)
+      graphics.drawRect(0, 0, bgImage.getWidth, bgImage.getHeight)
```
sihil, line 57: As it stands this seems to draw a 20px border - presumably as the stroke is centred on the very edge of the image. So we could either do `new BasicStroke(borderThickness * 2)` or 
```
vall borderOffset = borderThickness / 2
graphics.drawRect(borderOffset, borderOffset, bgImage.getWidth - borderOffset, bgImage.getHeight - borderOffset)
```

Both are pretty nasty but the second is slightly less of a hack.

akash1810: Updated:

image

PR #929, opened 2020-05-14
log which Grid asset has been chosen for the YouTube thumbnail

akash1810:
Just to provide some information when debugging issues - it can be helpful to know exactly which asset we're using for the thumbnail.


akash1810:
I can't remember the context in which this was raised and the PR isn't the most helpful at providing this insight. As it's been open for a while w/out review, will close for now and we can re-open later if necessary.

PR #935, opened 2020-07-22
Display warning when creating Hosted content

jennygrahamjones:
## What does this change?

When creating a video directly through MAM, a check is made on whether the category "GLabs - Hosted by" has been applied to the video. Previously, if this evaluated as true, the 'Open in Composer' button would display. This button would always link to https://composer.gutools.co.uk/content/undefined, as no corresponding Composer page is created by MAM on creation of a video. The relevant Composer video page will only be created if users create their Hosted video through the Composer dashboard, i.e. through the following menu under 'Create new':

Screenshot 2020-07-22 at 13 15 39

This PR changes the logic for when the 'Open in Composer' button should display, removing the button entirely if the category of the video is Hosted and there are no recorded usages of the video. Instead, a warning is displayed advising that the user should create their Hosted video through Composer.

## What prompted this change?

CP reported an issue whereby a Hosted video appeared to have a Composer video page of https://composer.gutools.co.uk/content/undefined. This happened be cause the content was created via Composer, and the Composer page itself was then deleted manually by CP from the relevant database. As a result, the 'Open in Composer' button had no content to link to.

## How to test

Running this branch locally, create a new video with the category "GLabs - Hosted by". You should see a warning message, advising that you need to create your Hosted video in Composer.

Users do not end up in a situation where they see an 'Open in Composer' button that links to nothing.

It's possible that users will complete the entirety of setting up their video in MAM before realising that this problem exists and this isn't an ideal user experience. After discussion, we believe this risk to be minimal. The current implementation would also cause the user to experience the same frustration, with the additional confusion of displaying the 'Open in Composer' button that linked to nothing. Additionally, Hosted video content has to be created through Composer directly, to provide the relevant 'call to action' metadata, so we do not believe this workflow is a common one.

Ideally, we would want to either a) provide a link to create a Hosted video via Composer or, even better, b) automatically created a corresponding Composer video page when a Hosted video is creating in MAM. This change is therefore only catering to edge cases, such as the uncommon series of steps mentioned in the 'What prompted this change?' section above.

## Images

Before: 
Screenshot 2020-07-22 at 13 01 30

After:
Screenshot 2020-07-22 at 13 00 51

### DIFF
--- a/public/video-ui/src/components/Header.js
+++ b/public/video-ui/src/components/Header.js
@@ -8,6 +8,7 @@ import ComposerPageCreate from './Videos/ComposerPageCreate';
 import Icon from './Icon';
 import { Presence } from './Presence';
 import { canonicalVideoPageExists } from '../util/canonicalVideoPageExists';
+import VideoUtils from '../util/video';
 
 export default class Header extends React.Component {
   state = { presence: null };
@@ -109,6 +110,15 @@ export default class Header extends React.Component {
       return null;
     }
 
+    if (VideoUtils.isHosted(this.props.video)) {
+      return (
+        <div className="header__fields__missing__warning">
+          Hosted video pages need a call to action. Please create this video in
+          Composer
+        </div>
+      );
+    }
+
     if (canonicalVideoPageExists(this.props.usages)) {
       return (
         <div className="header__fields__missing__warning">
@@ -164,7 +174,6 @@ export default class Header extends React.Component {
             {this.renderCreateVideo()}
             {this.renderHelpLink()}
           </div>
-
         </header>
       );
     } else {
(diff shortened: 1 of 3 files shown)


akash1810:
Worth noting that this was the original behaviour in https://github.com/guardian/media-atom-maker/pull/522. That is, somewhere between https://github.com/guardian/media-atom-maker/pull/522 and now a regression was introduced, which it looks like this fixes (well, actually improves on with the messaging change).

jennygrahamjones:
[MAM doesn't handle 'orphaned' videos well](https://trello.com/c/8kxHcPSZ/1977-mam-doesnt-handle-orphaned-videos-well)

PR #936, opened 2020-07-28
update node and node-sass

stephengeller:
## What does this change?

In order to start merging the dependabot deps, it would be good to upgrade Node and node-sass (which needs to be upgraded to work with node 10). This PR updates node and node-sass, in preparation of merging the other dependencies.

The project still runs (tested and running fine locally) and other dependencies can be merged in more easily.


stephengeller:
> It looks like the first build step in Team City still installs Node 6, I assume should this be updated to be 10?
> You could move it all into one script, as has been done in other projects.

Good point! Will update CI and add a ci.sh script.

stephengeller:
> > It looks like the first build step in Team City still installs Node 6, I assume should this be updated to be 10?
> > You could move it all into one script, as has been done in other projects.
> 
> Good point! Will update CI and add a ci.sh script.

So looking into it, there's already a `teamcity.sh` script for CI, and since I'm not sure how else to enforce a node version in CI, I'd maybe opt to just change it to setting node 10 in step one (which I've just done) and otherwise leaving as-is? The CI passes cleanly so this feels easier.

PR #937, opened 2020-07-28
Check node version before starting &  bump webpack-dev-server

stephengeller:
## What does this change?
This ensures you have the right version of Node set before setting up and starting the app

This also bumps webpack-dev-server as it has some sub-dependencies with security vulnerabilities.

## How to measure success

App continues to work, as it does when tested locally!

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "style-loader": "^0.13.1",
     "url-loader": "^0.5.7",
     "webpack": "^1.13.2",
-    "webpack-dev-server": "^1.16.2",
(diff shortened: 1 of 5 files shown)


stephengeller:
> This runs fine locally, although like [guardian/flexible-content#3566](https://github.com/guardian/flexible-content/pull/3566) I wonder whether we can just automate the changing of Node versions by adding in the initial setup to check that nvm is installed, followed by `install` and `use`.

Good shout, this has been implemented alongside the other PR.

PR #942, opened 2020-07-31
Modernise media-atom-maker's dependencies

stephengeller:
## What does this change?

This PR has (at time of writing) 26 open PRs, almost entirely due to security dependencies from dependabot. This appears to primarily be a symptom of the app using very old versions of webpack, react, eslint and jest among others.

Bring media-atom-maker to Webpack v4, and upgrade any linked dependencies that also need upgrading. This will allow us to patch security vulnerabilities more easily in the future as well as develop on the app since a large number of the libraries are currently unsupported and deprecated.

This PR bumps a number of core dependencies to a modern (or even latest) version, allowing us to make use of current docs and features in the future. Namely, the bumped dependencies are:

* Webpack & Webpack loaders (css/sass-loader)
* React & React-dom
* Eslint & Prettier

These upgrades introduced breaking changes that needed to be addressed, mainly in the form of:

* Updating `webpack.{env}.conf.js` config files to the new syntax
* Making the required changes to the library implementation (eg `react-datepicker` using `Date` types as opposed to `moment` and then prettifying the file, causing a larger diff
* Using relative imports for `url()` files in scss, as per new webpack rules

## Definition of Success

The app works exactly like before, maybe even better!

## How to test

Clone the repository locally, run the setup and start scripts and check that it works as intended. Alternatively, check CODE.


.babelrc
```diff
@@ -1,12 +1,14 @@
...
+    "@babel/preset-react"
   ],
   "plugins": [
-    "transform-object-assign",
-    "transform-class-properties",
+    "@babel/plugin-transform-object-assign",
+    "@babel/plugin-proposal-class-properties",
     "transform-es2015-modules-commonjs",
-    "transform-object-rest-spread"
+    "@babel/plugin-proposal-object-rest-spread"
```
stephengeller, line 12: https://babeljs.io/docs/en/v7-migration#switch-to-proposal-for-tc39-proposals-blog-2017-12-27-nearing-the-70-releasehtml-renames-proposal

build_config/webpack.dev.conf.js
```diff
@@ -1,43 +1,50 @@
-var ExtractTextPlugin = require("extract-text-webpack-plugin");
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
```
stephengeller, line 1: This library seems to be popular and more-commonly used in webpack builds, and was more effective at producing the `main.css` file for the app to ingest.

build_config/webpack.dev.conf.js
```diff
@@ -1,43 +1,50 @@
...
-        test: /\.css$/,
-        use: ["style-loader", "css-loader?sourceMap"]
+        test: /\.(sa|sc|c)ss$/,
+        use: [
+          {
+            loader: MiniCssExtractPlugin.loader
+          },
+          'css-loader',
+          'sass-loader'
+        ]
```
stephengeller, line 22: This compiles both css and scss: https://webpack.js.org/plugins/mini-css-extract-plugin/

build_config/webpack.prod.conf.js
```diff
@@ -1,49 +1,57 @@
...
     noParse: [/aws\-sdk/]
   },
   plugins: [
-    new ExtractTextPlugin("main.css"),
+    new MiniCssExtractPlugin({
+      filename: '[name].css',
+      chunkFilename: '[id].css',
+      ignoreOrder: false // Enable to remove warnings about conflicting order
+    }),
+    new webpack.optimize.ModuleConcatenationPlugin(),
```
stephengeller, line 46: https://webpack.js.org/plugins/module-concatenation-plugin/

package.json
```diff
@@ -3,10 +3,10 @@
   "description": "media atom editor",
   "scripts": {
     "build-autotrack": "autotrack -o public/video-ui/build/autotrack.js -p cleanUrlTracker,eventTracker,impressionTracker,maxScrollTracker,mediaQueryTracker,outboundFormTracker,outboundLinkTracker,pageVisibilityTracker,socialWidgetTracker,urlChangeTracker",
-    "build": "webpack -p ./public/video-ui/src/app.js ./public/video-ui/build/app.js --config ./build_config/webpack.prod.conf.js",
-    "build-dev": "webpack ./public/video-ui/src/app.js ./public/video-ui/build/app.js --config ./build_config/webpack.dev.conf.js --watch",
+    "build": "webpack -p ./public/video-ui/src/app.js -o ./public/video-ui/build/app.js --config ./build_config/webpack.prod.conf.js",
+    "build-dev": "webpack -o ./public/video-ui/build/app.js --config ./build_config/webpack.dev.conf.js --watch",
     "client-dev": "node ./webpack-dev-server.js",
```
stephengeller, line 8: Adding `-o` specifies that the second arg is the output, which looks like it wasn't the case in webpack v1. If you omit the `-o` it would just compile the build directory as well, meaning you're "building" the build again.

public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -1,11 +1,11 @@
...
-import Picker from 'react-datepicker';
+import DatePicker from 'react-datepicker';
 import moment from 'moment';
 import 'react-datepicker/dist/react-datepicker.css';
 import Icon from '../Icon';
 
-const DATE_FORMAT = 'DD MMM YYYY';
-const DATETIME_FORMAT = `${DATE_FORMAT} HH:mm`;
+const DATE_FORMAT = 'dd MMM yyyy';
+const DATETIME_FORMAT = `DD MMM yyyy HH:mm`;
```
stephengeller, line 9: Upgrading `react-datepicker` meant we had to format the dates a bit differently to achieve the same result as the current build.

public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -57,19 +61,23 @@ function MinuteSelector({ date, onChange }) {
...
-      onChange(newDate.hours(base.hours()).minutes(base.minutes()));
+      const onChangeDate = moment(newDate)
+        .hours(base.hours())
+        .minutes(base.minutes());
+      onChange(onChangeDate);
     }
   };
 
-  return <Picker {...datePickerParams} />;
+  return <DatePicker {...datePickerParams} />;
```
stephengeller, line 80: `react-datepicker` uses Date objects to work, or at least the examples suggest this: https://reactdatepicker.com/

Using `moment()` types didn't work after the upgrade, so these changes convert moment to date types when passing to `DatePicker`

public/video-ui/src/services/pandaReqwest.js
```diff
@@ -1,15 +1,15 @@
-import reqwest from "reqwest";
-import { reEstablishSession } from "babel-loader?presets[]=es2015!panda-session";
-import { getStore } from "../util/storeAccessor";
+import reqwest from 'reqwest';
+import { reEstablishSession } from 'babel-loader!panda-session';
```
stephengeller, line 2: `@babel/preset-env` handles this stuff now, so no need to set the preset in the import

public/video-ui/styles/components/_buttons.scss
```diff
@@ -74,7 +74,7 @@ button {
 
 .btn--loading {
   background: {
-    image: url(/assets/video-ui/images/progress-spinner.gif);
+    image: url(../images/progress-spinner.gif);
```
stephengeller, line 77: Can't find the docs on it now, but upgrading webpack resulted in needing to change these to relative imports rather than absolute.

public/video-ui/src/components/FormFields/DatePicker.js
```diff
@@ -57,19 +61,23 @@ function MinuteSelector({ date, onChange }) {
...
+  const minDate = moment().toDate();
   const datePickerParams = {
     className: 'form__field',
-    selected: date,
-    minDate: moment(),
+    selected: date ? date.toDate() : null,
+    minDate,
     dateFormat: DATE_FORMAT,
-    readOnly: true,
+    readOnly: false,
```
jennygrahamjones, line 70: Is there a reason this has changed?

stephengeller: Yeah, it looked like editing the date wasn't possible with readOnly being true. I couldn't work out why that was the case, so switching it to false seemed to allow the date to be edited. It would be good to confirm this with someone, I'll do that before merging this in.

PR #944, opened 2020-08-07
Begin using new syntax and clean up errors

stephengeller:
## What does this change?

Now that we've updated Webpack and other dependencies, we are using deprecated syntax across the app and could do some cleaning up. This is the first of a few cleanup PRs planned to address these deprecations:

* `componentWill<ReceiveProps|Mount>` are deprecated functions and need replacing, and we currently receive warnings in the console to do so. This PR uses `React.memo` and `componentDidUpdate` as recommended instead, as well as prettifies the file
* There is a random `r` (likely typo) in `TextInputTagPicker` that causes a warning. This PR removes it and prettifies the file

## How to test

The app should behave as normal. It's currently out in CODE to test.

Less warning messages should show up in the console when developing locally.


public/video-ui/src/components/CapiSearch/CapiSearch.js
```diff
@@ -1,27 +1,40 @@
...
+    ) {
       const listNode = this.refs.list;
       const elementHeight = listNode.children[0].offsetHeight;
       if (listNode) {
-        listNode.scrollTop = elementHeight * (nextProps.selectedTagIndex === 0 ? 0 : nextProps.selectedTagIndex - 1)
+        listNode.scrollTop =
+          elementHeight *
+          (nextProps.selectedTagIndex === 0
+            ? 0
+            : nextProps.selectedTagIndex - 1);
```
stephengeller, line 22: This is the main change, the rest is Prettier

public/video-ui/src/components/CapiSearch/CapiSearch.js
```diff
@@ -30,26 +43,29 @@ export default class CapiSearch extends React.Component {
...
 
       this.props.selectNewTag(newFieldValue);
-
     };
 
     return (
       <li
         className={getTagClassName()}
-        key={tag.id}
+        key={tag.id + index}
```
stephengeller, line 51: This is done to ensure unique keys, as we were getting errors

PR #945, opened 2020-08-11
Fix embedding into Composer issue by moving setup behaviour to constructor

stephengeller:
## What does this change?

While updating old deprecated `componentWillMount` syntax to the newer best practices, a mistake was made ~by me~ in moving some behaviour to `componentDidMount` rather than in the constructor. This meant that when rendering the `VideoSelectBar` component in the embedded MAM, it was missing a video prop, and would therefore bail on not being able to access `this.props.video.title` because `video` was `null`.

This PR ensures the component does not render until the setup behaviour is completed by moving the behaviour to the constructor, and also adds a ternary to ensure we don't run into the same issue again.

## How to test

If you create a space between paras in Composer and click "Video" in the embed menu, and then select a video to embed from the embedded version of MAM, the screen should not go blank and you can embed a video as expected.

Users can embed videos as normal.

## Images

Problem email:
image

### DIFF
--- a/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
+++ b/public/video-ui/src/components/VideoSelectBar/VideoSelectBar.js
@@ -37,11 +37,15 @@ export default class VideoSelectBar extends React.Component {
       return false;
     }
 
+    const title = this.props.video && this.props.video.title
+      ? this.props.video.title
+      : "Title missing";
+
     return (
       <div className="bar info-bar">
         <div className="bar__image">{this.renderItemImage()}</div>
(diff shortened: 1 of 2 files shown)


stephengeller:
> The code changes look straightforward but I haven't tested it. Happy to test it on CODE if that would be useful?

Good shout, I'm just deploying it now. You should be able to test it and see the embedding working?

SHession:
> Good shout, I'm just deploying it now. You should be able to test it and see the embedding working?

I've now tested this on composer CODE and I'm happy it works as expected.

PR #949, opened 2020-09-01
Fixes the byline rendering undefined

SHession:
## What does this change?

Users have been reporting issues with names entered in the byline input box. The names have been display as undefined when they do not match a tag. This PR reverts some changes made in #943, referencing a value which is not available. 

## How to test

Enter a combination of free text and tags into the byline input, the outputted string should display as expected.

Users no longer experience issues with MaM byline input. 

We are uncertain of the exact shape of the `value` object and so can't be 100% of this fix. Typescript would improve this situation!


jennygrahamjones (approved):
Tested in CODE. Byline contributors missing from Tag Manager no longer appear as `undefined` 👍

PR #950, opened 2020-09-01
Render web title for youtube furniture keywords

stephengeller:
## What does this change?

This PR (https://github.com/guardian/media-atom-maker/pull/949) unfortunately introduced a regression where the `TagFieldValue` component breaks the site, as it receives an object with `id` and `webTitle` fields that isn't being dealt with.

This PR fixes that and renders the `webTitle` field, and adds a comment to explain the reasoning.

## How to test

Spin up the branch, go to a video, add some keywords to YouTube Furniture and see it add successfully!

Adding keywords to YouTube Furniture doesn't crash the site!


jonathonherbert (approved):
Tested locally vs. master, creating keywords no longer crashes the React app 👍

PR #953, opened 2020-09-28
Set video adformats on all API uploads

jennygrahamjones:
We believe that ad formats were previously being applied based on the default settings for our channel, however this no longer appears to be the case. This PR explicitly sets a list of `adFormats` on all videos created or updated, as long as the video has not been flagged specifically to block ads. See [the YouTube developer docs](https://developers.google.com/youtube/partner/docs/v1/videoAdvertisingOptions/update) for more information on the APIs being used.

## How to test
In local or CODE, upload a new video clip that is longer than 30 seconds and publish it. Use the `.../api/youtube/video-info/<youtube-id-here>` endpoint in MAM to see which `adFormats` are applied to the video. Alternatively/additionally, navigate to the YouTube dashboard for the relevant channel and check the monetisation settings for the video. 

Desired ad formats are applied to videos uploaded via the API.

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
@@ -173,13 +173,15 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
     }
   }
 
-  private def updateTheVideoAdvertisingOptions(videoId: String): Either[VideoUpdateError, String] = {
+  private def updateTheVideoAdvertisingOptions(videoId: String, atomId: String): Either[VideoUpdateError, String] = {
     val formats: util.List[String] = List("trueview_instream", "long", "overlay", "product_listing", "standard_instream", "third_party", "display").asJava
     val advertisingOption: VideoAdvertisingOption = new VideoAdvertisingOption().setAdFormats(formats)
     try {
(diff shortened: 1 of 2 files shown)


jennygrahamjones:
[Explicitly set video ad formats in Media Atom Maker](https://trello.com/c/iie2Lu0m/2060-explicitly-set-video-ad-formats-in-media-atom-maker)

common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
```diff
@@ -171,8 +173,28 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
     }
   }
 
+  private def updateTheVideoAdvertisingOptions(videoId: String, atomId: String): Either[VideoUpdateError, String] = {
+    val formats: util.List[String] = List("standard_instream","trueview_instream","display").asJava
```
akash1810, line 177: Worth adding a comment to the YouTube docs for future reference if we decide to add/remove formats?

akash1810 (dismissed):
Looks good! Thanks!

common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
jennygrahamjones: Yeah, good shout 👍

akash1810 (approved):

PR #966, opened 2020-11-09
Fix defaultMediaHtml

shtukas:
## What does this change?

Fix a problem in `defaultMediaHtml` by which the list wasn't converted into a string before returning. This was only noticed because of the following output in the data coming from CAPI.

unnamed

In browsers that don't support HTML5 video, it would come out like this:

unnamed-1


oliverlloyd (approved):
👍

akash1810 (commented):
Would be nice to have a test for this.

PR #973, opened 2020-12-21
TEMPORARY: tolerate the thumbnail update failing with generic UI warning

twrichards:
Co-authored-by: @jennygrahamjones 

## Context
Much like https://issuetracker.google.com/issues/175959405 we've been experiencing 403s when trying to update the thumbnails on live (or previously live) YouTube videos. Previously the thumbnail updates are a critical part of the process and this was blocking publishing.

## What does this change?
We now catch and log the exception but allow the publish to complete. We present a generic warning in the UI with a link to update the thumbnail manually...
image

## How to test
With this deployed to CODE, update a thumbnail on a live (or previously live) YouTube video (e.g. https://video.code.dev-gutools.co.uk/videos/7d417a50-3df4-4db0-8ac8-325abd812fc6) and observe the `Failed to update thumbnail` in logs.gutools.co.uk

Video team can publish live video atoms (albeit having to update thumbnails manually).

The thumbnail auto-selected by google might be undesirable until it is updated manually (likely only CP and/or devs will be able to update manually).

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala
@@ -8,6 +8,7 @@ import com.gu.media.logging.Logging
 import com.gu.media.model.VideoUpdateError
 import com.gu.media.util.ISO8601Duration
 import com.gu.media.logging.{YoutubeApiType, YoutubeRequestLogger, YoutubeRequestType}
+import net.logstash.logback.marker.Markers
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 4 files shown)


twrichards:
given the temporary nature of this workaround (YouTube bug ticket is assigned to an engineer)... there were too many layers between the exception and the http response to our client to pass any sort of headers/flags to drive the display of the warning so we just did a generic/permanent one

common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala
```diff
@@ -131,13 +132,23 @@ trait YouTubeVideos { this: YouTubeAccess with Logging =>
...
 
             YoutubeRequestLogger.logRequest(YoutubeApiType.DataApi, YoutubeRequestType.UpdateVideoThumbnail)
-            set.execute()
+            try{
+              setOperation.execute()
+            }
+            catch {
+              case exception: Throwable =>
+                log.error(
+                  "Failed to update thumbnail",
```
akash1810, line 147: Is it worth adding a bit more context to this log line to aid debugging? Add the atom or video ids, maybe?

akash1810 (commented):
Is it worth narrowing this to only live videos, as only they seem to be affected by the issue linked? I don't think we model live video on the atom, but we might be able to use [`isLiveStream`](https://github.com/guardian/media-atom-maker/blob/1083940cd529054667de8e98158cbf037eb72afe/common/src/main/scala/com/gu/media/youtube/YouTubeProcessingStatus.scala#L41)?

jennygrahamjones:
Google are reporting that should now be fixed: https://issuetracker.google.com/issues/175959405#comment9

twrichards:
> Google are reporting that this should now be fixed: https://issuetracker.google.com/issues/175959405#comment9

As @jennygrahamjones points out, the bug appears to be fixed...
image

So closing this PR as it's no longer needed.

davidfurey:
> given the temporary nature of this workaround (YouTube bug ticket is assigned to an engineer)... there were too many layers between the exception and the http response to our client to pass any sort of headers/flags to drive the display of the warning so we just did a generic/permanent one

very pragmatic 👍

PR #975, opened 2021-01-18
Allow uploads over slower connections

jonathonherbert:
_Co-sleuthed with @mbarton_

## What does this change?

Adjusts the S3 upload options to allow uploads over slower connections. 

The aws-sdk S3 client makes a few choices in its defaults that prevent uploads over slower connections:

- it makes four upload requests concurrently by default
- it has a per-request timeout of 120s
- it has a default chunk size of around 8mb

This means that for upload speeds of <0.5-6mbps, assuming an even distribution of bandwidth across chunks, a single chunk will never complete 😢 – something we saw in PROD last Friday, and a scenario that's more common as people WFH.

This PR changes those defaults to better accommodate slow connections. Descriptions of the defaults and why they are changed are colocated with the settings in the code.

## How to test

Throttle your network connection to 'real slow', perhaps using e.g. Chrome's `slow 3g` setting (about [400kbps?](https://stackoverflow.com/questions/48367042/in-chrome-dev-tools-what-is-the-speed-of-each-preset-option-for-network-throttl)) and upload a video. You should find that previous to this PR, this doesn't work – now, it should. You should be able to verify that a single chunk is uploaded at a time in the network tab:  

Screenshot 2021-01-18 at 17 08 31

Tested locally and in CODE w/ [the video that caused problems for a user in PROD.](https://wetransfer.com/downloads/a9ca30c09448e1106a8915244326b8c920210115201513/4ea02a)

Users with slower connections are able to upload videos.

We could remove the timeout completely! Is there a good reason not to do this?

### DIFF
--- a/public/video-ui/src/services/UploadsApi.js
+++ b/public/video-ui/src/services/UploadsApi.js
@@ -5,6 +5,18 @@ import { errorDetails } from '../util/errorDetails';
 import 'aws-sdk/dist/aws-sdk';
 const AWS = window.AWS;
 
+// The timeout for individual upload requests. Defaults to 120s. This
+// default causes problems on slow connections – for example, w/ a 0.5mbps
+// upload speed (3.75MB/minute), uploads for 8mb chunks will never complete.
+AWS.config.httpOptions.timeout = 240_000; // in ms
+const httpOptions = {
+  // The number of multipart uploads to run concurrently. Defaults to 4,
+  // which has caused problems with slow connections timing out requests
+  // prematurely. We judge allowing uploads on slow connections to be
+  // more valuable than a minor boost in upload speed due to concurrent uploads.
+  queueSize: 1
+};
+
 export function getUploads(atomId) {
   return pandaReqwest({
     url: `/api/uploads?atomId=${atomId}`
@@ -63,7 +75,7 @@ function uploadPart(upload, part, file, progressFn) {
       ACL: 'private',
       Metadata: { original: file.name }
     };
-    const request = s3.upload(params);
+    const request = s3.upload(params, httpOptions);
 
     request.on('httpUploadProgress', event => {
       progressFn(part.start + event.loaded);


public/video-ui/src/services/UploadsApi.js
```diff
@@ -5,6 +5,18 @@ import { errorDetails } from '../util/errorDetails';
 import 'aws-sdk/dist/aws-sdk';
 const AWS = window.AWS;
 
+// The timeout for individual upload requests. Defaults to 120s. This
+// default causes problems on slow connections – for example, w/ a 0.5mbps
+// upload speed (3.75MB/minute), uploads for 8mb chunks will never complete.
+AWS.config.httpOptions.timeout = 240_000; // in ms
```
sihil, line 11: why the underscore?

public/video-ui/src/services/UploadsApi.js
```diff
@@ -5,6 +5,18 @@ import { errorDetails } from '../util/errorDetails';
...
+const httpOptions = {
+  // The number of multipart uploads to run concurrently. Defaults to 4,
+  // which has caused problems with slow connections timing out requests
+  // prematurely. We judge allowing uploads on slow connections to be
+  // more valuable than a minor boost in upload speed due to concurrent uploads.
+  queueSize: 1
+};
```
sihil, line 18: Is this two different ways of changing the same class of setting? The top a global override and the other per-call? If they could be done the same way then that might be less confusing?

jonathonherbert: Agreed – sadly, looking at the type declaration for `.upload`, I don't think that's possible.

public/video-ui/src/services/UploadsApi.js
jonathonherbert: Arguably easier to read than `240000`. https://v8.dev/features/numeric-separators

public/video-ui/src/services/UploadsApi.js
jonathonherbert: That is to say, the HTTP settings seem global, and cannot be passed in any shape that I can find to `upload`, but hmm, having said that, we might be able to configure queueSize globally as you suggest! I'll take a look.

akash1810:
Have you considered [S3 accelerated transfer](https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html)? I believe it's faster by adding a CloudFront distribution on top of S3, so you end up uploading to a POP, which is closer to you.

public/video-ui/src/services/UploadsApi.js
jonathonherbert: I can't see a way to do this, but v. happy to be corrected. (Can only see `queueSize` refs in `ManagedUpload.ManagedUploadOptions`, which are only referenced in the signature for `upload`.)

jonathonherbert:
@akash1810 this is an interesting suggestion – I don't think there's any evidence at the moment that uploading is problematically slow w/ MAM; the user in this had a generally slow connection. But if slow uploads are a problem with MAM this feels like a solution for sure. Could always ticket this and ask users, although we're somewhat stretched for health work as it is this quarter.

mbarton:
> Have you considered [S3 accelerated transfer](https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html)? I believe it's faster by adding a CloudFront distribution on top of S3, so you end up uploading to a POP, which is closer to you.

For a slow home internet connection I'm not sure this would make it much faster but may well make it more reliable and would almost certainly benefit users in AUS or the US.

It does need a different S3 endpoint (`bucketname.s3-accelerate.amazonaws.com`) so it might need code changes?

If we were working on MAM full-time I'd suggest doing it proactively now. As it stands I think we should try and gather more information about any further issues with uploading and then schedule the work if we think it will help

mbarton (approved):

PR #976, opened 2021-01-22
explicitly set break positions

akash1810:
## What does this change?

We're witnessing a varying combination of ad settings within YouTube; some videos have ads enabled for pre, mid and post roll, other videos have just mid roll, others have just pre roll and all the other permutations.

In this PR, we set the [break positions](https://developers.google.com/youtube/partner/docs/v1/videoAdvertisingOptions#breakPosition[]) explicitly to make this more deterministic and as a result make it easier to forecast and report on YouTube ad revenue.

Also only enables midroll advertising if the video is eligible ([over 8 minutes](https://support.google.com/youtube/answer/6175006?hl=en-GB)). For simplicity `AdSettings` has been created to encapsulate the logic, w/tests.

The advertising rules are:
- Ads are blocked if the video is from Labs (Paid or Hosted)
- Videos need to be over `minDurationForAds` in length to be eligible for advertising
- If Editorial enable ads, then:
  - Preroll will be enabled
  - Midroll will be enabled if and only if the video is over 8 minutes long
- If Editorial disable ads, then:
  - Preroll is disabled
  - Midroll is disabled

## How to test

- Deploy to CODE
- Upload a video of at least 30 seconds(1) to YouTube
- Publish atom
- Witness only preroll and midroll ad positions being enabled. Consistently.

(1) we have [a rule](https://github.com/guardian/media-atom-maker/blob/1083940cd529054667de8e98158cbf037eb72afe/app/model/commands/PublishAtomCommand.scala#L118) that says a video must be >30s to be eligible for monetisation.

The ad formats enabled for a video uploaded via MAM is deterministic and makes life easier for our colleagues in Commercial.

n/a

### Before 1 - midroll only
image

### Before 2 - preroll, midroll, postroll w/midroll editable
image

### Before 3 - preroll, midroll, postroll. Video is <8 mins and YouTube are warning about midroll selection!
image

### After (with a video > 8 minutes) - preroll, midroll only, w/both editable 🎉 
image


jonathonherbert (approved):
This is a very readable change: it was easy for me to move from the PR description to the code and understand what was going on, although I haven't worked with video ads before – (pre|mid|post)roll, TIL!

One idea – would it be nice to colocate the logic re: none, preroll, midroll etc. in one place, to make it clearer to readers looking for the reasons we take ad decisions? At the mooment, they'd need to look in `AdSettings` and `updateTheVideoAdvertisingOptions`. 

As an example, we could pass the videoSettings case class instance into `updateTheVideoAdvertisingOptions`, and add a method on the case class, `getBreakPositions`, which puts all of our logic in one place. It also provides a place to add a comment if anything bears further explanation.

V. happy to pair if that's unclear, and this is a nice-to-have, the code is very readable as it is.

akash1810:
> One idea – would it be nice to colocate the logic

Agreed! I tried this here, but the diff was getting pretty big so opted out. I think it makes total sense for anything of the same concern to be co-located for simplicity. We might want to simplify config too as currently config is read once as a big object which is passed around.

Let's make these improvements separately.

PR #977, opened 2021-03-05
Ensure channel is added when assetUrl field is used

tjsilver:
Co-authored-by: Akash Askoolum <akash1810@users.noreply.github.com>

## What does this change?
At the moment, if a user pastes an existing Youtube URL into MAM without setting the channel first, the channel field is locked and thus the video cannot be published.

## How to test
Deploy `main` to CODE and do the following: 

1. Obtain URL to a video on Youtube.com
2. Create a new atom, without setting the `Youtube channel` field
3. Use the `asset URL` feature to add the video from step 1 to the atom
4. Observe the `channel` field is now un-editable, the atom can't be published and uploading is disabled.

Deploy this branch to CODE and repeat steps 1-3 above. Observe the `channel` field is now filled in.

Users don't need to contact the team to manually set the channel.

### DIFF
--- a/app/model/commands/AddAssetCommand.scala
+++ b/app/model/commands/AddAssetCommand.scala
@@ -49,7 +49,7 @@ case class AddAssetCommand(atomId: String, videoUri: String, override val stores
     val newAsset = ThriftUtil.parseAsset(uri = videoUri, version = version, mimeType = None)
       .fold(err => AssetParseFailed, identity)
 
-    val channel = getYouTubeChannel(newAsset, mediaAtom)
+    val channel = getYouTubeChannel(newAsset, mediaAtom, atom.id)
     val metadata = mediaAtom.metadata.getOrElse(Metadata()).copy(channelId = channel)
 
     val updatedAtom = atom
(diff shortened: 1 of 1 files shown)


akash1810 (approved):

app/model/commands/AddAssetCommand.scala
```diff
@@ -63,42 +63,54 @@ case class AddAssetCommand(atomId: String, videoUri: String, override val stores
...
 
-        case (None, Some(video)) => {
-          // only GLabs atoms can have third party videos
-          atom.category match {
-            case ThriftCategory.Hosted | ThriftCategory.Paid => Some(video.getSnippet.getChannelId)
-            case _ => {
-              if (youTube.cannotReachYoutube) None
-              else NotGLabsAtom
+    if (youTube.cannotReachYoutube) {
+      log.info(s"Config says cannot reach Youtube")
```
twrichards, line 74: what does `Config` mean in this case, does it mean what the user has configured or system config (from S3 etc.)?

twrichards (approved):
LGTM 👍 - one small question around the log message

app/model/commands/AddAssetCommand.scala
akash1810: System configuration - a file on disk which is copied from S3 at instance launch.

PR #978, opened 2021-03-09
handle null responses when listing YouTube 'partner claims'

twrichards:
Occasionally users were getting an error when deleting an atom...
image
... the logs revealed this was due to null responses when trying to list 'partner claims' (YouTube API).

## What does this change?
This PR first checks for null on both the response itself but also all the fields that will be used and returns None, much like if the response had been fully formed but was an empty list.

## How to test
I guess one could try to use a problematic you tube video (for example the one from https://video.gutools.co.uk/videos/b9aca7f6-181d-4ae4-a043-ba04d74492bc) and create a new atom, then attempt to delete.

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
@@ -153,7 +153,7 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
     }
   }
 
-  private def getPartnerClaim(videoId: String): Option[ClaimSnippet] = {
+  private def getPartnerClaim(atomId: String, videoId: String): Option[ClaimSnippet] = {
     val request = partnerClient
       .claimSearch()
       .list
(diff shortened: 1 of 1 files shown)


common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
```diff
@@ -165,7 +165,11 @@ trait YouTubePartnerApi { this: YouTubeAccess with Logging =>
     YoutubeRequestLogger.logRequest(YoutubeApiType.PartnerApi, YoutubeRequestType.GetVideoClaim)
     val response = request.execute()
 
-    if (response.getPageInfo.getTotalResults == 0) {
+    if(response == null || response.getPageInfo == null || response.getPageInfo.getTotalResults == null || response.getItems == null) {
```
akash1810, line 168: Should we add an inline comment against this?

twrichards: personally I think that would be noise, null checks when using Java libs are fairly self-explanatory and the log message beneath hints at this, plus if anybody is really interested they can git blame to find this PR...

jonathonherbert (approved):
Nice to see logging for this, LGTM!

common/src/main/scala/com/gu/media/youtube/YouTubePartnerApi.scala
akash1810: I interpreted this change as "youtube's apis cannot be trusted" more than anything else. Happy to omit the comment if git blame is enough.

PR #980, opened 2021-03-10
better handle null responses when listing YouTube 'partner claims'

twrichards:
A slight tweak to https://github.com/guardian/media-atom-maker/pull/978 now that we've seen some log entries that show the response itself is not null, but in certain instances* doesn't include the `pageInfo` property and so we still want to avoid the original `NullPointerException` but we first check if there are `items` and check those... which I think is the more reliable behaviour.

\* youtube may have made a permanent _breaking_ change to their API


jonathonherbert (approved):
Happy to rubber stamp – guessing this is difficult to test!

PR #981, opened 2021-03-12
feat: never disable the channel selection box

akash1810:
## What does this change?

We've had a few reports around YouTube channel selection. Namely, in some scenarios we end up with an Atom without a YouTube channel unexpectedly.

This change lifts the editing restrictions on this field - you can now edit the channel at any time.

I'm not sure if this is the correct thing to do as the restriction was added for a reason, though I'm struggling to find a PR that explains this rule. My memory thinks one reason is we don't want to be in a position where we upload a video to channel A, switch channels, then upload an edited version of the video to channel B. Need to confirm with the video team.

I think we'd want to introduce tests against the [problematic code](https://github.com/guardian/media-atom-maker/blob/ad2bb4aa44a04847a3ac1de71f77d87fe8941f2c/app/model/commands/AddAssetCommand.scala#L66-L115) as a better solution to these user reports too... I'll let you be the judge!

## How to test

- Deploy to CODE
- Create an atom
- Set the channel
- Upload a video
- Observe the channel can be changed

Empowering our users!

See above concerns.

## Images

n/a

### DIFF
--- a/public/video-ui/src/components/YoutubeFurniture/index.js
+++ b/public/video-ui/src/components/YoutubeFurniture/index.js
@@ -96,7 +96,6 @@ class YoutubeFurniture extends React.Component {
     const availableChannels = VideoUtils.getAvailableChannels(video);
     const availablePrivacyStates = VideoUtils.getAvailablePrivacyStates(video);
     const hasYoutubeWriteAccess = VideoUtils.hasYoutubeWriteAccess(video);
-    const hasAssets = VideoUtils.hasAssets(video);
 
     return (
       <ManagedForm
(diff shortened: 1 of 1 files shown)


jonathonherbert:
Thanks for this, @akash1810! 

Re: _>we don't want to be in a position where we upload a video to channel A, switch channels, then upload an edited version of the video to channel B_
Not familiar w/ video workflows, could you go into detail as to why we wouldn't want that to happen?

If this needs shipping to make folks comfortable over the weekend, happy to approve.

akash1810:
> Not familiar w/ video workflows, could you go into detail as to why we wouldn't want that to happen?

We have channels for Sport, News, Culture etc.

The Atom tool makes it easy to upload to any of these channels as you only need permission to the Tool.

YouTube do not allow editing of a video. If you want to edit a video you have to upload a new video and remove (or make private) the old one. This change will mean that videos and their subsequent edits can live on different YouTube channels, should the user do the wrong thing.

I suspect this change will be fine, but we probably want to run it by the video team. I think it can wait for next week too, given the current time.

twrichards:
Closing in favour of slightly safer https://github.com/guardian/media-atom-maker/pull/982 🤞

PR #982, opened 2021-03-16
also allow channel selection when its currently null

twrichards:
A slight variant of https://github.com/guardian/media-atom-maker/pull/981 (which was closed in favour of this PR) which instead allows channel selection when channel is null or when there are no assets (as before) but not ALL the time. This PR seeks to unblock users whilst also trying to limit some of the risk outlined in https://github.com/guardian/media-atom-maker/pull/981.


jonathonherbert (approved):
Seems very sensible to me, although @akash1810 may have more thoughts!

akash1810 (approved):
Yes, this is a vast improvement from the other PR, thanks!

PR #983, opened 2021-03-16
Upgrade google dependencies

SHession:
## What does this change?

This PR updates various google dependancies in an effort to help resolve issues that users are currently experiencing.

This project includes an unmanaged dependency, the YouTube Content ID API library. The latest version of this can be found at https://developers.google.com/youtube/partner/client_libraries

## How to test

The unit tests for this build pass. We should deploy the changes to CODE to ensure no regressions are introduced. 

We can eliminate or rule out, API changes as the cause of the recent user issues.

The changes to these dependencies could cause unexpected issues.


twrichards:
hmm I was expecting compilation errors, in that they appear to have removed `pageInfo` for example from the response - perhaps we ought to raise a bug report with Google. This PR is likely still a good thing and may fix some of these channel issues 🙏

SHession:
> hmm I was expecting compilation errors

Me too. That partner library is still almost two years old, it's possible there may be a newer version available somewhere.

SHession:
Tested on CODE, seems to work fine. Ready for review!

twrichards:
> > hmm I was expecting compilation errors
> 
> Me too. That partner library is still almost two years old, it's possible there may be a newer version available somewhere.
> 
> Looking at the revision history for the API, I also can't see any significant recent changes: https://developers.google.com/youtube/partner/revision_history

I have reservations about merging this, since it clearly doesn't match the current state of the rest API (see https://github.com/guardian/media-atom-maker/pull/980) as we don't see compilation errors and as you say it's two years old... but its three years worth of changes since the one we have in the repo currently (2016) so things have def. changed in the internals and so my fear is it could cause really pernicious problems down the line?

That said perhaps it sends extra params now which means the responses come back in the right shape and so perhaps it is worth it, can we replicate in `CODE` the problem we're trying to solve, and if this version bump helps then it probably is worth it?

twrichards (approved):

SHession:
We are going to tentatively deploy this change and keep an eye on any user reported issues.

PR #988, opened 2021-04-14
Delete media-atom-maker-dev.yml

AWare:
Looks like this stack hasn't been used for a while, can I delete it? 

If not, where is the IAM user used so I can update the key?


davidfurey:
I believe this is still used

PR #991, opened 2021-04-28
Removes bintray resolver and changes panda-hmac to panda-hmac-play

SHession:
## What does this change?

This PR removes the bintray resolver ahead of the end of service. The only change required was to use the Maven version of `panda-hmac` which I believe is `panda-hmac-play`: https://search.maven.org/search?q=panda-hmac

## How to test

Does the project compile and work as expected?


jonathonherbert (approved):
Deployed to CODE, app loads, authenticates as expected 👍

PR #992, opened 2021-04-28
Stop just targeting to commissioning desks

blishen:
Change the tag type for the picker to keyword to allow video targeting.


blishen:
Akash and Mateusz both think sane.

akash1810 (approved):

PR #995, opened 2021-05-05
Fix build issues

SHession:
## What does this change?

This PR fixes an intermittent issue with the MAM build whereby the `pluto-message-ingestion` artifact was intermittently failing to create. This would result in [deploy failures](https://riffraff.gutools.co.uk/deployment/view/82de8646-5bf2-4005-ac17-3c98589b00af), as riff-raff wouldn't be able to update the lambda.

The issue is fixed by removing `node-riffraff-artefact` which was being used to zip up the project, and replaces it with `ncc` and `ncc-zip`. As the NCC is not simply including the entire project directory, the artifact size has gone from 24.9 mb -> 11.4 kb. This should have significant benefits in terms of warmup time for the lambda, and may also improve the overall project build time.

The PR also upgraded the `sbt-riffraff-artifact` plugin to the latest version for good measure.

## How to test

Ensure the project consistently builds in TeamCity and deploys correctly to `CODE`.

We have tested the `pluto-message-ingestion` lambda works correctly by getting pluto to generate a request and ensuring it was correctly added to the Dynamo tables.

## Images

image


twrichards (approved):

SHession:
Made some additional changes to fix an issue with builds on clean agents. When doing a clean build it, there were problems generating the zip file as the target directory was missing. This is because it isn't checked in to VC. I have added a `mkdir` step to the yarn script to resolve this issue.

twrichards (approved):

PR #1005, opened 2021-05-27
Adds logging to outgoing YouTube requests

SHession:
## What does this change?

This PR attempts to improve YouTube request logging, allowing us to more easily diagnose any issues with our requests. We had hoped to simply adjust the logback configuration to log "INFO" level but the output was not especially helpful. This approach adds a `YouTubeRequestInitializer` to log the import details of outgoing requests. Examples of the logs can be found here: https://logs.gutools.co.uk/s/editorial-tools/goto/0df56431b947b393ccf1531ece20c2b2

image

## How to test

Deploy changes to CODE and make changes which would result in YouTube requests. Check the log messages produced are helpful enough.

We are better able to determine if issues with the YouTube APIs are on the MAM or YouTube side.

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubeAccess.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubeAccess.scala
@@ -1,17 +1,19 @@
 package com.gu.media.youtube
 
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
+import com.google.api.client.http.{HttpRequest, HttpRequestInitializer}
 import com.google.api.client.http.javanet.NetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
-import com.google.api.services.youtube.{YouTubeScopes, YouTube => YouTubeClient}
+import com.google.api.services.youtube.{YouTubeRequest, YouTubeRequestInitializer, YouTubeScopes, YouTube => YouTubeClient}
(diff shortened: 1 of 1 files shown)


common/src/main/scala/com/gu/media/youtube/YouTubeAccess.scala
```diff
@@ -49,10 +51,23 @@ trait YouTubeAccess extends Settings {
     .build
     .setRefreshToken(refreshToken)
 
+  private val youTubeRequestLogger = new YouTubeRequestInitializer{
+    override def initializeYouTubeRequest(request: YouTubeRequest[_]): Unit = {
+      super.initializeYouTubeRequest(request)
+      val markers: LogstashMarker = Markers.appendEntries(Map(
```
mbarton, line 57: Should we log `POST` body parameters as well? Only if they are used by the YouTube SDK and are readable of course

SHession: I've tried to log everything in the [YouTubeRequest](https://developers.google.com/resources/api-libraries/documentation/youtube/v3/java/latest/com/google/api/services/youtube/YouTubeRequest.html#getPrettyPrint--) that looks useful, we're logging the `httpContent`, if there something I've missed I'm happy to add it.

mbarton (approved):

PR #1007, opened 2021-06-02
Update Node version

tjsilver:
Co-authored-by: @AWare 

## What does this change?
1. sets Node to version 14 in the app
2. sets `pluto-message-ingester` Node to version 14 and reinstates pluto .nvmrc (deleted in https://github.com/guardian/media-atom-maker/pull/995)
3. splits out the CI steps to run in Docker.

## How to test
Deploy build 4352 to CODE and test Pluto messages. This was tested with @fredex42 and works as expected.


AWare (approved):
pluto is happy with node 14

scripts/pluto-ci.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd pluto-message-ingestion
+
+# The pluto lambda uses a different version of node from main app
+# so we run it separately
+
+yarn
```
akash1810, line 10: You'll want to install dependencies in CI with `--frozen-lockfile` to ensure a deterministic build.

Similarly in `scripts/app-ci.sh` above.

From https://classic.yarnpkg.com/en/docs/cli/install/:

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass `--frozen-lockfile` flag.

AWare: This wasn't present in the original scripts, we'll do this in a new pr.

PR #1008, opened 2021-06-04
Delete teamcity.sh

tjsilver:
## What does this change?

Following from #1007, this PR removes the old `teamcity.sh` script which has been replaced by the three separate scripts.


akash1810:
I'd strongly advocate to keeping a single CI script as it makes it that much easier to reproduce CI issues locally, in that you just run the one file rather than remembering which files CI is running and in which order.

Rather than delete `teamcity.sh` I'd suggest updating it to call out to the other scripts added in #1007 ,

tjsilver:
> I'd strongly advocate for keeping a single CI script as it makes it that much easier to reproduce CI issues locally, in that you just run the one file rather than remembering which files CI is running and in which order.
> 
> Rather than delete `teamcity.sh` I'd suggest updating it to call out to the other scripts added in #1007 and continue to have CI execute `teamcity.sh` only.

That sounds sensible, will do that 👍

tjsilver:
Closing and will create another PR with the updated single `teamcity.sh` script.

AWare:
Putting them in individual scripts allows each build step to run in its own docker environment and means we don't have to mess around with nvm on teamcity. While you're right that this makes it slightly harder to run the ci process locally- this issue is also present in all similar github actions builds and has not presented an impediment to its takeup.

twrichards (approved):

PR #1010, opened 2021-06-15
Adds requests logger to partner API client

SHession:
## What does this change?

In my [previous PR](#1005 ) I attempted to improve the logging of outgoing requests to the YouTube API, this didn't include requests to the partner API.  This PR adds similar logging to the partner API client which should allow us to investigate various oddities with claims we are seeing in PROD.

## How to test

Deploy the changes to CODE, make a change which result in a call to the partner API (i.e. activating a video), [observe the logs](https://logs.gutools.co.uk/goto/78054784b84cb3b65b91eddbbbdf798a) and check they tell us as much as possible.

We have greater insight into our outgoing requests which we can use to identify where issues lie.

### DIFF
--- a/common/src/main/scala/com/gu/media/youtube/YouTubeAccess.scala
+++ b/common/src/main/scala/com/gu/media/youtube/YouTubeAccess.scala
@@ -1,11 +1,10 @@
 package com.gu.media.youtube
 
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
I can see this going through with plenty of detail as a result of your recent deploy to code (looks like the content fields are logged discretely, presume this is deseriable? Shouldn't matter if we don't index them) – LGTM.

Screenshot 2021-06-15 at 10 04 00

PR #1027, opened 2021-11-11
Split CAPI and logging roles to enable us to log to our account-specific logging stream

jonathonherbert:
_co-authored-by: @dskamiotis_

## What does this change?

At the moment, the permission to put records on our CAPI and logging streams are contained within a single role, that's consumed cross-account.

We're updating our apps to log into a stream that's contained within the app's account.

As a result, we'll need to split our current role into two – keeping the original role for cross-account access to our CAPI stream, and adding a new logging role to allow access to our in-account logging stream.

This PR makes that change, splitting `stsRoleToAssume` into two configuration options, `stsCapiRoleToAssume` and `stsLoggingRoleToAssume`. We've updated our configuration and code to assume the latter role for our logging.

## How to test

Deploy the [relevant cloudformation change](https://github.com/guardian/editorial-tools-platform/tree/jsh-dm/add-account-logging-to-mam) to the CODE stack, update the CODE configuration in S3 with the new parameters, and deploy this branch. The app should continue to work and log as normal. 

Tested in CODE:

Screenshot 2021-11-11 at 11 30 36

We log to our new, account-specific stream.

### DIFF
--- a/common/src/main/scala/com/gu/media/aws/AwsAccess.scala
+++ b/common/src/main/scala/com/gu/media/aws/AwsAccess.scala
@@ -9,7 +9,6 @@ trait AwsAccess { this: Settings =>
   def readTag(tag: String): Option[String]
 
   val credentials: AwsCredentials
-  val loggingCredentials: AwsCredentials
   // To avoid renaming references everywhere
   def credsProvider: AWSCredentialsProvider = credentials.instance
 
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
Code makes sense to me. Nice and clear! 👍

PR #1030, opened 2021-11-26
Log to Fluent Bit via stdout instead of directly to Kinesis

rhystmills:
## What does this change?

This PR changes our logging setup. We no longer log directly to Kinesis, instead; Fluent Bit will pick up logs from `stdout` and send them to Kinesis. As part of this:
- `.conf` files for Fluent Bit are including in our build by `build.sbt`.
- We include new logging appenders in our `logback.xml` file.
- We remove the existing `KinesisLogging`
- (I've also specified a more specific long term support Node version.)

This PR is accompanied by a related PR in `editorial-tools-platform`.

As a side effect of this change, we now log slightly more, as our logs will include everything from `stdout` including logs that take place on the instance during build and deploy.

## How to test

1. Use the `media-atom-maker.yaml` file in our [ed-tools-platform PR](https://github.com/guardian/editorial-tools-platform/pull/578) to update the media-atom-maker-CODE stack in CloudFormation (`media-service` account). (I've already done this).
2. Ensure the AMI used for the boxes includes FluentBit. (I've done this too)
3. Deploy to CODE the latest build for this PR. (At time of writing, this has been done)
4. Check [logs.gutools.co.uk for media-atom-maker logs](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-18h,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),exists:(field:app),meta:(alias:!n,disabled:!f,index:'67604290-baa6-11e9-bea2-633437abb232',key:app,negate:!f,type:exists,value:exists)),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'67604290-baa6-11e9-bea2-633437abb232',key:app.keyword,negate:!f,params:(query:media-atom-maker),type:phrase),query:(match_phrase:(app.keyword:media-atom-maker))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'67604290-baa6-11e9-bea2-633437abb232',key:stage.keyword,negate:!f,params:(query:CODE),type:phrase),query:(match_phrase:(stage.keyword:CODE)))),index:'67604290-baa6-11e9-bea2-633437abb232',interval:auto,query:(language:kuery,query:''),sort:!())) while making changes on [CODE Media Atom Maker](https://video.code.dev-gutools.co.uk/). Any log doc with an `ami-id` field has arrived via fluentbit.

## Further info:

Along the way I logged via both Fluent Bit and the existing logger. All logs were duplicated, other than new logs that only came via Fluent Bit. This indicates that Fluent Bit is passing on all logs as intended
Screenshot 2021-11-25 at 14 10 52
:


common/src/main/scala/com/gu/media/logging/KinesisLogging.scala
```diff
@@ -1,38 +0,0 @@
...
-
-trait KinesisLogging { this: Settings with AwsAccess =>
-  private val rootLogger = LoggerFactory.getLogger(SLFLogger.ROOT_LOGGER_NAME).asInstanceOf[LogbackLogger]
-
-  def startKinesisLogging(sessionId: String): Unit = {
-    rootLogger.info("bootstrapping kinesis appender if configured correctly")
-
-    for {
-      _stack <- stack
-      loggingCredentials <- credentials.logging
```
SHession, line 18: I think we should be able to remove other code which sets up these logging creds. I.e.https://github.com/guardian/media-atom-maker/blob/e707e35627e955c7b5e5ff5e9bfc2573138eafd5/common/src/main/scala/com/gu/media/aws/AwsCredentials.scala#L55-L60

SHession (commented):
Changes look sensible, I have added a comment about the logging role, which also has implications for the CFN. I will double check some logs to confirm it's working

SHession (commented):
Spot checked various changes in CODE and tested them against comparable logs in PROD. The same kind of logs existed with the same fields. All looks good to me.

Additionally, we are now logging instance Id, which we haven't been up until now.  This is a great addition, along with the other new fields.

Holding off +1ing until the comments are addressed.

common/src/main/scala/com/gu/media/logging/KinesisLogging.scala
rhystmills: Great spot, have removed that and related code. Hopefully got everything!

SHession (commented):
Changes look great. There should be one more bit we should be able to remove with confidence, which is the Logback Kinesis Appender library.

- https://github.com/guardian/media-atom-maker/blob/e707e35627e955c7b5e5ff5e9bfc2573138eafd5/build.sbt#L35
- https://github.com/guardian/media-atom-maker/blob/e707e35627e955c7b5e5ff5e9bfc2573138eafd5/build.sbt#L103

SHession (approved):
Tested on CODE after the updates. Logs are coming through as expected 🎉

PR #1032, opened 2022-01-12
 Fix MatchError exceptions in the log 

fredex42:
## What does this change?

We found a non-exhaustive match in `PlutoUploadActions` which was causing errors in the log. It turns out that there is nothing more that needs to be done for the extra cases, so a simple do-nothing "default" case is added

## How to test

Once deployed, there should be no MatchError nor 500 reported in response to a pluto resync request

No more MatchErrors from the `PlutoUploadActions` class

n/a


davidfurey (approved):
lgtm

PR #1043, opened 2022-04-26
Allow NVM from Brew

davidfurey:
## What does this change?

Allows you to use NVM from brew.  Also avoids turning your console red permanently.

I don't know whether there was some good reason not to allow NVM in the past, perhaps @akash1810 knows?

## How to test

* Install NVM via brew
* Run `scripts/setup.sh`
* No errors

### DIFF
--- a/scripts/check-node-version.sh
+++ b/scripts/check-node-version.sh
@@ -12,15 +12,6 @@ fileExists() {
 }
 
 changeNodeVersion() {
-    BREW_NVM_SH="$(brew --prefix nvm)/nvm.sh"
-
-    if fileExists "$BREW_NVM_SH"; then
-        echo -e "${red}NVM was installed using brew, this is unsupported."
-        echo -e "Uninstall it using brew uninstall nvm and then"
-        echo -e "Install it from https://github.com/creationix/nvm#installation${plain}"
-        exit 1
-    fi
-
     if [[ -z "${NVM_DIR}" ]]; then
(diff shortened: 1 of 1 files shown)


jonathonherbert (approved):
I do feel that while our run scripts should alert users when their environments are unsuitable, having opinions about how they should get into suitable states is often problematic. For example, my using `fnm` here would cause problems.

akash1810:
> I don't know whether there was some good reason not to allow NVM in the past, perhaps @akash1810 knows?

The primary reason is `nvm` [does not officially support homebrew](https://github.com/nvm-sh/nvm#:~:text=Homebrew%20installation%20is%20not%20supported.%20If%20you%20have%20issues%20with%20homebrew%2Dinstalled%20nvm%2C%20please%20brew%20uninstall%20it%2C%20and%20install%20it%20using%20the%20instructions%20below%2C%20before%20filing%20an%20issue.
), and offer no support.

> Homebrew installation is not supported. If you have issues with homebrew-installed nvm, please brew uninstall it, and install it using the instructions below, before filing an issue.

Following @jonathonherbert 's comment, it's worth noting other repositories employ a ["bring your own" approach](https://github.com/guardian/grid/blob/746cee8b6772a3349af300233d503cca6d02eedf/dev/script/start.sh#L107-L115), which I prefer (as an `asdf` user, but previously an `fnm` user).

PR #1044, opened 2022-04-27
Simplify caching

davidfurey:
## What does this change?

Replaces Play's EhCache with memoization from Google's Guava library. Play's cache seems a bit heavyweight for some basic memoization.  

My motivation for changing this is really because migrating from EhCache in Play 2.5 to Play 2.6 will be more work than switching to a simpler solution.

### DIFF
--- a/app/di.scala
+++ b/app/di.scala
@@ -5,12 +5,11 @@ import controllers._
 import data._
 import play.api.ApplicationLoader.Context
 import play.api._
-import play.api.cache.EhCacheComponents
 import play.api.inject.DefaultApplicationLifecycle
 import play.api.libs.ws.ahc.AhcWSComponents
 import router.Routes
 import util._
-import scala.concurrent.duration._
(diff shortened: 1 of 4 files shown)


jonathonherbert:
@davidfurey this branch [has failed a deploy to code](https://riffraff.gutools.co.uk/deployment/view/f0217c9d-03a5-4f1a-907c-e69daecb0c6c) – [logs](https://logs.gutools.co.uk/s/editorial-tools/goto/0996f650-c6c8-11ec-8bf8-65e077395832) suggest there's a problem with the Memoize class as the application comes up.

jonathonherbert (approved):
Tested on CODE, can see the Youtube categories listed correctly. I haven't tested that they change correctly w/in the cache window, as I'm not sure how to do that safely – the code looks sensible so I'm approving, but happy to go further in testing if we're unsure.

PR #1046, opened 2022-05-09
Play 2.6 upgrade

davidfurey:
## What does this change?

Out of the box Play 2.6 only accepts requests with the correct Host header, which is a bit of a pain and provides little security benefit, so it has been disabled.

Play 2.6 has a very strict Content-Security-Policy, which I've replaced with something more permissive.

Play 2.6 has a CSRF filter enabled by default, so we now pass a CSRF token with all requests that mutate data.

Tested locally.


andrew-nowak:
Trying to open the video modal in composer CODE fails:

image

Console contains:

```
Refused to display 'https://video.code.dev-gutools.co.uk/' in a frame because it set 'X-Frame-Options' to 'deny'.
```

rhystmills:
> Trying to open the video modal in composer CODE fails:
> 
> image
> 
> Console contains:
> 
> ```
> Refused to display 'https://video.code.dev-gutools.co.uk/' in a frame because it set 'X-Frame-Options' to 'deny'.
> ```

I observed the same thing in a separate branch this morning so I think this may be a pre-existing issue.

davidfurey:
Thanks @andrew-nowak for spotting this and @rhystmills for reproducing.  I've fixed this issue (currently deployed to CODE).  Can you have another look at the PR?

public/video-ui/src/services/VideosApi.js
```diff
@@ -207,6 +231,9 @@ export default {
     return pandaReqwest({
       url: composerUrl,
       method: 'post',
+      headers: {
```
jonathonherbert, line 234: Do we need this header for a Composer request?

jonathonherbert (dismissed):
Tested in CODE by creating an atom, embargoing, removing the embargo, embedding in a CODE piece via the modal, publishing the piece, and publishing the atom. Not super familiar with media-atom-maker – a good opportunity to have an explore – but all worked as expected.

One risk would be missing a CSRF token – I've been over the client requests and everything I can find for `delete`, `put` and `post` is covered. There's one instance where this may be redundant.

Minor: it feels like a bit of a gotcha that we must include this header in every request within code that's specific to particular calls – an inelegant alternative might be inspecting the url passed to `pandaReqwest` to see if we're making calls to the same host with a mutating verb, and automatically including the headers. Perhaps there are better ways.

public/video-ui/src/services/VideosApi.js
davidfurey: You are quite right, too much copy+paste

jonathonherbert (approved):

PR #1056, opened 2022-06-28
Fix scheduledDate bug

rhystmills:
Co-authored with @aracho1 and @jonathonherbert 

## What does this change?

This PR fixes a bug with the `scheduledLaunch` - where editing text fields could remove the `scheduledLaunch` or `embargo` from the associated Composer article for a video. This PR fixes the issue by consistently storing the `scheduledLaunch` as a number (milliseconds since Unix epoch) rather than a `moment` or a formatted string.

The UI is updated where it was expecting anything other than a number.

The PR also fixes a bug with the management tab where all times were rendered as 00:00 due to an incorrect format string, and a bug where embargoed videos showed `Embargoed: 1st Jan 2500 00:00` rather than `Embargoed indefinitely`.

## How to test

1. Deploy to CODE.
2. Add a video and schedule it. 
3. Publish the Composer page, and check schedule is working.
4. Make changes to required text fields on the MAM page. 
5. Is the Composer schedule still present?
6. Repeat this for an embargo.

Also:
- Check the date picker is working as expected
- Check that dates are rendered properly on the front page of the app

### DIFF
--- a/public/video-ui/src/actions/VideoActions/getVideo.js
+++ b/public/video-ui/src/actions/VideoActions/getVideo.js
@@ -38,6 +38,10 @@ export function getVideo(id) {
         if (scheduledLaunch) {
           res.contentChangeDetails.scheduledLaunch.date = moment(scheduledLaunch).valueOf();
         }
+        const embargo = res?.contentChangeDetails?.embargo?.date;
+        if (embargo) {
+          res.contentChangeDetails.embargo.date = moment(embargo).valueOf();
+        }
         dispatch(receiveVideo(res));
       })
       .catch(error => dispatch(errorReceivingVideo(error)));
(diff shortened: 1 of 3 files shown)


andrew-nowak:
code looked good to me, but testing on CODE crashes with this error in the console when setting a schedule

image

rhystmills:
> code looked good to me, but testing on CODE crashes with this error in the console when picking a date while setting a schedule
> 
> image

Cheers for taking a look Andrew; that's now fixed. It turns out the following date format strings are used by different libraries which follow different conventions:

```
const DATE_FORMAT = 'dd MMM yyyy';
const DATETIME_FORMAT = `DD MMM yyyy HH:mm`;
```

I've renamed them to be more descriptive, which should help explain why the latter uses `YYYY`:

```
const REACT_DATEPICKER_DATE_FORMAT = 'dd MMM yyyy';
const MOMENT_DATETIME_FORMAT = `DD MMM YYYY HH:mm`;
```

PR #1050, opened 2022-06-20
Youtube service account

davidfurey:
## What does this change?

Changes the mechanism for authenticating with Youtube.  We now use a service account associated with the media atom maker Google cloud project.  Previously we used oath credentials generated using https://github.com/akash1810/google-oauth2-token-viewer which were associated with a developer.

NB - depends on https://github.com/guardian/editorial-tools-platform/pull/615

PR #1058, opened 2022-06-29
Youtube credentials fix

davidfurey:
## What does this change?

Roles out #1050 but fetching `stage` from environment variables in lambdas to avoid:

```
{
  "errorMessage": "Missing stage ",
  "errorType": "java.lang.IllegalArgumentException",
  "stackTrace": [
    "com.gu.media.Settings$class.getMandatoryString(Settings.scala:16)",
...
```


akash1810 (dismissed):

uploader/src/main/resources/lambda-template.yaml
```diff
@@ -19,6 +19,9 @@
         CONFIG_KEY: !Sub
           - "${Stage}/media-atom-maker.private.conf"
           - { Stage: !Ref Stage }
+        CREDENTIALS_KEY: !Sub
+          - "${Stage}/youtube-service-account.json"
+          - { Stage: !Ref Stage }
```
akash1810, line 24: Does the lambda have read permission for this file?

akash1810 (approved):
Small question on IAM, but LGTM 👍🏽

uploader/src/main/resources/lambda-template.yaml
davidfurey: Yes, thanks to https://github.com/guardian/media-atom-maker/blob/2c67a5964df81baf1bc0c68314de2caa7ad8f721/uploader/src/main/resources/cfn-template.yaml#L78

davidfurey:
Looks okay on PROD. We need to remove the old config once we are confident we won't have to roll back.

PR #1059, opened 2022-06-30
Warn user when node version is wrong but don't require nvm specifically

rhystmills:
Shamelessly borrowed from https://github.com/guardian/flexible-content/pull/3813

The `nvm` script makes running the project difficult when using e.g. `fnm` (which some of us are now doing due to command line slowness caused by `nvm`). This PR changes the scripts to warn users with the wrong version of node, without forcing the use of `nvm`.

### DIFF
--- a/scripts/check-node-version.sh
+++ b/scripts/check-node-version.sh
@@ -5,34 +5,18 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 red='\x1B[0;31m'
-plain='\x1b[0m'
-
-fileExists() {
-  test -e "$1"
-}
-
-changeNodeVersion() {
-    if [[ -z "${NVM_DIR}" ]]; then
-        echo -e "${red}NVM not found. NVM is required to run this project"
-        echo -e "Install it from https://github.com/creationix/nvm#installation${plain}"
-        exit 1
-    else
-        [ -s "$NVM_DIR/nvm.sh" ] && \. "${NVM_DIR}/nvm.sh"
-        # nvm will need to be sourced
-        nvm install
-    fi
-}
 
 checkNodeMatchesNvm() {
     NODE_MAJOR_VERSION=$(node -v | cut -d "." -f 1)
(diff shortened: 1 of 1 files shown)


scripts/check-node-version.sh
```diff
@@ -5,34 +5,18 @@ set -e
...
     NODE_MAJOR_VERSION=$(node -v | cut -d "." -f 1)
     DESIRED_NODE_VERSION=$(cat "${DIR}/../.nvmrc")
 
     if [[ "v${DESIRED_NODE_VERSION}" != "${NODE_MAJOR_VERSION}"* ]]; then
-        echo -e "${red}Your node version ${NODE_MAJOR_VERSION}" does not match "${DESIRED_NODE_VERSION}"
-        echo -e "Please run 'nvm use' to get the desired node version${plain}"
+        echo -e "${red}Your node version ${NODE_MAJOR_VERSION} does not match the version the project expects (v${DESIRED_NODE_VERSION})."
+        echo -e "${red}Please ensure it matches before starting the project."
+        echo -e "${red}If you are using nvm, for example, run \`nvm use\`."
+        echo -e "${red}https://github.com/creationix/nvm"
```
davidfurey, line 17: Printing `{red}` to the terminal isn't scoped to a single command, so you only need to prefix line 14 with `{red}`.

Also, you should add `{plain}` as a suffix to line 17 to set the terminal back to normal.  Otherwise all subsequent commands that the user runs in this terminal will be red.  Which is quite irritating.

davidfurey (changes requested):
I'm happy with the idea behind this PR, but there is a bug in the terminal colours code.

Also, it would be great if we could standardise on nvm or fnm across the department (or even just the team). I don't see any advantage to us all doing our own thing.

scripts/check-node-version.sh
rhystmills: In my case (and a few others) `nvm` was causing noticeable lag on changing directory - which is why I switched. I'm not sure others have experienced this so it might be a bit difficult to encourage uptake. I agree that using consistent tools is a good thing.

davidfurey (approved):

PR #1060, opened 2022-06-30
Add TypeScript support

rhystmills:
## What does this change?

This PR adds TypeScript support to `media-atom-maker`. This will allow developers to type-check new additions and migrate existing JavaScript files to TypeScript.

The PR also moves a small JS file and a test file to TypeScript to ensure that application build and tests are both are working as intended.

## How to test

Run the application locally according to the instructions in the readme, and run the tests. Is the application building, and are the tests running?


build_config/webpack.dev.conf.js
```diff
@@ -11,6 +11,11 @@ module.exports = {
         exclude: /node_modules/,
         loader: 'babel-loader?cacheDirectory=true'
       },
+      {
```
jonathonherbert, line 14: Can I recommend introducing typechecking to our watch and build tasks? There are a couple of ways I know of –

- use [`fork-ts-checker-webpack-plugin`](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin)
- use the approach detailed by @twrichards in [this thread](https://chat.google.com/room/AAAAWaoV0IE/uozJCuL5Ej4), and use either `&` or `run-p` to run `tsc` alongside the watch process, and on build (we'll need to change the build script, and fail fast when the build fails, if that isn't already the case.)

In this way we've got the best possible view of our compiler's opinion during `watch`, and we can't ship type errors in deployables.

jonathonherbert: (More detail: `?transpileOnly=true'` ensures that ts-loader will simply strip the type information from .ts files as it converts them to js. This is great for fast builds and a good developer experience, but explicitly skipping the typecheck leaves us vulnerable to errors.)

PR #1067, opened 2022-08-17
Convert expiry from string to number, like scheduledLaunchDates

andrew-nowak:
## What does this change?

Need to test this connects correctly on CODE.

Currently, MAM sends the expiry date as a string, but composer expects to see a number. On old versions of lift-json I think this means the field gets silently ignored - on the recent upgrade attempting to parse throws.

## How to test

Deploy to CODE along with the relevant versions of composer. Do video pages get created with expiry dates?

 - [x] Tested on CODE

## Accessibility

### DIFF
--- a/public/video-ui/src/actions/VideoActions/getVideo.js
+++ b/public/video-ui/src/actions/VideoActions/getVideo.js
@@ -42,6 +42,10 @@ export function getVideo(id) {
         if (embargo) {
           res.contentChangeDetails.embargo.date = moment(embargo).valueOf();
         }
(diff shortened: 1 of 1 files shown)


davidfurey (commented):
This isn't going to be very nice when we convert to TS, but since it is following the same pattern as scheduled and embargo, it isn't making anything any worse.

davidfurey (approved):

andrew-nowak:
Tested on CODE against both the old and new versions of flexible-content, all lgtm! merging

PR #1071, opened 2022-11-09
Upgrade to panda v1.2.0

andrew-nowak:
Depends on guardian/atom-maker#63

## What does this change?

Also does other required upgrades (scala, sbt, play 2.7), fixes deprecation warnings

## How to test

Deploy to CODE and try MAM functionality

## Accessibility


rhystmills (approved):
Lovely to see!

Fixes the authentication problem in the `.start.sh` script. `client-dev.sh` still not working for me, but not a reason not to merge (was already not working).

Deployed to CODE and made some standard changes (editing fields, publishing, changing holding image) - all worked as expected.

.sbtopts
```diff
@@ -0,0 +1,5 @@
+-J-Xms512M
+-J-Xmx1536M
+-J-Xss4M
```
alinaboghiu, line 3: this changed from 1M, assuming on purpose?

common/src/main/scala/com/gu/media/aws/AwsCredentials.scala
```diff
@@ -2,7 +2,7 @@ package com.gu.media.aws
 
 import com.amazonaws.auth._
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
-import com.amazonaws.services.securitytoken.{AWSSecurityTokenServiceClient}
+import com.amazonaws.services.securitytoken.{AWSSecurityTokenServiceClient, AWSSecurityTokenServiceClientBuilder}
```
alinaboghiu, line 5:
```suggestion
import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder
```

.sbtopts
andrew-nowak: yes - I think one of the upgrades is very recursive at compilation, so I was getting stack overflows at compile time; this increases the stack so compilation can succeed

alinaboghiu (dismissed):
wonderful!

andrew-nowak:
> There's a couple of places where we still use play.api.Logger. I am guessing from this PR that there's a preference for com.gu.media.logging.Logging. Do we want to be consistent and use the latter everywhere or is there a reason we might want to stick with play.api.Logger?

woah great catch! I'd been working from deprecation warnings but several places are doing something very sneaky (accessing the underlying object from the singleton) that doesn't get picked up by the deprecation notices 😮 I'm going to clean them up now

alinaboghiu (approved):
yay

PR #1072, opened 2022-11-11
Copyable keywords

rhystmills:
## What does this change?

This PR renders the keywords as a comma-separated string list, which will allow users to copy and paste keyword lists between different videos. 

When a series of videos reference the same new event (e.g. US Midterms), video editors will want to use many shared keywords. The field already supports pasting a comma-separated list. This PR will render a comma-separated list of the keywords in a video so that a user can also copy a parseable keyword list.

It also restyles the field as previously only about 100 pixels of width were selectable.

## How to test

1. Run locally (I've found only the `/scripts/start.sh` script works currently, not `client-dev.sh`) or deploy to CODE.
2. Add keywords to a video. Try copying the bottom list of keywords into another video's keyword field, then hit enter. Does it populate the video with the same keywords?
3. Make sure other fields based on multiple tags (e.g. the byline field) are not broken

## Images

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/public/video-ui/src/components/Tags/TagFieldValue.js
+++ b/public/video-ui/src/components/Tags/TagFieldValue.js
@@ -2,6 +2,7 @@ import React from 'react';
 
 export default class TagFieldValue extends React.Component {
   renderFieldValue(value, index) {
+    console.log(value);
     if (value.detailedTitle) {
       return (
         <span key={`${value.id}-${index}`}>
@@ -23,7 +24,7 @@ export default class TagFieldValue extends React.Component {
   }
   getRenderedValues(){
     const values = this.props.tagValue.map(this.renderFieldValue);
-    if (values.every(value => typeof value === "string")){
+    if (values.every(value => typeof value === "string" && !value.includes(','))){
       // E.g. a list of YouTube keywords
       return values.join(',');
     } else return values;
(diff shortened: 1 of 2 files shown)


akash1810:
If it's useful, https://github.com/guardian/media-atom-maker/pull/857 for a previous attempt at this.

akash1810:
not blocking: would a "copy all tags to clipboard" button provide a simpler UX here?

rhystmills:
> not blocking: would a "copy all to clipboard" button provide a simpler UX here?

That was my first thought - but I figured this would be the smallest change to allow the user to easily copy and paste keywords, and seems intuitive given that the existing rendered preview of keywords (one big unbroken string) isn't useful. (Also ran the idea past Adam in the Videos team and he was happy with the solution)

jonathonherbert (approved):
Tested locally, works as expected! Avoiding touching if the text already contains commas seems sensible (e.g. custom bylines.)

Do agree that 'copy to clipboard' arguably better UX – the copy/paste action is possible in both cases, but a button makes it clearer that there's an affordance when the data is pasted, which might it more likely that users will figure out how this works without guidance.

paperboyo:
Thanks for this! I will agree this is best and easiest. Maybe one day chips will be susceptible to Cmd/Shift clicking to select and to Cmd+a/x/c. And pasting them into/from text editor will turn them back/forth into text/queries.
But I think this is easiest and provides some flexibility (being able to not select all).

PR #1077, opened 2022-11-25
Allow user to override the image for the YouTube thumbnail

rhystmills:
## What does this change?

This PR allows the user to override the image for the YouTube thumbnail, to be used instead of the poster image if specified. 

The videos team occasionally want to create a specific YouTube thumbnail that will improve video performance on that platform, but we wouldn't necessarily want to use on our site. They currently do this manually on YouTube, but updates to the Media Atom overwrite the thumbnails updated on YouTube itself.

I've modified the layout slightly to make a better use of the space, and I've also provided a description of what the images are used for as I don't think it's especially clear, especially now that we have a third option.

Some examples of custom thumbnails set in YouTube:
image

## Dependencies - all now merged
Because content atoms are stored in CAPI, this PR required a change to the thrift model and elasticsearch mappings enabled by these PRs:
- https://github.com/guardian/content-atom/pull/155 - (update the thrift definition for content atoms)
- https://github.com/guardian/content-api/pull/2721 - (update the elastic search mapping to include the youtubeOverride field and specify its shape)
- https://github.com/guardian/content-api/pull/2730 - (update capi porter to reflect the updated thrift model)
- https://github.com/guardian/content-api-models/pull/219 - (update content-api-models to reflect the change to the content atom)
- https://github.com/guardian/content-api/pull/2732 - (update comtent-api-models version used by capi concierge)
- https://github.com/guardian/atom-maker/pull/86 - (update atom-maker's content-api-models version because media-atom-maker accesses the model indirectly via atom-maker

This is a lot of PRs for a data change that won't be manifested anywhere other than YouTube - and which therefore doesn't really need to be in CAPI. If we are to support the videos team with future changes specific to YouTube, it might make sense to provide a separate database for YouTube specific information, e.g. a DynamoDB table keyed against the video id in CAPI.

(For example, the videos team would ideally like to be auto-populate channel-specific branding for channels like Guardian Sport from media-atom-maker).

## Before and after:

| Before | After |
| --- | --- |
| image | image |

## Testing
The easiest way to test this change is probably deploying to our CODE environment via RiffRaff, where media-atom-maker shows as `media-service:media-atom-maker`.

1. Add a video to YouTube via [media atom maker](https://video.code.dev-gutools.co.uk/). I've observed that thumbnails for videos with small dimensions don't tend to work properly so use a landscape video with a decent resolution.
  i. There are required fields in the 'Furniture' and 'YouTube Furniture Tab' which must be completed 
  ii. After the above, a video can be updated from the pencil icon in the top right corner of the 'Video Preview' panel (shows the tooltip 'Add Assets' on hover). 
  iii. Choose 'Upload to YouTube', upload an video you have locally on your machine and activate it
2. Back on the main video page, add a YouTube Thumbnail Image from the Grid, in addition to a Main Image.
3. Publish the video
4. Is the YouTube Thumbnail Image used as the video thumbnail rather than the Main Image?

How does the user experience feel to you? I've experienced thumbnail update delays of 10-15 minutes from YouTube which can make it feel like media-atom-maker isn't doing its job. It also doesn't feel obvious that you need to publish the video in order to apply the thumbnail.

### DIFF
--- a/app/model/commands/PublishAtomCommand.scala
+++ b/app/model/commands/PublishAtomCommand.scala
@@ -253,19 +253,22 @@ case class PublishAtomCommand(
     handleYouTubeMessages(youTubeMetadataUpdate, "YouTube Metadata Update", previewAtom, asset.id)
   }
 
+  private def setYoutubeThumbnail(atom: MediaAtom, image: Image, asset: Asset): MediaAtom = {
+    val thumbnail = atom.isOnCommercialChannel(youtube.commercialChannels) match {
+      case Some(isCommercial) if isCommercial => thumbnailGenerator.getThumbnail(image)
+      case _ => thumbnailGenerator.getBrandedThumbnail(image, atom.id)
+    }
+
+    val thumbnailUpdate = youtube.updateThumbnail(asset.id, thumbnail)
+
+    handleYouTubeMessages(thumbnailUpdate, "YouTube Thumbnail Update", atom, asset.id)
+  }
+
   private def updateYoutubeThumbnail(atom: MediaAtom, asset: Asset): Future[MediaAtom] = Future{
-    atom.posterImage match {
-      case Some(image) => {
-        val thumbnail = atom.isOnCommercialChannel(youtube.commercialChannels) match {
-          case Some(isCommercial) if isCommercial => thumbnailGenerator.getThumbnail(image)
-          case _ => thumbnailGenerator.getBrandedThumbnail(image, atom.id)
-        }
-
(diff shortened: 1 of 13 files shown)


Fweddi (approved):
Tested and works as expected. Code LGTM.

If you're worried about the delay, and the fact you need to publish / launch to update the override, I would add a small sentence to the description - something like:

```
Youtube Image Override
--
Used on YouTube when specified. (Otherwise Main Image is used).

The atom must be published for the override to take effect. Changes can take up to fifteen minutes to apply.
```

rhystmills:
Updated copy in response to Freddie's feedback.

PR #1079, opened 2022-12-05
Use focal amigo recipe

aracho1:
Use Ubuntu focal amigo recipe as bionic will become EOL in April 2023


aracho1:
Deployed to CODE and tested creating and publishing a new video, seems to work fine

andrew-nowak (approved):

PR #1096, opened 2023-02-15
Replace Scribe with ProseMirror editor

rhystmills:
The base for this PR is currently https://github.com/guardian/media-atom-maker/pull/1095

## What does this change?

This PR replaces the `Scribe`-based trail text and standfirst editors with `ProseMirror` editors as `Scribe` is deprecated and some of its packages contain vulnerabilities. This will also be more consistent with our other tools using ProseMirror (notably flexible-content).

The PR is much larger than I'd like it to be, but it turned out these editor touched a lot of areas of the ProseMirror landscape and I ended up having to implement lots of core functionality, despite there only being two small editors in the application. It would be nice to take this all out in the future and replace it with an imported implementation that meets our needs (especially as we have custom implementations in facia-tool and flexible-content, and I believe will be needed in Targeting, which still uses Scribe).

## How to test

Run `media-atom-maker` locally according to the instructions in the README or deploy to CODE.

Test the following functionality:
- Bold
- Italic
- Add and remove link
- Bullet list
- Copy from standfirst to trail text (yellow pencil button - will lose bullet lists and replace paragraph breaks with hard breaks in markup)

## Images

prosemirror-editor

## Accessibility

-   [X] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [X] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [X] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)


dskamiotis (commented):
Would it be safe to suggest that the actual PM Editor is `RichTextInput`? and that `RichTextEditor` is a wrapper?

rhystmills:
> Would it be safe to suggest that the actual PM Editor is `RichTextInput`? and that `RichTextEditor` is a wrapper?

Yeah that's a good observation. I think RichTextField for current 'editor' and RichTextEditor for current 'input' would be more accurate names.

rhystmills:
> > Would it be safe to suggest that the actual PM Editor is `RichTextInput`? and that `RichTextEditor` is a wrapper?
> 
> Yeah that's a good observation. I think RichTextField for current 'editor' and RichTextEditor for current 'input' would be more accurate names.

(they're now renamed)

public/video-ui/src/components/FormFields/RichTextEditor.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }, []);
+
+  const setContent = (tr: Transaction, node: Node): Transaction => {
+    const sel = new AllSelection(tr.doc);
+    sel.replaceWith(tr, node);
+    return tr;
+  };
+
+  useEffect(() => {
+    if (copiedValue.text && editorView && shouldAcceptCopiedText){
```
jonathonherbert, line 46: Is there any reason we couldn't do this with `value` itself?

For example – 

1. User clicks 'copy trail text from description'
2. Button updates underlying model
3. RichTextEditor component rerenders
4. Effect updates ProseMirror instance (we could diff the incoming prop against the current state if we were worried about frequent updates)

This may have already been tried! But keen to avoid the complexity here if poss.

public/video-ui/src/components/FormFields/RichTextField.tsx
```diff
@@ -0,0 +1,164 @@
...
+import RequiredForComposer from '../../constants/requiredForComposer';
+import { EditorConfig } from './richtext/config';
+import { getWords, isTooLong } from './richtext/utils/richTextHelpers';
+import { RichTextEditor } from './RichTextEditor';
+
+type EditorState = {
+  wordCount: number;
+  copiedValue: {
+    text: string | null;
+    seed: number | null;
```
jonathonherbert, line 11: Am I right in thinking that this is used more as an `updateID` than `seed`?

jonathonherbert (commented):
Point in PR well-taken – as soon as we're able to add paragraphs and lists, there's a lot to do! Totally agree that this should be a library ASAP.

A few comments.

public/video-ui/src/components/FormFields/RichTextEditor.tsx
rhystmills: I tried and failed - I'd much prefer to do that so it would be good if we could pair.

For some reason anything I tried to update the ProseMirror instance with would leave it empty, except the `pasteHTML` method. I assume there was some subtle error in my implementation but after several hours I ended up with this - it works but it's indirect and messy.

public/video-ui/src/components/FormFields/RichTextField.tsx
rhystmills: Yeah - though as alluded above I hope we can get rid of it completely.

PR #1099, opened 2023-02-20
Bump AWS SDK

rhystmills:
Bump AWS SDK in `pluto-message-ingestion` to fix vulnerability.

### DIFF
--- a/pluto-message-ingestion/package.json
+++ b/pluto-message-ingestion/package.json
@@ -20,7 +20,7 @@
   },
   "devDependencies": {
     "@vercel/ncc": "^0.28.5",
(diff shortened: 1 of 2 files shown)


davidfurey:
Rebuilding the yarn.lock effectively bumps all dependencies to the highest allowed by semver. You might instead want to try running `yarn upgrade aws-sdk@^2.814.0` which should update package.json and yarn.lock

rhystmills:
Cheers, I will probably close this PR as it is - it turns out it's the inherited aws-sdk version from `guardian/elk-kinesis-logger` that's the problem rather than the actual dependency.

PR #1100, opened 2023-03-03
Add `/live/` urls to youtube url parser

andrew-nowak:
## What does this change?

Looks like a new format added to youtube, this is what you get if you click "share" on a livestream page now

eg. `https://youtube.com/live/abcdefghijk?feature=share`

also handles missing `www.` and tightens `.` in regex to match exactly

we could also think about adding `studio.youtube` links, leaving that for a future think though


twrichards (approved):
LGTM

👏 rapid work

PR #1107, opened 2023-05-15
Fix invalid embargoed state

davidfurey:
## What does this change?

Should make it impossible for someone to modify an atom without having visibility of the previous most-recent version.  This should avoid the issue we've seen before:

* Video is embargoed
* User A and B open video in MAM
* User A removes embargo
* User A publishes the video
* User B edits the headline [which unintentionally re-adds the embargo]
* Content is now Published but also has an embargo so cannot be re-published. Embargo cannot be removed because UI hides this feature on published content.

The change sets revision ID based on what the client has sent, not what we've just fetched from dynamodb. This way, if the user has edited an old version of the atom, atom-publisher-lib will refuse to update the db. Atom-publisher-lib will only publish if the new revision is higher than the current revision.

### Test 1

Edit a video in MAM - check that new revision is exactly 1 higher than the old revision

### Test 2

* Open the same video in 2 tabs.
* Make a change in one tab.
* Attempt to make a change in another tab - change should fail and an informative error message displayed

### Test 3

* Open the same video
* Make a change
* Make a change
* No error should be seen

This test was added because I noticed a bug with the first attempt where subsequent edits were rejected because the local revision was not updated.

### DIFF
--- a/app/model/commands/CommandException.scala
+++ b/app/model/commands/CommandException.scala
@@ -39,6 +39,7 @@ object CommandExceptions extends Results {
   def commandExceptionAsResult: PartialFunction[Throwable, Result] = {
     case CommandException(msg, 400) => BadRequest(msg)
     case CommandException(msg, 404) => NotFound(msg)
+    case CommandException(msg, 409) => Conflict(msg)
     case CommandException(msg, 500) => InternalServerError(msg)
     case YouTubeError(msg, true) => InternalServerError(msg)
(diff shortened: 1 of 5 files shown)


rhystmills (approved):
Code looks good and worked as expected when running locally (preventing an update when I had saved an edit in the meantime in a separate tab).

I think ideally the edit panel wouldn't close when the save has failed as it can give the impression that the save has succeeded (for example in my case I already had an error banner visible related to usages so I didn't notice it switching to the new error banner, and initially thought the save had succeeded).

I wouldn't block the PR for this though as the change is an improvement on the existing state of the app.

PR #1108, opened 2023-05-15
Only update the embargo/expiry/scheduler user on change

davidfurey:
## What does this change?

Only update the embargo/expiry/scheduler user on change. Current behaviour makes it look like whoever touched the piece last set the expiry/embargo/schedule.  It is probably more useful to know the original setter

## How to test

* Set an embargo on a piece of content
* Ask someone else to make a change to that piece of content
* Check dynamodb (or look in network tab) to verify that the ChangeData still lists you as the embargo-er


davidfurey:
Tested in CODE. David applied an embargo, Rhys later edited the content, but the embargo is still attributed to David.

Screenshot 2023-05-16 at 11 26 34

rhystmills (approved):
Tested on CODE and worked as expected (embargo didn't appear to be set by another user when that user changed unrelated video content).

PR #1109, opened 2023-05-15
Add more typescript types

davidfurey:
## What does this change?

Adds types to the VideoAPI. I thought I was going to need to understand how revisions worked, but it turned out that my changes were all server side. But since I'd added the types, I thought they might be useful for someone in future.

The most interesting commit is probably "Fix unexpected behaviour".

NB- types are based on a fairly cursory look at the Scala API so should probably come with a health warning.


rhystmills (approved):
All looks good to me 👍 Usages look to be working as expected on CODE.

PR #1111, opened 2023-05-15
use vertical crop for vertical video

marjisound:
## What does this change?
This PR is using 9:16 crop for thumbnail images if the video page has a `vertical-video` series tag.

image

In the upcoming PRs (In the near future) we will be using the youtube video response that gives us the aspect ratio, and pass that data through the CAPI data, in order to avoid relying on the tag. 

## How to test

- Upload a vertical video
- add the `vertical-video` series tag
- add thumbnail image
- crop thumbnail image - there should only be 9:16 crop available


abeddow91 (approved):
Looks good to me!

twrichards (approved):
fairly minor suggestion, but otherwise code looks good and looks like you've tested and approved within your team, so please go ahead when you're ready

PR #1112, opened 2023-05-24
Bump some packages to sort vulnerabilities

rhystmills:
## What does this change?

This bumps some packages to solve some high-priority vulnerabilities. There should be no changes in functionality.

## How to test

1. Run the application locally according to the instructions in the readme and check is builds as expected.
2. Run the tests to check they all pass (`sbt test` from the root)

High-priority vulnerabilities reported as solved in Snyk:

image


davidfurey (approved):
lgtm. I haven't run it locally, but I presume you have, and TC says the tests pass

rhystmills:
Thanks, yes it built and ran fine locally.

PR #1114, opened 2023-06-05
Replace elk-kinesis-logger with console log/error

rhystmills:
## What does this change?

This replaces the `elk-kinesis-logger` with standard node `console.log`/`console.error`, since we now get Lambda log-shipping for free from [guardian/cloudwatch-logs-management](https://github.com/guardian/cloudwatch-logs-management. Removing the `elk-kinesis-logger` dependency will sort two vulnerabilities in the Lambda.

This PR also fixes an issue with non-JSON payloads reaching the Lambda. Formerly, the Lambda would return an error response in these circumstances - the faulty record would remain in the Kinesis stream, and subsequent additions to the stream would not invoke the Lambda until the faulty record had expired after (I think) 24 hours, while the Lambda would be continually invoked for the faulty payload.

Payloads that aren't now parsed will log an error, but the Lambda will still complete, allowing subsequent messages to be invoked (after all, repeated attempts to parse the non-JSON would also fail).

## How to test

Pluto CODE is down since the incident in December, but we can still deploy the CODE lambda and manually insert our own data into the Kinesis stream, in order to do so:

1. Create a directory with an `example.json` file in it. This should contain data along the lines of:
```{"type":"project-updated","id":"foo","title":"bar","status":"New","commissionId":"baz","commissionTitle":"baz","productionOffice":"UK","created":"2017-04-24T17:20:11.503Z"}```
2. Set a base-64 encoded version of that text to a bash variable, e.g: `EXAMPLE=$(cat ./example.json | base64)`
3. Deploy this branch to CODE (you can click 'Preview' in RiffRaff to deploy only the `pluto-message-ingestion` `updateLambda` and `uploadLambda` steps, which deploy very quickly). 
4. Add the message to the stream via the following bash command, where CodeStreamName is your name. The partition key can be anything:
```
aws kinesis put-record --stream-name CodeStreamName --data $EXAMPLE --partition-key foobar --profile media-service --region eu-west-1
```
5. Monitor the logs [here](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-10m,to:now))&_a=(columns:!(),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'67604290-baa6-11e9-bea2-633437abb232',key:stage,negate:!f,params:(query:CODE),type:phrase),query:(match_phrase:(stage:CODE)))),index:'67604290-baa6-11e9-bea2-633437abb232',interval:auto,query:(language:lucene,query:'app:%22pluto%22'),sort:!(!('@timestamp',desc)))). Do you see a PUT command logged for your JSON payload?

You may also want to try sending something that isn't JSON as your payload to see if that functionality is working as expected.

I've tested both these paths on CODE and they worked as expected.

Valid JSON payloads lead to a PUT command and faulty JSON don't clog up the Kinesis stream.

Pluto CODE isn't running currently so we can't end-to-end testing before deployment, though Pluto PROD is. If something goes wrong and we disrupt `media-atom-maker`'s commission system, we will need to roll back this PR.

### DIFF
--- a/pluto-message-ingestion/hmac-request.js
+++ b/pluto-message-ingestion/hmac-request.js
@@ -3,10 +3,9 @@ const reqwest = require('reqwest');
 const url = require('url');
 
 class HMACRequest {
-  constructor({ serviceName, secret, logger }) {
+  constructor({ serviceName, secret }) {
     this.serviceName = serviceName;
     this.secret = secret;
-    this.logger = logger;
   }
 
   _getToken(remoteUrl, date) {
@@ -38,7 +37,7 @@ class HMACRequest {
       requestBody.data = JSON.stringify(data);
     }
 
-    this.logger.log(`Making ${method} request to ${remoteUrl}`, data);
+    console.log({message: `Making ${method} request to ${remoteUrl}`, data});
 
     return reqwest(requestBody);
   }
--- a/pluto-message-ingestion/hmac-request.js
+++ b/pluto-message-ingestion/hmac-request.js
@@ -37,7 +37,8 @@ class HMACRequest {
       requestBody.data = JSON.stringify(data);
     }
 
-    console.log({message: `Making ${method} request to ${remoteUrl}`, data});
+    // eslint-disable-next-line no-console
+    console.log({ message: `Making ${method} request to ${remoteUrl}`, data });
 
     return reqwest(requestBody);
   }
--- a/pluto-message-ingestion/index.js
+++ b/pluto-message-ingestion/index.js
@@ -23,7 +23,11 @@ exports.handler = (event, context, callback) => {
           const jsonPayload = JSON.parse(payload);
           kinesisMessageProcessor.process(jsonPayload);
         } catch (e) {
-          callback(`Failed to parse kinesis message as JSON: ${payload}`);
+          // eslint-disable-next-line no-console
+          console.error(
+            `Failed to parse kinesis message as JSON: ${payload}. Removing from the stream.`
+          );
+          callback(null, 'Message removed from stream due to invalid data.');
         }
       });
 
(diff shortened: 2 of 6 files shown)


jonathonherbert:
Code looks great! Two minor things –

- thanks for the detail test instructions – I didn't need the base64 encoding (the lambda wasn't sure what to do with base64 input), but catting the file straight into the data parameter worked a treat.
- the logs are present, but the markers are not, I think b/c we must encode our data as JSON to have it parsed by the lambda. [Everything gets added to the `message` field.](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/67604290-baa6-11e9-bea2-633437abb232/logstash-ed-tools-media-service-2023.06.15?id=PyMOvogBFgg4Hh9o6lX3) 
 
Perhaps we could keep a `logger` construct and have it handle the serialisation so we don't need to stringify logged objects every time?

rhystmills:
> Code looks great! Two minor things –
> 
> * thanks for the detail test instructions – I didn't need the base64 encoding (the lambda wasn't sure what to do with base64 input), but catting the file straight into the data parameter worked a treat.
> * the logs are present, but the markers are not, I think b/c we must encode our data as JSON to have it parsed by the lambda. [Everything gets added to the `message` field.](https://logs.gutools.co.uk/s/editorial-tools/app/discover#/doc/67604290-baa6-11e9-bea2-633437abb232/logstash-ed-tools-media-service-2023.06.15?id=PyMOvogBFgg4Hh9o6lX3)
> 
> Perhaps we could keep a `logger` construct and have it handle the serialisation so we don't need to stringify logged objects every time?

Well spotted and good suggestion with the logger, should now be sorted. Not sure what's causing the discrepancy with base64 thing as I'm still encoding on my end, curious...

rhystmills:
Example log: 
image

jonathonherbert (approved):

PR #1123, opened 2023-07-13
Upgrade to panda v1.3.0

andrew-nowak:
## What does this change?

See https://github.com/guardian/pan-domain-authentication/releases/tag/v1.3.0

## How to test

Auth continues to function

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1124, opened 2023-07-13
Revert "Upgrade to panda v1.3.0"

rhystmills:
Reverts guardian/media-atom-maker#1123 which appears to have led to some issues in the tool, with 'could not get uploads' and 'could not get youtube categories' error messages in the UI and some `500` status network errors showing up.


andrew-nowak (approved):
reluctantly lgtm

PR #1129, opened 2023-08-04
Regenerate youtubePartner API client lib; upgrade to google-api-client 2.x

andrew-nowak:
## What does this change?

Previous attempt to pull in an updated pan-domain-auth #1123 failed because it introduced an updated google-api-client at 2.x. Normally this wouldn't be a problem - we'd update MAM's usage of google clients also to 2.x to resolve the incompatibility and move on with our lives.

Unfortunately MAM uses the semi-private youtubePartner API, which doesn't have a published library. In the past, Google would generate one and host it on their docs pages, but some time in the past few years they've stopped doing so, which would threaten our ability to upgrade to newer versions of the APIs.

Fortunately, google API clients are generated from specifications by scripts - the script is available publicly on github, the spec is available publicly on their APIs, which means we can generate our own! This PR adds a script/docker image to do exactly that, then commits the output from running today, alongside upgrading all of the other Google client libraries to make sure we're in sync, as well as reintroducing #1123 

## How to test

Deploy to CODE, and make sure to do various operations which exercise the various google APIs - get an auth token from panda, view some previously uploaded videos, make some edits, upload a new video etc.

This is only indrectly documented by google - some links are here:
https://developers.google.com/discovery/v1/building-a-client-library
https://github.com/googleapis/google-api-java-client-services/tree/main/generator


rtyley:
I've rebased this PR on main, and we can see that it does well with the test from #1146 - @rhystmills & I are planning to roll this out in a pairing session shortly!

rhystmills (approved):
Ran through code with @rtyley and tested video upload & auth in CODE - all working nicely 👍

andrew-nowak:
I just checked the generator script one more time to check it still works 6 months later (not committing the result though) - it failed on creating the javadocs, but I don't think the javadocs are worth failing on, and not really necessary in these jars anyhow so might save us some repo size anyhow

PR #1137, opened 2023-12-06
Increase the overhead available to the thumbnail generator when picking images from grid

andrew-nowak:
youtube requires thumbnails to be < 2MB. when picking which grid crop resize to use, we pick those under 1.8MB, so that we have 200kb available for the addition of the branding logo.
today we've encountered an image with a 1.75MB crop that is > 2MB after logo has been added, so dropping that limit by another 100kb to unblock that image.
(we could/should test what the limit really should be, rather than guess and trial & error, but that's a problem for another day)


paperboyo (approved):
Practical. Thank you!

PR #1140, opened 2024-01-30
Upgrade to Play v2.9

rtyley:
This upgrade from Play 2.8 → 2.9 was prompted by https://github.com/guardian/atom-maker/pull/94, but it's a good idea anyway!

Note that this PR is using a preview release of https://github.com/guardian/atom-maker/pull/94 (which updates those libraries to compile against Play 2.9), which should be merged before this PR.

Preparation for this PR included several prior PRs:

* https://github.com/guardian/media-atom-maker/pull/1147 and its [supporting PRs](https://github.com/guardian/media-atom-maker/issues?q=label%3A%22Scala+2.13%22)
* https://github.com/guardian/media-atom-maker/pull/1149 - [Play 2.9 requires Java 11, 17, or 21](https://www.playframework.com/documentation/2.9.x/Requirements#Play-Requirements)
* https://github.com/guardian/media-atom-maker/pull/1145

PR #1141, opened 2024-01-30
Drop unmaintained `diff` library for logging `MediaAtom` changes

rtyley:
This removes the unmaintained [bizzabo/diff](https://github.com/bizzabo/diff) library ([`"ai.x" %% "diff"`](https://index.scala-lang.org/bizzabo/diff/artifacts/diff)) from media-atom-maker, as a prelude to the Scala 2.13 upgrade in https://github.com/guardian/media-atom-maker/pull/1140 (the `diff` library is only available for Scala 2.11 & 2.12).

The library is _only_ used for the media-atom-maker audit log, logging a diff when certain fields on `MediaAtom` entities change. The diff text _only_ goes to [the ELK logs](https://logs.gutools.co.uk/s/editorial-tools/goto/ecc536c0-c023-11ee-87d0-13c8789b7f4e) in the free-format message & description fields:

image

### Precise format of the Audit Log diff is not crucial

Originally, there was some functionality in the media-atom-maker UI involving an audit button, but this was removed with https://github.com/guardian/media-atom-maker/pull/759 in February 2018, and the ELK became the place this information was sent to (the _only_ place) with https://github.com/guardian/media-atom-maker/pull/767. See also https://github.com/guardian/media-atom-maker/pull/868#issuecomment-480065134 for a bit of additional context.

All this means that **the exact format of the diff logged to the ELK is not that important** - right now, the information could only really be used by a developer as an informative diagnostic - there is no process that cares if the diff format changes. This is probably a good thing, as 060adabbb8390cf1b0677edaba2bd0488437f83e in this PR (which captures the behaviour of the old diff by adding a unit test) shows that the format of the diff was not perfect!

### Replacement

To replace the old `diff` library, we looked at using alternative diffing libraries ([`diffx`](https://diffx-scala.readthedocs.io/en/latest/) or [`difflicious`](https://github.com/jatcwang/difflicious)) to produce the diff, but getting them to produce the format we wanted (a one-line diff with only _changed_ fields taken from an accept-list, and no colouring) looked like it would take a bit of work - @JamieB-gu pointed out that, given that there were only a few fields to look at, we could write our own diffing logic with little effort! The new logic produces output that is aesthetically better, I think - see [the updated unit test](https://github.com/guardian/media-atom-maker/pull/1141/commits/f60f0b77980b99ec35592d376c178ad1593d044c#diff-064da8baf7dc8bc5eddc4d9d23fddc171ca02aa1d16e756d1bc0ae68ed748b11) for some examples of how the audit log message changes.

### DIFF
--- a/app/model/commands/UpdateAtomCommand.scala
+++ b/app/model/commands/UpdateAtomCommand.scala
@@ -11,6 +11,7 @@ import com.gu.media.util.MediaAtomImplicits
 import com.gu.pandomainauth.model.{User => PandaUser}
 import data.DataStores
 import model.commands.CommandExceptions._
+import model.commands.UpdateAtomCommand.createDiffString
 import org.joda.time.DateTime
 import util.AWSConfig
 
@@ -120,7 +121,9 @@ case class UpdateAtomCommand(id: String, atom: MediaAtom, override val stores: D
     val message = AtomAssignedProjectMessage.build(newAtom)
     plutoActions.sendToPluto(message)
   }
+}
 
+object UpdateAtomCommand {
   private val interestingFields = List("title", "category", "description", "duration", "source", "youtubeCategoryId", "license", "commentsEnabled", "channelId", "legallySensitive")
 
   // We don't use HTTP patch so diffing has to be done manually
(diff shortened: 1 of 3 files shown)


rebecca-thompson (approved):
I haven't tested this, but the code looks good 👍

rtyley:
This was deployed at 11:03am today, the updated log messages look to be working well:

image

PR #1145, opened 2024-02-07
Pre-Play 2.9: Upgrade pan-domain-authentication (Panda) library

rtyley:
This is in preparation for the Play 2.9 upgrade in https://github.com/guardian/media-atom-maker/pull/1140 - the current Panda artifacts used actually _do_ have Scala 2.13 versions at the current artifact version number:

- https://repo1.maven.org/maven2/com/gu/pan-domain-auth-play_2-8_2.13/1.2.0/
- https://repo1.maven.org/maven2/com/gu/panda-hmac-play_2-8_2.13/2.1.0/

...but they _don't_ have Play 2.9 versions, not at this artifact number - so for https://github.com/guardian/media-atom-maker/pull/1140, we will need to update the Panda artifact version number, and it makes sense to separate this change out to make https://github.com/guardian/media-atom-maker/pull/1140 smaller.

We have tried upgrading Panda in this project back in July 2023 with https://github.com/guardian/media-atom-maker/pull/1123, but as @andrew-nowak points out below, a mismatch in the Google client libraries meant the change needed to be rolled back - now that we have https://github.com/guardian/media-atom-maker/pull/1129 to give updated YouTube partner client libraries, and https://github.com/guardian/media-atom-maker/pull/1146 to check for compatibility in those libraries, we should be in a less risky state for this upgrade.

Note that the `panda-hmac-play` artifacts have been moved out of the archived [guardian/panda-hmac repo](https://github.com/guardian/panda-hmac) repo, with the last version there being v2.2.0 released in June 2023. They are now in the main [guardian/pan-domain-authentication](https://github.com/guardian/pan-domain-authentication) repo, and releases are now done with the other panda artifacts - so version numbers of all the panda artifacts, including panda-hmac, are now aligned.

## Testing

[Deployed to CODE](https://riffraff.gutools.co.uk/deployment/view/e7bf37cc-b10a-4408-8713-f802a70c8e0f) at https://video.code.dev-gutools.co.uk/videos, if I delete Panda's [`gutoolsAuth-assym`](https://github.com/guardian/pan-domain-authentication/blob/2e25b662a5eb8f4fcd72734d63d6a9b95995075c/README.md?plain=1#L297) cookie and refresh the page, you can see that the Panda cookie is regenerated.

https://github.com/guardian/media-atom-maker/assets/52038/564a75e5-231b-4f24-9072-8ecdbdb8f9f0

[...]

PR #1146, opened 2024-02-08
Add unit test for `YouTube` client initialisation

rtyley:
In the past, we've had runtime failures when we've updated dependencies (like Panda in PR https://github.com/guardian/media-atom-maker/pull/1123) that update the version of google-api-client library used, without also updating the YouTube client SDK to match (note, for the reasons below, it can be hard to get updated compatible versions of the YouTube client).

The YouTube client specifically checks whether it is using an expected version of the google-api-client library, and will throw an exception like this if its requirements aren't met:

```
java.lang.IllegalStateException: You are currently running with version 2.1.2 of google-api-client. You need at least version 1.15 of google-api-client to run version 1.22.0 of the YouTube Data API library.
```

This change adds a unit test that just exercises the constructor of both the `YouTube` & `YouTubePartner` clients, and which will fail in the same way the client did at runtime - but, better for our purposes, it will fail the unit test, not the running app server!

An example of the test failing, and thus notifying of us a problem as we need, is [here](https://github.com/guardian/media-atom-maker/actions/runs/7828479116/job/21358445038#step:8:3721), against PR #1145.

## Getting updated compatible versions of the YouTube clients

Media Atom Maker uses the semi-private YouTube content partner API. From an SDK point of view, this involves two particular clients, which we started using as of https://github.com/guardian/media-atom-maker/pull/399 in April 2017:

* `com.google.api.services.youtube.YouTube` - a _modified_ version of the standard client, with additional fields (like `onBehalfOfContentOwner`) that are visible in Google's APIs Discovery Service, but _not_ typically included in the standard [`google-api-services-youtube`](https://central.sonatype.com/artifact/com.google.apis/google-api-services-youtube/versions) client SDK published to Maven Central.
* `com.google.api.services.youtubePartner.YouTubePartner` - an _additional_ client, that hits a different API endpoint and has different actions entirely from the standard `YouTube` client (claims, etc).

In the past, Google _would_ share special jars for both clients, which were hosted on their docs pages at:

https://developers.google.com/youtube/partner/client_libraries

However, that page now links off to just the standard versions of the Google API Client Library for Java - with the unmodified `YouTube` client, and no `YouTubePartner` at all.

Consequently, @andrew-nowak has produced https://github.com/guardian/media-atom-maker/pull/1129, which scripts the production of the custom Content Partner API clients.

See also:

* https://github.com/guardian/media-atom-maker/pull/1145#issuecomment-1931940022
* https://github.com/guardian/media-atom-maker/pull/1124

### DIFF
--- a/test/YouTubeSDKCompatibilityCheckSpec.scala
+++ b/test/YouTubeSDKCompatibilityCheckSpec.scala
@@ -1,0 +1,24 @@
+import com.google.api.client.googleapis.testing.auth.oauth2.MockGoogleCredential
+import com.google.api.client.http.javanet.NetHttpTransport
+import com.google.api.client.http.{HttpRequestInitializer, HttpTransport}
+import com.google.api.client.json.JsonFactory
+import com.google.api.client.json.gson.GsonFactory
+import org.scalatest.FunSpec
+
+class YouTubeSDKCompatibilityCheckSpec extends FunSpec {
+
+  val transport: HttpTransport = new NetHttpTransport()
+  val jsonFactory: JsonFactory = GsonFactory.getDefaultInstance
+  val httpRequestInitializer: HttpRequestInitializer = new MockGoogleCredential.Builder().build()
+
+  describe("The YouTube SDK checks for a matching version of the google-api-client library") {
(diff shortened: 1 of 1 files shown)


andrew-nowak (approved):
👏 lovely!

PR #1164, opened 2024-06-11
Bump dependencies to resolve vulnerabilities

rhystmills:
## What does this change?

This bumps some dependencies in order to reduce the number of high priority vulnerabilities reported by Snyk.

It reduces the number of Scala vulnerabilities from 6 to 2. Of the remaining two:
1. One is present in the latest release of Play, so we can't solve it for now.
2. One is present in the latest version of [panda](https://github.com/guardian/pan-domain-authentication) so we'll need to sort it there first. More details are available in Snyk.

The JS changes don't solve the vulnerabilities, but they do bring ts-loader to its latest version (where the vuln is still present).

## How to test

Deploy to CODE and see if everything appears to be working normally.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,7 @@ val scalaCheckVersion = "1.14.0" // to match ScalaTest version
 val awsLambdaCoreVersion = "1.1.0"
 val awsLambdaEventsVersion = "1.3.0"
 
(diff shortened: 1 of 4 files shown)


Georges-GNM (approved):
Checked this on code, had a bit of a look/click around, looked ok to me 👍

PR #1155, opened 2024-02-16
Add some user experience improvements to the uploads page

rhystmills:
This PR fixes some bits of awkward functionality on the uploads page, specifically:

- The video previews on the uploads are bigger, accommodate more text by default and the description bar expands to accommodate longer text. Before, they took up an oddly small amount of screen space and consequently the information about the progress through the upload process was usually unreadable.

| Before | After |
| --- | --- |
| image | image |

- Disables the 'Activate' and 'Delete' buttons until the video processing is done. Clicking on these before the video is processed wasn't functional and gave a misleading error ('Could not revert asset'). I've added a tooltip explaining why they're disabled on hover.

| Before | After |
| --- | --- |
| image | image |

- Adds a message explaining why the 'Upload to Youtube' 'Choose file' button is disabled when it is - previously it wasn't clear (though there was a suitable error message in the code that wasn't ever rendered). Also removes the hover styling when it's disabled.

| Before | After |
| --- | --- |
| image | image |

- Adds a 'not-allowed' cursor for disabled buttons.

## How to test

This PR will be simpler to test on CODE because getting the end-to-end YouTube upload working locally will be complicated.
1. Deploy this branch to CODE.
2. Fill out the 'YouTube Furniture' **without setting the Privacy status**. This will let you check the validation is showing properly in the 'Edit Assets' page.
3. Click the pencil icon in the top right, which shows a tolltip saying 'Edit Assets' on hover.
4. Is there an explanation of why you can't upload a video? On hover, does the 'choose file' button under 'Upload to YouTube' remain the same colour?
5. Go back to the YouTube furniture tab and add a privacy status, e.g. unlisted.
6. Try uploading a .mp4 video. Are the Activate and Delete buttons disabled with a tooltip until the processing is done? Can you read the various messages shown while the processing takes place?

PR #1167, opened 2024-06-21
Add some user experience improvements to the uploads page (again)

rhystmills:
Re-run of https://github.com/guardian/media-atom-maker/pull/1155 - without disabling the buttons. That PR [was reverted](https://github.com/guardian/media-atom-maker/pull/1158) because it inadvertently stopped users being able to activate YouTube live feeds, which media-atom-maker apparently supports.

The logic around this is not clear so I'm just going to introduce the more simple UI tweaks without touching this area. My guess is that YouTube live feeds never come out of the `processing` phase.

This PR fixes some bits of awkward functionality on the uploads page, specifically:

- The video previews on the uploads are bigger, accommodate more text by default and the description bar expands to accommodate longer text. Before, they took up an oddly small amount of screen space and consequently the information about the progress through the upload process was usually unreadable.

| Before | After |
| --- | --- |
| image | image |

- Adds a message explaining why the 'Upload to Youtube' 'Choose file' button is disabled when it is - previously it wasn't clear (though there was a suitable error message in the code that wasn't ever rendered). Also removes the hover styling when it's disabled.

### DIFF
--- a/public/video-ui/src/components/VideoUpload/AddAssetFromURL.js
+++ b/public/video-ui/src/components/VideoUpload/AddAssetFromURL.js
@@ -25,6 +25,12 @@ export default class AddAssetFromURL extends React.Component {
           <header className="video__detailbox__header video__detailbox__header-with-border">Asset URL</header>
           <div className="form__row">
             <div>
+              <p className="form__message form__message--warning">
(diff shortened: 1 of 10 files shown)


dblatcher (approved):
look good!

PR #1168, opened 2024-07-03
Bump Node and some packages

rhystmills:
## What does this change?

This bumps Node and some packages to sort some client-side vulnerabilities in media-atom-maker. I've replaced the deprecated [`node-sass`](https://www.npmjs.com/package/node-sass) with [`sass`](https://www.npmjs.com/search?q=sass) (the former wasn't working for me with the newer node version, it might be an M1-related issue).

After bumping node, the `ncc build` command in 'pluto-message-ingestion' stopped working. Bumping some ncc-related packages seems to have fixed this, though I've had to move from inline CLI args to an ncc config file.

I had a pop at bumping `webpack-dev-server`, but it turned out to be a time-sink getting hot module loading working in `v4`, so I have rolled that back. The [migration from v3 to v4](https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md) changes the conf structure in a lot of ways so I found it a bit complicated figuring out which change caused hot module loading not to work. 

I'd be tempted to replace Webpack with Vite instead (as we did [in Typerighter](https://github.com/guardian/typerighter/pull/279) and [facia-tool](https://github.com/search?q=repo%3Aguardian%2Ffacia-tool+vite&type=pullrequests)) so we have less conf and fewer loaders and plugins to maintain.

## How to test

The version bumps are related to build tools or testing, so if the app builds, runs and tests it's probably good to go.

Run locally:
- `./script/start.sh`
...or deploy to CODE.

Does the app build and run as expected?

Run `yarn test` and `yarn lint`. Do those both work?


rebecca-thompson (approved):
Tested locally - all works as expected

PR #1170, opened 2024-07-08
enforce media_atom_maker_access permissions

andrew-nowak:
Does what it says in the title: enforce the newly created MAM access permission.

Follow up to #1163

### DIFF
--- a/app/controllers/PanDomainAuthActions.scala
+++ b/app/controllers/PanDomainAuthActions.scala
@@ -5,10 +5,15 @@ import com.gu.media.logging.Logging
 import com.gu.pandahmac.HMACAuthActions
 import com.gu.pandomainauth.model.AuthenticatedUser
 import play.api.Configuration
+import play.api.mvc.{RequestHeader, Result}
+import play.api.mvc.Results.Forbidden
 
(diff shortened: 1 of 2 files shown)


app/controllers/PanDomainAuthActions.scala
```diff
@@ -15,20 +15,15 @@ trait PanDomainAuthActions extends HMACAuthActions with Logging {
...
-    if(!isValid) {
+    if (!isValid) {
       log.warn(s"User ${authedUser.user.email} is not valid")
-    }
-    else if (!hasAnyAtomMakerPermission) {
-      log.warn(s"User ${authedUser.user.email} does not have any MAM permissions")
-    }
-    else if (!hasBasicAccessExplicitly) {
+    } else if (!hasBasicAccess) {
       log.warn(s"User ${authedUser.user.email} does not have media_atom_maker_access permission but does have other MAM permissions")
```
bryophyta, line 23: is this the right thing to log still? It doesn't look like we're checking for other MAM permissions anymore

andrew-nowak: Thanks, sorted now

bryophyta (approved):
looks good to me, and great to add the extra user-facing message as well 👍

PR #1173, opened 2024-08-21
make permissions load from CODE, even if running locally

andrew-nowak:
MAM currently loads permissions from the same stage as it's running in, even if running locally. Most other tools will load from CODE when running locally - let's make MAM behave the same.


davidfurey (approved):

rhystmills (approved):
Tested locally and working nicely 👍

PR #1179, opened 2024-11-18
Use Vite

rhystmills:
## What does this change?

This PR moves our build and local development server to Vite from Webpack. I have done this for a few reasons:

- We have 7 (when I counted) high priority client-side vulnerabilities related to `webpack-dev-server` and other `webpack` packages - I've spent many hours unsuccessfully trying to bump that package while getting hot reloading to work, or even the current setup (page refresh on local change) and had no luck. I want to sort the vulnerabilities out.
- I'm not convinced hot reloading via Webpack is working properly. Instead, the page refreshes completely on local changes (perhaps based on the file 'watch' specified in the Webpack config), which is a bit annoying to work with. With Vite working properly, local development feels smoother and slightly faster.

| Before | After |
| --- | --- |
| Kapture 2024-11-28 at 09 58 40 copy | Kapture 2024-11-28 at 09 46 09 |

- Full builds are ~30% faster - taking 12 seconds instead of 18.
- We have lots of config for our Webpack setup - which makes major version bumps to Webpack packages difficult. We have three different conf files, and even more configuration in [webpack-dev-server.js](https://github.com/guardian/media-atom-maker/blob/main/webpack-dev-server.js). I'm fairly sure some of that config has become redundant over various package bumps in the past, and it's now very messy, especially the local dev server where config is split across two files. In this PR, **we lose ~350** lines of config and **gain only 29**.
- We end up with far fewer direct dependencies to manage, e.g. for security vulnerabilties, which is a time consuming chore, especially with so many tools to manage. We **lose 16 package**s and **gain only 2**.
- We've moved from Webpack to Vite in [facia-tool](https://github.com/guardian/facia-tool/pulls?q=is%3Apr+vite+is%3Aclosed) and [typerighter](https://github.com/guardian/typerighter/pull/279) already and everyone cheered.

## What did I change in order to do this?

I did a few things:
1. Removed all webpack config, related packages, the local websocket server on `9002` used for (faulty?) hot reloading
2. Added `vite` and some minor config, using it for build (via rollup, under the hood) and the local dev server.
3. Renamed any `.js` files containing react to end in `.jsx` to make Vite happy (it's also good practice)
4. Remove `reqwest` because it wasn't playing nicely with Vite - and we can [safely use](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) the native `fetch` api these days anyway.

The `reqwest` change is probably the one most likely to generate problems in this PR, it changes the mechanism used by all the requests in the application, and the arguments expected by `reqwest` are not the same as those expected by `fetch`.

I've tested every interaction the client-side application makes with outside services(e.g. Workflow, Targeting and usages) in our CODE environment) and they all seem to be working.

## Will there be any follow-up work?

This PR should work by itself, but there will be some tidying as a follow-up. I'd like to:
- Rename 'panda-reqwest' because we don't use `reqwest` any more.
- Remove any redundant arguments still passed to 'panda-reqwest' in invocations across the application, e.g. `crossOrigin` - and do some other tidying of requests.
- Make sure `.jsx` and `.tsx` files are being linted properly and sort any linting errors.

Also, things that predate this PR which I noticed while working on it:
- We send a request to e.g. `https://workflow.code.dev-gutools.co.uk/api/atom/undefined` on load of a video page - this is obviously unintentional, though it doesn't seem to be causing major problems.
- There are lots of 500 error to pages on the domain https://video.local.dev-gutools.co.uk/support/liveCapi locally
- Add some more client-side tests, we test almost nothing.
- Move more files to TypeScript.

## How to test
1. Run `./script/start.sh` locally. Does it start a non-hot reloading server locally? If so, build is working.
2. Run `./script/client-dev.sh` locally. Does it start a hot reloading server locally? If so, hot reloading is working.
3. Deploy to CODE. Does the app work? If so, artefact building is working.

It's also worth testing requests to external services on CODE, where things are wired up better than in local development. Do the Workflow, Targeting and YouTube integrations still work as expected (i.e. do requests in their respective tabs work properly, and does information seem to propagate to those systems?). I can help out with suggestions on testing these integrations, but mostly you'll want to check that you can create and update related settings in related tabs.

Things I tested (you might also want to):

**Furniture:**	
- [x]	Furniture text fields
-	[x]	Commissioning desks
-	[x]	Composer keywords (tags?)
-	[x]	Block ads tickbox
-	[x]	Allow comments tickbox
-	[x]	Sensitive tickbox

**YouTube**	
-	[x]	Youtube furniture text field
-	[x]	Youtube furniture dropdown
-	[x]	Youtube Tags
-	[x]	Also updates YouTube itself?
-	[x]	Youtube upload

**Workflow**	
-	[x]	Workflow text fields
-	[x]	Workflow dropdowns
-	[x]	Passes through to workflow?

**Usages:**	
-	[x]	Usages tab
-	[x]	Updates usages in real time?

**Targeting:**	
-	[x]	Targeting tab
-	[x]	Targeting rule
-	[x]	Updates targeting?

**Other:**	
-	[x]	Management tab
-	[x]	Pluto tab
-	[x]	Pluto commissions
-	[x]	Presence
-	[x]	Publish to composer
-	[x]	Composer updating on furniture change

- Build and local development still work. 
- All requests in the application still work. 
- No new error messages appear in normal usage of the app.


.eslintrc.js
```diff
@@ -17,5 +17,10 @@ module.exports = {
     'prefer-const': 'error',
     semi: ['warn', 'always'],
     'comma-dangle': ['error', 'never']
+  },
+  settings: {
+    react: {
+      version: 'detect'
```
rhystmills, line 23: Added this to sort an error when eslint was run.

rhystmills:
N.b. I would not merge this before Christmas!

Fweddi (approved):
Have tested non-hot / hot reloading locally, and tested integrations on CODE. All LGTM!

PR #1180, opened 2024-11-25
pluto-message-ingestion: remove externals from ncc config

andrew-nowak:
## What does this change?

Remove the "externals" field from pluto message ingestion's ncc config.

The externals field excludes a package from the bundled JS output. This could be used in AWS Lambda runtimes <= 16 which bundled the v2 AWS JS SDK. However 18 and above bundle the v3 SDK, so we need to either 1) stop excluding the SDK from the bundle or 2) migrate to the v3 SDK. We should probably do both (the first so we have confidence about which version of the SDK is being used at runtime, the second to keep uptodate with library updates) but for now I'm focused on upgrading the runtime, so we'll come back to the SDK upgrade at a later date.

## How to test

`cd pluto-message-ingestion; yarn; yarn build`. Before this change, the bundled output is 34kB (11kB zipped). After this change it is ~10.7MB (1.3MB zipped).

(That's a huge increase but still well within AWS limits, so we'll live with it. Moving the v3 SDK will help as the unused parts of the SDK are more easily excluded by bundlers like ncc).

Lambda can run on the Node 18 runtime.

## Accessibility

### DIFF
--- a/pluto-message-ingestion/ncc.config.json
+++ b/pluto-message-ingestion/ncc.config.json
@@ -1,7 +1,3 @@
 {
-  "externals": {
-    "aws-sdk": "aws-sdk",
-    "/aws-sdk(/.*)/": "aws-sdk$1"
-  },
   "minify": true
 }


bryophyta (approved):
lgtm! and it builds locally in the way described in the PR 👍

PR #1190, opened 2025-02-04
Store output in year/month/day not year/minute/day

davidfurey:
## What does this change?

I noticed that we had files in suspicious looking folder names like "2024/58/29/", uploaded on 29th July 2024.

This PR fixes a bug in our date format string so that this would now be stored in "2024/07/29".

I don't think the incorrect output folder actually matters beyond causing confusion.


andrew-nowak (approved):
😂 great spot!! thanks!

PR #1202, opened 2025-05-19
backport stream encryption to dev cloudformation to maintain parity with real

andrew-nowak:
## What does this change?

The pipeline cloudformation is currently not deployable to DEV, as it requires the existence of an encryption key which is not currently included in the DEV stack. Backport https://github.com/guardian/editorial-tools-platform/pull/435 to here, so we have one and can include it in the parameter list for DEV pipeline.

## How to test

Already deployed and working on DEV stacks.


bryophyta (approved):
lgtm!

PR #1205, opened 2025-06-09
add tracking pixel

dblatcher:
## What does this change?

This PR adds tracking to the media-atom-maker using the [user-telemetry tracking pixel](https://github.com/guardian/editorial-tools-user-telemetry-service?tab=readme-ov-file#tracking-pixel). This will allow us to find out more about the usage of the tool. On merge, results should be visible in the [tool audit dashboard](https://metrics.gutools.co.uk/d/aegzv59r9eghsd/tool-audit?orgId=1&var-domain=video.discussion.gutools.co.uk&var-path=All&var-interval=1d&var-chosen_apps=All).

The PR uses an '[onEnter](https://github.com/remix-run/react-router/blob/v3.2.6/docs/Glossary.md#enterhook)' hook to send the telemetry as a side effect of the a react route being opened client-side, so we get telemetry for every route/location the app serves.

## How to test

Running locally, you should see request for the pixel in the network tab on every client-side navigation between routes - the requests will fail if you do not have the telemetry service running locally.

On code, the pixel request should work and populate the code dashboard with usage data.

We have a clear idea how much the tool is being used.

### DIFF
--- a/public/video-ui/src/app.jsx
+++ b/public/video-ui/src/app.jsx
@@ -7,19 +7,10 @@ import { browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
 import { configureStore } from './util/configureStore';
 import { setStore } from './util/storeAccessor';
+import { extractConfigFromPage } from './util/config';
 import { routes } from './routes';
 
 import '../styles/main.scss';
-
-function extractConfigFromPage() {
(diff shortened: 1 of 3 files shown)


waisingyiu (approved):
Nice work, I've validated the change on CODE and it works great.

PR #1206, opened 2025-07-03
Upgrade to latest scala 2.13 and override jackson versions to 2.19

andrew-nowak:
There's a high vuln CVE on jackson-core, so upgrade to a later version. 

The jackson module versions should be kept in sync, including the scala module.

The recent scala modules require being on later versions of Scala 2.13, so we need to overcome the problem which kept us from upgrading to 2.13.12.

This is fixed by moving to the guardian fork of play-json-extensions, which has [commits to deal with that compiler bug](https://github.com/scala/bug/issues/12862#issuecomment-1799369339) - thanks @rtyley !

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -1,6 +1,6 @@
 package controllers
 
-import ai.x.play.json.Encoders._
+import com.gu.ai.x.play.json.Encoders._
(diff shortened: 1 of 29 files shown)


bryophyta (approved):
I tested [in CODE](https://riffraff.gutools.co.uk/deployment/view/ce941758-0a97-49eb-bcff-e460a65c7f32) and the app seems to be working fine afaict 👍

PR #1207, opened 2025-07-09
Constrain height in video transcoder

davidfurey:
## What does this change?

The existing transcoder constrains the video to 1280x720. MediaConvert doesn't allow us to constrain in quite the same way, but constraining the height and hoping that videos are uploaded in appropriate aspect ratios seems reasonable.

## How to test

Already tested in CODE with manual change. Main test is just whether the format of the cloudformation is correct. Deploy to CODE, check that height in mediaconvert template is set.


davidfurey:
Deployed to CODE, parameter changed as expected

Fweddi (approved):

PR #1209, opened 2025-07-14
Disable ElasticTranscoder, use MediaConverter output

Fweddi:
## What does this change?

ElasticTranscoder is reaching EOL. This PR completes the migration to MediaConverter, began in https://github.com/guardian/media-atom-maker/pull/1203.

We should hold off from merging this PR until we are satisfied that the video quality produced by each version of the transcoder is roughly the same.

## How to test

- Deploy to code
- Upload a "self hosted" video to media atom maker (you may need to give yourself this permission if you don't already have it)
- Check that the VideoPipeline state machine execution for your upload completed successfully
- Look in the uploads-origin.code.dev-guim.co.uk S3 bucket
- You should see your video in a date structured folder - this is the new V2 transcoded video
- There should _not_ be a corresponding video in a date structured folder under the "media-convert-testing" folder


gnm-notona (approved):
Makes sense.  This will remove the problem of not being able to overwrite the mp4 file on reprocessing because the V2 Transcoder can overwrite it.

PR #1217, opened 2025-07-24
Add permission for adding subtitles to self hosted atoms

abeddow91:
## What does this change?
Adds a new permission for uploading subtitles to self hosted atoms.


gnm-notona (approved):
💯

PR #1220, opened 2025-07-30
Do not silently install different dependency versions on CI, or inital setup

liamness:
## What does this change?

Changes in scripts that install npm depdencies to do so using the Yarn option `--frozen-lockfile`. This ensures that exactly the dependencies in the lockfile are installed, otherwise it could silently install different dependencies and update `yarn.lock` (which should only ever be done intentionally by a developer, with the results being committed to this repo).

Also a minor docs tweak.

## How to test

Hopefully should be possible to test locally, and by watching the CI run for this PR. Could deploy to CODE for extra safety.

Greater assurance about which NPM dependencies this project is using

Could be that the current process is somehow masking an issue which will be uncovered by doing things this way. In which case, we should update the dependencies to fix the issue in a more robust way.

## Images

Not applicable

## Accessibility

Not applicable

### DIFF
--- a/scripts/pluto-ci.sh
+++ b/scripts/pluto-ci.sh
@@ -7,5 +7,5 @@ cd pluto-message-ingestion
 # The pluto lambda uses a different version of node from main app
 # so we run it separately
 
-yarn
+yarn install --frozen-lockfile
 yarn build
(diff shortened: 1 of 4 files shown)


scripts/pluto-ci.sh
```diff
@@ -7,5 +7,5 @@ cd pluto-message-ingestion
 # The pluto lambda uses a different version of node from main app
 # so we run it separately
```
emdash-ie, line 8: Interesting though that this isn’t true anymore: maybe it could now be run together? Not for right now anyway

bryophyta (approved):
this looks good to me, thanks! 🤩

Issue #1226, opened 2025-08-04
RiffRaff Deploy Warnings


SHession:
## Problem

Currently deploys via RiffRaff report:

> riff-raff.yaml exposes the bucket name. Prefer to use bucketSsmLookup=true and store bucket name in parameter store, removing private information from version control.

<img width="1139" height="407" alt="Image" src="https://github.com/user-attachments/assets/b3fd4030-ea58-4dc1-8a42-e9a625f55f33" />

We should investigate a fix for this issue.

## Solution

We should use the `bucketSsmLookup` setting to resolve the deploy warning.

PR #1227, opened 2025-08-04
Fixes RiffRaff bucket warning by using SSM lookup

SHession:
## What does this change?

Addressing #1226, this PR uses SSM lookup to allow for deploying the project without including the bucket name in the repo. 

## How to test

Can we deploy to CODE successfully? 

No more RiffRaff deploy warnings.


waisingyiu (approved):
It looks great!

PR #1231, opened 2025-08-05
Adds a project architecture markdown doc with mermaid graph

SHession:
## Accessibility


docs/10-architecture.md
```diff
@@ -0,0 +1,130 @@
...
+
+### CAPI
+
+Media atoms are published via CAPI using the Publish Atom SNS topic.
+
+The Expirer lambda periodically checks the state of media atoms in CAPI and if they are expired updates their status in
+YouTube to private.
+
+The Scheduler lambda periodically checks the state of media atoms in CAPI and if they are scheduled for launch sends a
+request to the media atom maker backend API to publish them.
```
emdash-ie, line 114: I was confused by this at first because I didn’t understand why media atoms would already be in CAPI before they’ve launched. Looking at the code it looks like this is querying preview CAPI, which explains that – is it worth explicitly saying the lambda is querying preview CAPI here, or is that unnecessary complication?

emdash-ie (approved):

docs/10-architecture.md
```diff
@@ -0,0 +1,135 @@
+# Architecture
+
+## Graph
+
+```mermaid
+graph LR
```
emdash-ie, line 6: I spent a while trying to figure out whether we could make the diagram nodes link to the things (e.g. codebases) that they’re talking about, but unfortunately it looks like [markdown links are not supported by mermaid](https://github.com/mermaid-js/mermaid/issues/2239), and attempting to use a raw html link didn’t work in the github output either (it made a blue underline but didn’t work as a hyperlink). Trying to use mermaid locally to generate an SVG also didn’t work for me (couldn’t get mermaid installed and working). It’s a shame, I think it would’ve been nice. (I mention here only in case someone else likes the idea and manages to get it working!)

bryophyta:
>> whether it's possible to encode some of the information about what events kick off different pieces of communication

> I like this idea! I agree it might be a bit noisy, perhaps we could put together a separate diagram. Are there any events you think are especially worth illustrating?

The actions that the user can take to trigger pieces of work in the system would be very useful imo. If I understand correctly, most of the other triggers are time/schedule based, and those seem to already be covered well by annotations on the various lambdas 👍

PR #1232, opened 2025-08-06
Improve save failure behaviour by not leaving editing state

SHession:
## What does this change?

This PR intends to improve the save failure behaviour by not progressing to a non-editing event but persisting value changes in the event of a save failure. This differs from the current behaviour where a save failure appears to have worked as the fields are updated regardless. 

| Before | After |
|--------|--------|
|  |  | 

## How to test

Testable locally or on CODE. Trying editing the furniture and saving. To force a save failure delete your cookies and try saving changes.

> [!Note]
> I found testing locally still reported errors usage requests failing due to a missing CAPI key, but the functionality should otherwise work as expected. 

Less unexpected behaviour from the UI and a more consistent UX.

Think so...

### DIFF
--- a/public/video-ui/src/components/ReactApp.jsx
+++ b/public/video-ui/src/components/ReactApp.jsx
@@ -13,8 +13,7 @@ class ReactApp extends React.Component {
     }
 
     this.state = {
-      fetchedVideoFor: '',
-      errorKey: 0
+      fetchedVideoFor: ''
     };
   }
 
@@ -32,9 +31,8 @@ class ReactApp extends React.Component {
       });
     }
 
-    if(this.props.error && prevProps.saveState.saving !== this.props.saveState.saving && !this.props.saveState.saving) {
+    if(this.props.errorKey !== prevProps.errorKey ) {
       document.body.scrollIntoView({block: 'start', behavior: 'smooth'});
-      this.setState((prev) => {prev.errorKey++;});
     }
   }
 
@@ -84,7 +82,7 @@ class ReactApp extends React.Component {
         />
         {this.props.error
           ? <div
-              key={this.state.errorKey}
+              key={this.props.errorKey}
               className={`error-bar error-bar--animate`}
               dangerouslySetInnerHTML={{ __html: this.props.error }}
             />
@@ -121,6 +119,7 @@ function mapStateToProps(state) {
     video: state.video,
     publishedVideo: state.publishedVideo,
     error: state.error,
+    errorKey: state.errorKey,
     uploads: state.uploads,
     s3Upload: state.s3Upload,
     videoEditOpen: state.videoEditOpen,
--- a/public/video-ui/src/reducers/errorKeyReducer.js
+++ b/public/video-ui/src/reducers/errorKeyReducer.js
@@ -1,0 +1,12 @@
+export default function errorKeyState(
+  state = 0,
+  action
+) {
+  switch (action.type) {
+    case 'SHOW_ERROR':
+      return state + 1;
+
+    default:
+      return state;
+  }
+}
--- a/public/video-ui/src/reducers/rootReducer.js
+++ b/public/video-ui/src/reducers/rootReducer.js
@@ -2,6 +2,7 @@ import { combineReducers } from 'redux';
 import { routerReducer } from 'react-router-redux';
 import config from './configReducer';
 import error from './errorReducer';
+import errorKey from './errorKeyReducer';
 import video from './videoReducer';
 import publishedVideo from './publishedVideoReducer';
 import videos from './videosReducer';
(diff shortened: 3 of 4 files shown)


public/video-ui/src/actions/VideoActions/saveVideo.js
```diff
@@ -45,24 +45,28 @@ export function saveVideo(video) {
...
+            video,
+            usages,
+            'preview'
+          );
+        })
+          .then(() => dispatch(receiveVideoPageUpdate(video.title)))
+          .catch(error => {
+            const message = 'Could not update canonical pages';
+            dispatch(errorVideoSave(error, message));
+          });
```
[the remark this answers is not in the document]
SHession: The reason for this is we want to handle failures at different points in different ways. If we fail to save the video we reject the promise by throwing an error in the catch. If we save successfully but fail to update video usages, which can happen locally because of missing config, we only want to warn the user.

public/video-ui/src/pages/Video/index.jsx
```diff
@@ -277,11 +277,17 @@ class VideoDisplay extends React.Component {
...
-            this.saveAndUpdateVideo(video);
+            this.saveAndUpdateVideo(video)
+              .then(() => {
+                this.updateEditingState({
+                  key: 'editingFurniture',
+                  editing: false
+                });
+              })
+              .catch(() => {
+                // Keep editing state as true on save failure
```
dblatcher, line 288: Probably out of scope for this PR, but no we have access to the error, we could possibly improve the user feedback by triggering some transient messaging specific to the last click to compliment the main error bar - IE telling them what happened for the last click of the save button. 
This might be most useful for missing/invalid fields where the user might not have noticed what was wrong.

I'm thinking of a pattern something like https://mui.com/material-ui/react-snackbar/#supplementary-components

public/video-ui/src/actions/VideoActions/saveVideo.js
```diff
@@ -45,24 +45,29 @@ export function saveVideo(video) {
...
+      })
+      .catch(error => {
+        const defaultError = 'Could not save video';
+        const conflictError = 'Could not save video as another user (or tab) is currently editing this video';
+        const message = error.status === 409 ? conflictError : defaultError;
+        dispatch(errorVideoSave(error, message));
+        throw error;
+      });
```
dblatcher, line 71: I do have a concern that allowing a dispatching function to throw errors might break other usages - if I'm understanding the redux correctly, components like `PlutoProjectPicker` have this `saveVideo` passed down to them as `props.saveVideo` via `mapDispatchToProps` (for `PlutoProjectPicker`, the prop comes from its parent, `VideoUpload` ), but they don't allow for it throwing like `VideoDisplay` now does.

This wouldn't be ideal, but we could add an optional "allowThrowing" argument to `saveVideo` so it could be made to only throw the error with calling by a component that allows for it?

dblatcher (commented):
Looks good - I do have a concern about changing the behaviour of the function to allow throwing, but I'm not that familiar with the application, so this might be alright.

public/video-ui/src/actions/VideoActions/saveVideo.js
```diff
@@ -45,24 +45,29 @@ export function saveVideo(video) {
...
+            const message = 'Could not update canonical page';
+            dispatch(errorVideoSave(error, message));
+          });
+      })
+      .catch(error => {
+        const defaultError = 'Could not save video';
+        const conflictError = 'Could not save video as another user (or tab) is currently editing this video';
+        const message = error.status === 409 ? conflictError : defaultError;
+        dispatch(errorVideoSave(error, message));
+        throw error;
```
emdash-ie, line 70: Why does this function re-throw the error? The callsites catch and ignore the errors (e.g. [here](https://github.com/guardian/media-atom-maker/blob/e3ce35ab74fdfd21f8ff464ef4d4066e592389bd/public/video-ui/src/pages/Video/index.jsx#L288-L289)), so it seems like it would involve less code to ignore the error here (and have the same effect).

I guess maybe there are callsites (maybe [this](https://github.com/guardian/media-atom-maker/blob/e3ce35ab74fdfd21f8ff464ef4d4066e592389bd/public/video-ui/src/components/VideoImages/VideoImages.jsx#L20)?) which don’t ignore the errors?

SHession: The intention is to reject the promise so that the caller can handle the rejected promise appropriately. If we just catch the error the promise will resolve successfully.

The links you have provided don't work for me, could you possible update them?

vite.config.js
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
   plugins: [react()],
   publicDir: false, // This feature is deactivated because the 'public' dir also has a special meaning for the Play framework
   build: {
+    sourcemap: "inline",
```
Fweddi, line 9: What's this for?

Fweddi (approved):
Code looks reasonable to me

vite.config.js
SHession: Just useful for local dev: https://github.com/guardian/media-atom-maker/issues/931

PR #1241, opened 2025-08-07
typed pluto commissions

dblatcher:
## What does this change?

Converts the PlutoAction modules (which export the functions for creating redux Actions for fetching from Pluto Api) to typescript. 

Adds a module defining a union type describing some of the different Actions that can be dispatched into the redux reducer by those action modules (`KnownActions`) .  

Uses the new type information to convert two related jsx files to tsx.

## How to test

Should still be able to use the "Pluto" panel on the  uploads page (to reach this page, click the "edit assets" button top right of the Video Preview on a video page) to select and persist the Pluto commissions and project on the video as before. Setting the commission should still have the side effect of unsetting the project.

Running locally, the lists of available commissions and projects from the Pluto API are empty, so best tested on CODE (not aware if there's a way to have populated pluto data locally).

Improves visibility of the application data flow - the aim would be to make similar changes to other action modules to it is possible to convert more component to tsx and fix any type errors.

There are minor changes to the implementation of `PlutoProjectPicker` to address type errors that were revealed by the conversion, so should smoke-test for issues, but mostly this PR is limited to adding typing.


public/video-ui/src/actions.ts
```diff
@@ -0,0 +1,60 @@
...
+    video: Video,
+}
+
+/**
+ * A union of the Action types that dispatched from ts tiles.
+ * 
+ * This is not yet a complete list of all the actions the application handles as
+ * some actions dispatched from js files.
+ */
+type KnownAction =
```
emdash-ie, line 51: Is it worth updating this type to capture that `receivedAt` is defined on every action? I can see the argument that we haven’t typed every action yet, but I’d be quite tempted by the reduction in duplication.

public/video-ui/src/components/Pluto/PlutoProjectLink.tsx
```diff
@@ -3,7 +3,11 @@ import PropTypes from 'prop-types';
...
 
-export default class PlutoProjectLink extends React.Component {
+interface Props {
+  projectId: string
+}
+
+export default class PlutoProjectLink extends React.Component<Props> {
   static propTypes = {
     projectId: PropTypes.string.isRequired
   };
```
emdash-ie, line 13: Can we remove propTypes now?

public/video-ui/src/actions/PlutoActions/getProjects.ts
```diff
@@ -0,0 +1,40 @@
...
+function errorReceivingProjects(error: unknown): KnownAction {
+  return {
+    type: 'SHOW_ERROR',
+    message: 'Could not get Pluto Projects',
+    receivedAt: Date.now(),
+    error: error
+  };
+}
+
+export function getProjects({ commissionId }: { commissionId: string }) {
```
SHession, line 29: It might be simpler if this function just took `commissionId` as the argument and not an object.

SHession (commented):
Looks like a sensible improvement! I would be interested to spike out Redux Toolkit and see how much easier it makes things.

SHession (approved):
Tested on CODE, LGTM!

PR #1250, opened 2025-08-12
Make presence symbols circular

SHession:
## What does this change?

Small follow-on to the Presence PR, makes the symbols circular! The oval ones make me oddly nervous and imply to me something is amiss! 

| Before | After |
|--------|--------|
| image | image | 
| image | image | 
## How to test

Not sure about testing locally, but simple to test on CODE.


emdash-ie:
This also bothered me, but I hadn’t quite put my finger on it – thanks for fixing.

PR #1261, opened 2025-08-19
Removes Pluto Videos reducer as it's not used

SHession:
## What does this change?

As this functionality is not being used, it seems sensible to remove rather than migrate it: 

> // this function isn't currently being used and the module could be removed
// if there is no future work planned that would need it.

## How to test

Should have no functional impact, a quick search for plutoVideos should show no results! 

A small improvement, tidying the code up.

### DIFF
--- a/public/video-ui/src/app.jsx
+++ b/public/video-ui/src/app.jsx
@@ -5,6 +5,7 @@ import qs from 'querystringify';
 import Raven from 'raven-js';
 import { browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
+import { setConfig } from './slices/config';
 import { setupStore } from './util/setupStore';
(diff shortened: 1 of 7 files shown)


public/video-ui/src/actions/actions.ts
```diff
@@ -11,22 +8,7 @@ type ShowError = BaseAction<'SHOW_ERROR'> & {
...
- * A union of the Action types that dispatched from ts files.
- *
- * This is not yet a complete list of all the actions the application handles as
- * some actions dispatched from js files.
- */
-type KnownAction =
-    ShowError |
-    AddProjectRequest |
-    AddProjectReceive;
+type KnownAction = ShowError
```
SHession, line 12: It should be possible to remove this file after #1259 is also merged. RTK will provide us typing for our actions, so this should be unnecessary.

PR #1267, opened 2025-08-20
Media Atom Maker - minor version for subtitle uploads

gnm-notona:
## What does this change?

https://trello.com/c/lTJd3tqe/1439-mam-add-minor-version-to-processed-m3u8-files-to-reflect-subtitle-change

changes the m3u8 filenames used in S3 and the media atom dynamo db record to include a minor version that is incremented every time subtitles are changed or removed.

before url example:
https://uploads.guim.co.uk/2025/08/20/Loop%3A+Japan+fireball--1c44ce4e-760a-4312-a803-40939aeea355-2.m3u8

after url example:
https://uploads.guim.co.uk/2025/08/20/Loop%3A+Japan+fireball--1c44ce4e-760a-4312-a803-40939aeea355-2.0.m3u8

i.e. version `2` changes to `2.0`

Note that the minor version is also applied to mp4 files e.g. 
https://uploads.guim.co.uk/2025/08/20/Loop%3A+Japan+fireball--1c44ce4e-760a-4312-a803-40939aeea355-2.0.mp4

And that the minor version of the mp4 files stays at zero when subtitles are changed, because the mp4 content doesn't change.

## How to test
- Upload a self-hosted video to an atom.
- Add a subtitle file to the video
- remove a subtitle file from the video
- check in s3 uploads bucket that there are three versions of the m3u8 files, with version X.0, X.1 and X.2

### DIFF
--- a/app/util/UploadBuilder.scala
+++ b/app/util/UploadBuilder.scala
@@ -6,7 +6,6 @@ import com.gu.media.model.{MediaAtom, PlutoSyncMetadataMessage, SelfHostedAsset,
 import com.gu.media.upload.{TranscoderOutputKey, UploadPartKey}
 import com.gu.media.upload.model._
 import model.commands.CommandExceptions.AtomMissingYouTubeChannel
-import util.SubtitleUtil.getNextSubtitleVersion
 
 object UploadBuilder {
   def build(atom: MediaAtom, email: String, version: Long, request: UploadRequest, aws: AwsAccess with UploadAccess): Upload = {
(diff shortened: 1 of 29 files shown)


app/util/UploadBuilder.scala
```diff
@@ -45,28 +45,34 @@ object UploadBuilder {
...
     )
   }
 
   private def getAsset(
         selfHosted: Boolean,
         title: String,
-        id: String,
+        atomId: String,
+        version: Long,
+        subtitleVersion: Long,
```
Fweddi, line 63: is `version` the same as `majorVersion`, and `subtitleVersion` the same as `minorVersion`?

Fweddi: or does `subtitleVersion` include both major and minor versions, e.g. 4.X, whereas `version` is just major, e.g version 4.0?

Fweddi: I wonder if we should be more specific. e.g majorVersion and subtitlesMinorVersion?

gnm-notona: added commit https://github.com/guardian/media-atom-maker/pull/1267/commits/8c12df666920f22794376db0a8e698ce8c45397c to use `assetVersion` (videoVersion is too specific) and `subtitleVersion` everywhere.

PR #1268, opened 2025-08-20
Fixes the TextAreaInput scrolling to the bottom on input

SHession:
## Accessibility


public/video-ui/src/components/FormFields/TextAreaInput.jsx
```diff
@@ -1,6 +1,21 @@
...
+  // to the bottom on character input
+  shouldComponentUpdate(nextProps) {
+    return (
+      this.props.fieldValue !== nextProps.fieldValue ||
+      this.props.editable !== nextProps.editable ||
+      this.props.fieldName !== nextProps.fieldName ||
+      this.props.maxLength !== nextProps.maxLength ||
+      this.props.rows !== nextProps.rows ||
+      this.props.notification !== nextProps.notification ||
+      this.props.placeholder !== nextProps.placeholder
```
liamness, line 15: Logging out the actual props gives 20 keys (shown below), should these all be considered?

`fieldName, fieldValue, rawFieldValue, onUpdateField, editable, maxLength, notification, placeholder, touched, fieldDetails, hasError, hasWarning, displayPlaceholder, derivedFrom, maxWordLength, tagType, inputPlaceholder, tooltip, fieldLocation, updateSideEffects`

Annoyingly it looks like just wrapping in `React.memo` doesn't fix this, the issue still occurs on the first render. I know that "strict" mode will cause double rendering anyway, but that doesn't seem to be used here.

I still don't really understand _why_ this would resolve the issue but I've tested it locally and it does!

PR #1269, opened 2025-08-21
fix: replace special characters in transcoder s3 output keys

gnm-notona:
## What does this change?

https://trello.com/c/rKfd4JLJ/1455-vlc-player-cant-play-m3u8-with-special-characters

The MediaConvert video transcoder doesn't cope well with special characters (such as `:`) in the output destination (see trello card).

This fix replaces any special characters in the transcoder output keys so they can be more reliably interpreted by video players.

## How to test

- add a new atom in MAM CODE with a title that has special characters e.g. `Loop: special/unusual characters test %%`
- upload a video and wait for it to process
- look for the m3u8 file in s3 e.g. `s3://uploads-origin.code.dev-guim.co.uk/2025/08/21/Loop__special_unusual_characters_test___--705c1e3c-2d56-485c-8782-822648d30aff-1.m3u8`
- verify that the file contains relative filenames that match the accompanying file names e.g. `Loop__special_unusual_characters_test___--705c1e3c-2d56-485c-8782-822648d30aff-1hls.m3u8`
- Download all the files (m3u8s, vtt, ts) and check that the main m3u8 can be played in VLC player
- If possible, access the main m3u8 file via https and check that it plays in VLC player e.g. https://uploads.guimcode.co.uk/2025/08/21/Loop__special_unusual_characters_test___--705c1e3c-2d56-485c-8782-822648d30aff-1.mp4

image

### DIFF
--- a/common/src/main/scala/com/gu/media/upload/UploadKeys.scala
+++ b/common/src/main/scala/com/gu/media/upload/UploadKeys.scala
@@ -7,10 +7,6 @@ case class UploadKey(folder: String, id: String) {
   override def toString = s"$folder/$id"
 }
 
-case class UploadUri(bucket: String, key: String) {
-  override def toString: String = s"s3://$bucket/$key"
-}
-
 case class UploadPartKey(folder: String, id: String, part: Int) {
   override def toString = s"$folder/$id/parts/$part"
 }


domlander (approved):
Is there any _real_ risk to having duplicate filenames? e.g. `myfilename%` would have the same output as `myfilename$`, but this seems like it would be unlikely. If this were to happen, would it fail gracefully?

gnm-notona:
> Is there any _real_ risk to having duplicate filenames? e.g. `myfilename%` would have the same output as `myfilename$`, but this seems like it would be unlikely. If this were to happen, would it fail gracefully?

I don't think there's a risk of duplicates because the filenames include atom id and version as well as the title text.  What I haven't check though it what happens if you change the headline property after you've uploaded a video..  Will check.

PR #1275, opened 2025-08-27
WIP: Remove canCancel props

emdash-ie:
React is throwing a warning about these canCancel props that we’re passing via the spread operator to components that don’t support it. While I don’t think this is breaking anything, this commit removes them to reduce the noise in the console.

## Accessibility


dblatcher (approved):
Looks good! 
It seems there is no types package for the "react-tabs" package - at some point we might need to add a local declaration file for it to use in tsx files.

PR #1280, opened 2025-08-27
replace reqwest with fetch

andrew-nowak:
## What does this change?

reqwest is beyond ancient and unmaintained. modern javascripts have a pretty good http client builtin. switch to that


andrew-nowak:
replaced by #1345

PR #1283, opened 2025-09-01
Fixes CanCancel YouTube furniture console error and functionality

SHession:
## What does this change?

Introduced in #1256, this PR promoted a warning highlighting an oversight in the original fix:

Error message:
image

The YouTubeFurniture component was no propagating the prop to the EditSaveCancel component, resulting in a bug in which you can cancel a YouTube furniture change inflight: 

## How to test

Possible to the different between local and CODE. In local when a request is inflight, the cancel button should become disabled. You can slow requests using browser network throttling to see the difference. 

User's can't cancel YouTube furniture save requests mid-flight as this is likely to result in unexpected behaviour.


waisingyiu (approved):
It looks great.  I've tested it locally and it works as expected.

PR #1284, opened 2025-09-01
Fix errors from deleting self hosted video assets

waisingyiu:
## What does this change?

It was found that MAM would not let us delete the self hosted video asset. The error returned was 415 Unsupported Media Type.  This pull request fixes this problem.

There were three issues that caused the error:

1. The data models of `asset` objects are different between Youtube videos and self-hosted videos.  The logic MAM used to look up the asset to be deleted worked for Youtube, but not for self-hosted videos.  

Solution: The PR fixes [the logic](https://github.com/guardian/media-atom-maker/blob/wsy/fix-delete-self-hosted-assests/public/video-ui/src/components/VideoUpload/VideoTrail.jsx#L72-L84) it uses to locate the asset ID so it should work for both Youtube and self-hosted videos. 

2. Unlike Youtube video, each self-hosted video has two assets, the video itself and the playlist.  The backend provides an endpoint for deleting a single asset only.

Solution: While we could hit the "delete asset" endpoint multiple times, it could lead to more complicated logic in the client side as we should call it sequentially.  Calling the "delete asset" endpoint in parallel results in "conflict" error response.  And we may need to deal with edge cases where we delete one but fail to delete the other in this approach.

Therefore, it may seem a cleaner and simpler solution to [create a new endpoint](https://github.com/guardian/media-atom-maker/blob/wsy/fix-delete-self-hosted-assests/app/controllers/Api.scala#L161-L182) for deleting multiple assets at the same time. 

3. The client side maintains the videos in the atom with two set of related data objects, `uploads` and `assets`.  When you uploaded a self-hosted video, the MAM did not update its `assets` objects in the client side.  As a result, when I attempted to delete a self-hosted video immediately after I've uploaded it, the client side failed to look up the video in its `assets` objects.

Solution: The challenge here is that after MAM uploads a self-hosted video, the video is then post-processed in the server side _asynchronously_.  The server side adds the new video to its asset list only after the post-processing finishes.

The client side [polls the `getUploads` endpoint](https://github.com/guardian/media-atom-maker/blob/wsy/fix-delete-self-hosted-assests/public/video-ui/src/components/VideoUpload/VideoTrail.jsx#L22-L29) for showing the current stage of a video that is being post-processed (at server side) on the UI.

The PR adds a new upload status called `post-processing` to the `VideoTrail` component.  Instead of going to `idle` status upon completion of upload, the component is changed to [go to this new status `post-processing`](https://github.com/guardian/media-atom-maker/blob/wsy/fix-delete-self-hosted-assests/public/video-ui/src/components/VideoUpload/VideoTrail.jsx#L34-L37).  And when it [detects that all processing is done](https://github.com/guardian/media-atom-maker/blob/wsy/fix-delete-self-hosted-assests/public/video-ui/src/components/VideoUpload/VideoTrail.jsx#L38-L43) in this status, it calls `getVideo` endpoint to update the client side state and change the component status to `idle`. 

## How to test

I tested the change locally with the browser's inspection tool -

1. Upload a self-hosted video.  We should still be able to see the progress of the video processing at server side.  After all the processing finishes, we should see a call to `api/atoms/:atom_id` to update the client state.

2. Upload a self-hosted video and after it has been processed at server side, hit the "delete" button.  The video should be deleted successfully and it is then removed from the "video trail" UI.  We should see a call to the new endpoint `/api/atoms/:atom-id/asset-list` which returns 200 success response.

I also did the following regression tests -
1. Upload a Youtube video, and hit the delete button.  The video should be deleted successfully and it is then removed from the "video trail" UI.

It allows users to delete self-hosted video assets from an atom.

It aims to fix the bug.  There seems little risk as long as we did the test on CODE before deployment.

### DIFF
--- a/public/video-ui/src/pages/Upload/index.jsx
+++ b/public/video-ui/src/pages/Upload/index.jsx
@@ -107,7 +107,7 @@ import * as getUpload from '../../actions/UploadActions/getUploads';
 import * as s3UploadActions from '../../actions/UploadActions/s3Upload';
 import * as createAsset from '../../actions/VideoActions/createAsset';
 import * as revertAsset from '../../actions/VideoActions/revertAsset';
-import * as deleteAsset from '../../actions/VideoActions/deleteAsset';
+import * as allDeleteAssetActions from '../../actions/VideoActions/deleteAsset';
 import { fetchCategories, fetchChannels } from '../../slices/youtube';
 
 function mapStateToProps(state) {
@@ -128,7 +128,7 @@ function mapDispatchToProps(dispatch) {
         saveVideo,
         createAsset,
         revertAsset,
-        deleteAsset
+        allDeleteAssetActions
       ),
       dispatch
     ),


app/model/commands/DeleteAssetListCommand.scala
```diff
@@ -0,0 +1,56 @@
...
+        mediaAtom.activeVersion match {
+          case Some(activeVersion) if asset.version == activeVersion => {
+            log.warn(markers, s"Cannot delete asset version ${asset.version} on atom $atomId because it is active")
+          }
+          case _ =>
+        }
+      }
+
+      val updatedAtom = mediaAtom.copy(
+        assets = mediaAtom.assets.filterNot(asset => assets.exists(_.id == asset.id))
```
SHession, line 47: For my own knowledge, do the physical assets ever get cleaned up or are they just orphaned?

public/video-ui/src/pages/Upload/index.jsx
```diff
@@ -70,7 +70,7 @@ class VideoUpload extends React.Component {
               activeVersion={activeVersion}
               s3Upload={this.props.s3Upload}
               uploads={this.props.uploads}
-              deleteAsset={this.props.videoActions.deleteAsset}
+              deleteAsset={this.props.videoActions.deleteAssets}
```
SHession, line 73: Just check, is this right? I can't see `deleteAssets` in the `mapDispatchToProps`

SHession: Ah gotcha! For what it's worth that really confused me because effectively we have `deleteAsset.deleteAsset` and `deleteAsset.deleteAssets`. Can we update the imports to make this clearer?

dblatcher (approved):

public/video-ui/src/pages/Upload/index.jsx
waisingyiu: That sounds good to me.  

While we can just import the new `deleteAssets`, I am a bit uncertain about deviating from the pattern of importing all exported functions from those modules.  Worried that future developers would assume to have all exported functions available, especially when JS don't have type checking.

So for now, I just change the import name as `allDeleteAssetActions`, which is an improvement anyway.

PR #1285, opened 2025-09-02
Converts client app root to TypeScript

SHession:
## What does this change?

Small change to increase the coverage of TypeScript around the root of the project.

## How to test

Should be a functional no-op, does the project work as expected?

Areas to check specifically:

- Embedding videos into Composer
- Creating new atoms 

Uncontroversially, more TypeScript!


public/video-ui/src/routes.tsx
```diff
@@ -29,7 +30,7 @@ export const routes = (
       <Redirect from="/videos/create" to="/create" />
       <Route path="/videos/:id" component={Video} onEnter={sendTelemetry} />
       <Route path="/videos/:id/upload" component={Upload} onEnter={sendTelemetry} />
-      <Route path="/create" component={Video} mode="create" onEnter={sendTelemetry} />
+      <Route path="/create" component={Video} props={{mode: "create"}} onEnter={sendTelemetry} />
```
SHession, line 33: This seems to be the way ReactRouter 3 expects to pass props to components, however the previous version was valid, just resulting in type errors.

emdash-ie (approved):
This looks good to me! Have tested locally (creating new atoms but not embedding into composer, because I can’t figure out how to embed an atom from a local instance of media atom maker into a local instance of composer).

SHession:
> This looks good to me! Have tested locally (creating new atoms but not embedding into composer, because I can’t figure out how to embed an atom from a local instance of media atom maker into a local instance of composer).

Given it a quick test on CODE. LGTM!

PR #1292, opened 2025-09-04
Model subtitles directly on atom's sources

Fweddi:
## What does this change?

To allow for better backwards compatibility for downstream clients, model any potential subtitles tracks directly on the atom itself. The subtitles file is already being transcoded, but currently the only reference to it comes from within the m3u8. Now we store the location of the subtitles file as if were another source (of new `assetType`, subtitles).

## To test
With the relevant permissions, upload subtitles to a video and let it transcode. You should be able to see the subtitles correctly on the main atom preview page, and if you find the atom in CAPI, i.e. `https://content.code.dev-guardianapis.com/atom/video/[ATOM_ID]?show-elements=all&show-atoms=all&show-rights=all&show-fields=all&show-tags=all&show-blocks=all&show-references=all&format=json&api-key=[CODE-CAPI-KEY]` you will be able to see the subtitles file in the sources list. This should correspond to the correct file in S3.

### Notes
We added a new asset type, Subtitles in https://github.com/guardian/content-atom/pull/178.

We are handling this new asset type in:
* https://github.com/guardian/frontend/pull/28244
* https://github.com/guardian/dotcom-rendering/pull/14533
* https://github.com/guardian/mobile-apps-api/pull/3917
* https://github.com/guardian/facia-tool/pull/1859

### DIFF
--- a/public/video-ui/src/components/utils/VideoEmbed.jsx
+++ b/public/video-ui/src/components/utils/VideoEmbed.jsx
@@ -36,17 +36,15 @@ export function VideoEmbed({ sources, posterUrl }) {
 }
 
 function prepareSourcesAndSubtitles(sources) {
-  // if m3u8 and mp4 are both present, put m3u8 first and add subtitle track for browsers that don't support m3u8
+  // if m3u8 and mp4 are both present, put m3u8 first
   const m3u8 = sources.find(source => source.mimeType === "application/vnd.apple.mpegurl");
   const mp4 = sources.find(source => source.mimeType === "video/mp4");
   const videoSources =
     (m3u8 && mp4) ? [m3u8].concat(sources.filter(source => source !== m3u8)) : sources;
-  const subtitleTrack =
-    (m3u8 && mp4) ? <track default kind="subtitles" src={getVttName(m3u8.src)}/> : undefined;
-  return {videoSources, subtitleTrack};
-}
 
-function getVttName(m3u8Src) {
-  // captions suffix is an assumption based on MediaConvert output we've seen
-  return m3u8Src.replace(/.m3u8$/, "captions_00001.vtt");
+  // add subtitle track for browsers that don't support m3u8
+  const subtitles = sources.find(source => source.assetType === "subtitles");
+  const subtitleTrack =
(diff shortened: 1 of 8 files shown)


public/video-ui/src/components/utils/VideoEmbed.jsx
```diff
@@ -36,17 +36,15 @@ export function VideoEmbed({ sources, posterUrl }) {
 }
 
 function prepareSourcesAndSubtitles(sources) {
-  // if m3u8 and mp4 are both present, put m3u8 first and add subtitle track for browsers that don't support m3u8
+  // if m3u8 and mp4 are both present, put m3u8 first
   const m3u8 = sources.find(source => source.mimeType === "application/vnd.apple.mpegurl");
   const mp4 = sources.find(source => source.mimeType === "video/mp4");
   const videoSources =
     (m3u8 && mp4) ? [m3u8].concat(sources.filter(source => source !== m3u8)) : sources;
```
gnm-notona, line 43: I think videoSources will need to exclude the vtt source, otherwise we'll get a `<source>` element for the vtt.

public/video-ui/src/components/utils/VideoEmbed.jsx
```diff
@@ -36,17 +36,15 @@ export function VideoEmbed({ sources, posterUrl }) {
...
+
+  // add subtitle track for browsers that don't support m3u8
+  const subtitles = sources.find(source => source.assetType === "subtitles");
```
gnm-notona, line 46: Not sure if source has an `assetType` field.  A alternative would be `source.mimeType === "text/vtt"`

Fweddi: I considered mimeType but the point of the content-atom model change is that we are adding a new `assetType`, subtitles, to the source. So it _should_ be available here (unable to debug at the moment due to those compliation errors).

public/video-ui/src/components/utils/VideoEmbed.jsx
Fweddi: Used the mimeType

gnm-notona (approved):

PR #1301, opened 2025-09-09
Migrate video Reducer to RTK 

SHession:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #1305, opened 2025-09-10
Migrate save state reducer to RTK

SHession:
## What does this change?

This PR continues the work of #1301 converting the saveState reducer to RTK. It and builds on those changes to reduce a significant amount of boilerplate, splitting what was previously a single action with multiple observing reducers, into separate actions. 

For example `X_Request` and `X_Received` actions may have been used to update the save state and then set the video state. These are now handled as separate actions, overall reducing the number of superfluous actions and making thunks more readable. Note: [This may be slightly contentious](https://redux.js.org/style-guide/#model-actions-as-events-not-setters).

## How to test

This change is fairly big, but should be a no-op. Various save actions should correctly update the UI Save State. We should also continue to be updating video state as expected.

### Unit tests(?)

I experimented with adding some tests to the Thunks to confirm actions correctly updated the state. We could adopt this approach if we thought it was worthwhile: https://github.com/guardian/media-atom-maker/commit/252b0180e7f9ace7a8c04c85a14bce7d5475fece#diff-005cff2ee4b4f4bcc83d433b488c5beeb9c71672320daf918968f1393e63eba4. I would also be open to adding more integration tests to confirm this change works as intended. Open to discussion on this topic.  

Another significant step towards RTK adoption with improvements to the readability of the code.


waisingyiu (approved):
Thank you Sam for the great work!  I tested it locally e.g. creating a new atom, opening an existing one, uploading a self-hosted video, uploading a Youtube video, activating an asset and deleting one.  It all works great.

I don't think the discussion on the Redux reducers should block this PR.  It is an improvement, and we could adjust the reducers if needed.

PR #1312, opened 2025-09-15
Add filter for Youtube videos

Fweddi:
## What does this change?

image

Follow up to #1297. Editorial would like to be able to filter for Youtube videos as well as Loops. This PR changes the self-hosted filter (boolean) to a media platform filter (string), which is extensible and can support any media platform type.

## How to test
Test the filter dropdown works on both DEV (uses dynamoDB) and CODE (uses CAPI).

Tested on:
- [x] DEV
- [x] CODE


public/video-ui/src/components/Header.jsx
```diff
@@ -85,8 +85,9 @@ export default class Header extends React.Component {
             );
           }}
         >
-          <option value="false">All videos</option>
-          <option value="true">Loops</option>
+          <option value="">All videos</option>
+          <option value="url">Loops</option>
```
Fweddi, line 89: Any self-hosted video (platform is `url`) is currently a loop.

Fweddi: Note this will lead to url query params which are slightly more opaque to users. I think this is okay; the dropdown will indicate they are searching for loops.

public/video-ui/src/components/Header.jsx
```diff
@@ -85,8 +85,9 @@ export default class Header extends React.Component {
...
+          <option value="youtube">YouTube</option>
```
Fweddi, line 90: We could extend to other media platform types (https://github.com/guardian/content-atom/blob/main/thrift/src/main/thrift/atoms/media.thrift#L10-L16) but I'm not sure there's much appetite for this at present. Would be trivial to add if we wanted to, now this code is extensible.

app/data/AtomListStore.scala
```diff
@@ -1,30 +1,38 @@
...
 class CapiBackedAtomListStore(capi: CapiAccess) extends AtomListStore {
-  override def getAtoms(search: Option[String], limit: Option[Int], shouldUseCreatedDateForSort: Boolean, shouldFilterForSelfHosted: Boolean): MediaAtomList = {
+  override def getAtoms(search: Option[String], limit: Option[Int], shouldUseCreatedDateForSort: Boolean, mediaPlatform: Option[String]): MediaAtomList = {
     // CAPI max page size is 200
     val cappedLimit: Option[Int] = limit.map(Math.min(200, _))
 
+    val dateSorter = shouldUseCreatedDateForSort match {
+      case true => Map("order-date" -> "first-publication")
+      case false => Map.empty
+    }
```
Fweddi, line 23: Could be an if statement. I'm using pattern matching to achieve consistency with the filter below - makes it a little easier to read.

Issue #1309, opened 2025-09-11
Migrate checkedFormFields reducer to RTK


LAKSHMIRPILLAI:
Migrate checkedFormFields reducer to RTK

PR #1314, opened 2025-09-16
Migrate checkedFormFields reducer to Redux Toolkit

LAKSHMIRPILLAI:
## What does this change?

Migrate the checkedFormFields reducer to Redux Toolkit as part of an effort to modernise the codebase

[Issue](https://github.com/guardian/media-atom-maker/issues/1309)

## How to test

Test if the form data fields are getting populated foe scenarios.
Ex: It should throw an error message "This field is required" in case title is absent. Also once the title field is updated, it should change back to null.

image

image

Relatively safe


public/video-ui/src/actions/FormErrorActions/updateFormErrors.js
SHession, line 1: This file is small and not especially useful anymore, it would be nice to remove it and replace callers of `updateFormErrors` with `updateCheckedFormFieldsErrors` directly

SHession (commented):
Looking good, I've left one comment. There are also some reasonable suggestions from GitHub CoPilot worth actioning

public/video-ui/src/slices/checkedFormFields.ts
```diff
@@ -0,0 +1,25 @@
+import { createSlice, Action, PayloadAction } from '@reduxjs/toolkit';
+
+export type CheckedFormFieldsState = {
+  [key: string]: Record<string, string | null | undefined>;
```
SHession, line 4: it would be nice to define this once rather than here as `[key: string]: Record<string, string | null | undefined>` and then on line 7 as `Record<string, Record<string, string | null | undefined>>`. 

Furthermore, I think we can be more specific about the type of the state. It appears to be:
`Record<string, Record<string, FieldNotification>>`

Taken from https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/constants/FieldNotification.ts

SHession (commented):
Changes looking good! One comment about the slice state typing

SHession (approved):
LGTM!

PR #1316, opened 2025-09-16
migrate usage reducer to rtk

dblatcher:
## What does this change?

Replace the usage reducer with an RTK slice.

Removes the `fetchingUsages` state from the `saveState` slice, instead representing it as `isFetching` on the usage slice (no components were using that state).

To address the other reducers responding to or using  actions "owned" by the usage reducer:
 - the new slice exports reducer actions to replace the "VIDEO_PAGE_CREATE_POST_RECEIVE" and "VIDEO_PAGE_UPDATE_POST_RECEIVE" actions. 

**update** - also removes the "composerPageReducer" - its state was not being used.

## How to test

Usages should be loaded an reflected in the UI as before - this is mainly seen in the "usages" tab on the video pages.

The UI should be correctly updated after creating a composer page from MAM (the usage tab should include this a preview video usage).

After changing the headline in the "furniture" tab, and saving the change, the headline change should be updated if you switch to the "usages" tab.

Screenshot 2025-09-16 at 10 36 56

Note that running locally, (on this branch and main) the usages cannot be loaded as requests for live capi usages fail (the MAM backend returns a 500 error and the UI will show the "Could not get video usages" error). Haven't investigated the cause for this, but it was possible to silence this issue locally by adjusting `VideoApi.getVideoUsages` to not request the published usages:
```
  getVideoUsages: (videoId: string):Promise<UsageData> => {
    return Promise.all([
      getUsages({ id: videoId, stage: ContentApi.preview }),
      // getUsages({ id: videoId, stage: ContentApi.published })
    ]).then(data => {
      const [previewUsages] = data;
      const publishedUsages= [] as typeof previewUsages;

```

Modernises and simplifies the code.

## Accessibility

### DIFF
--- a/public/video-ui/src/actions/VideoActions/saveVideo.ts
+++ b/public/video-ui/src/actions/VideoActions/saveVideo.ts
@@ -14,6 +14,8 @@ const errorMessages = {
 
 export const saveVideo = (video: Video) => async (dispatch: AppDispatch) => {
   dispatch(setSaving(true));
+  dispatch(setVideo(video));
+
   const savedVideo: Video | undefined = await VideosApi.saveVideo(video.id, video)
     .catch(error => {
       const message = error.status === 409


public/video-ui/src/slices/usage.ts
```diff
@@ -0,0 +1,107 @@
...
+>(
+    'usage/fetchUsages',
+    (id, { dispatch }) => {
+        return VideosApi.getVideoUsages(id).then(
+            data => {
+                const { preview, published } = data;
+                // a lot of components conditionally render based on the number of usages,
+                // rather than constantly call a utility function, they use the stateful value.
+                // It should be possible to refactor those components to derive the number
+                // from the data so the value does not need to be in the UsageState.
```
SHession, line 49: This sounds like a nice improvement. In the short term can we change the payload of the action to only include the data and have the logic to determine the totals in the reducer? 

https://redux.js.org/style-guide/#put-as-much-logic-as-possible-in-reducers

Ultimately it would be nice to have the total provided as a selector.

public/video-ui/src/reducers/composerPageReducer.js
```diff
@@ -1,10 +1,10 @@
 export default function usage(state = {}, action) {
   switch (action.type) {
-    case 'VIDEO_PAGE_CREATE_POST_RECEIVE': {
-      const videoId = action.newPage.videoId;
+    case 'usage/addNewlyCreatedVideoUsage': {
```
SHession, line 3: Small improvement, you can reference the action type with `addNewlyCreatedVideoUsage.type`

public/video-ui/src/slices/usage.ts
```diff
@@ -0,0 +1,107 @@
...
+            state.data = initialState.data;
+            state.totalUsages = initialState.totalUsages;
+            state.totalVideoPages = initialState.totalVideoPages;
+        },
+        updateVideoUsageWebTitle(state, action: Action & { payload: string }) {
+            const updateWebtitle = (usage: CapiContent): CapiContent => ({ ...usage, webTitle: action.payload });
+            state.data.preview.video = state.data.preview.video.map(updateWebtitle);
+            state.data.published.video = state.data.published.video.map(updateWebtitle);
+        },
+        addNewlyCreatedVideoUsage(state, action: Action & { payload: CapiContent }) {
```
SHession, line 84: There is a helper Type for `Action & { payload: CapiContent }` as `PayloadAction<T>`

public/video-ui/src/reducers/composerPageReducer.js
```diff
@@ -1,10 +1,10 @@
...
+      const videoId = action.payload.videoId;
       state[videoId] = {
-        composerId: action.newPage.composerId,
-        usage: action.newPage.usage
+        composerId: action.payload.composerId,
+        usage: action.payload.usage
```
SHession, line 7: The expected payloads between the nice slice action and this JS file don't seem to match: 

Image

public/video-ui/src/actions/VideoActions/videoPageCreate.js
```diff
@@ -8,14 +10,6 @@ function requestVideoPageCreate() {
...
 
-function receiveVideoPageCreate(capiPage) {
-  return {
-    type: 'VIDEO_PAGE_CREATE_POST_RECEIVE',
-    newPage: capiPage,
-    receivedAt: Date.now()
-  };
-}
-
 function errorReceivingVideoPageCreate(error) {
```
SHession, line 13: Can we use the showError action to replace this and slim the code down a little?

public/video-ui/src/actions/VideoActions/saveVideo.js
SHession, line 1: Could we potentially convert this file to TypeScript?

public/video-ui/src/actions/VideoActions/videoPageCreate.js
```diff
@@ -48,7 +42,7 @@ export function createVideoPage(id, video, isTrainingMode) {
           // so keep trying until success or timeout
           return ContentApi.getByPath(pagePath, true).then(capiResponse => {
             return dispatch(
-              receiveVideoPageCreate(capiResponse.response.content)
+              addNewlyCreatedVideoUsage(capiResponse.response.content)
```
SHession, line 45: I'm seeing some errors when I use the create video page button: "Could not create a video page"

Image

public/video-ui/src/actions/VideoActions/videoPageUpdate.jsx
```diff
@@ -8,14 +9,6 @@ function requestVideoPageUpdate() {
...
 
-function receiveVideoPageUpdate(newTitle) {
-  return {
-    type: 'VIDEO_PAGE_UPDATE_POST_RECEIVE',
-    newTitle: newTitle,
-    receivedAt: Date.now()
-  };
-}
-
 function errorReceivingVideoPageUpdate({error, message}) {
```
SHession, line 12: Another case for showError()?

SHession (commented):
This looks like a nice improvement!

> Removes the fetchingUsages state from the saveState slice, instead representing it as isFetching on the usage slice (no components were using that state).

If nothing is using `fetchingUsages` can we simply remove it?

public/video-ui/src/actions/VideoActions/videoPageCreate.js
dblatcher: On some investigation, I found we got similar errors messages on main locally, (but not on CODE). As noted, the page was actually created, despite the message. On main, the error thrown was because of non serialisable `Error`s being dispatched in the "SHOW_ERROR" actions. 

Switching to the new `showError` function cleared that up and revealed the underlying exception (IE the exception triggering the "SHOW_ERROR" that raised another error) was to do with state modification detected within an action - happening within the composerPageReducer.

I haven't verified this, but I assume the error message doesn't show on CODE (or PROD?) because the "non serialisable state" and "state modification in action" exceptions are ignored in Production mode.

In short, I think removing the composerPageReducer will address this issue.

public/video-ui/src/slices/usage.ts
dblatcher: 👍  I had shied away from doing that as it would involve changing the implementation of saveVideo.js (since it uses the action returned by `fetchUsages`), but after converting it to typescript, that seemed less risky

dblatcher:
> > Removes the fetchingUsages state from the saveState slice, instead representing it as isFetching on the usage slice (no components were using that state).
> 
> If nothing is using `fetchingUsages` can we simply remove it?

We certainly could - I'd tend to keep it as there are probably bits of the UI that should know if the usages are being fetched, but happy to remov it it you think that's best.

public/video-ui/src/actions/VideoActions/saveVideo.js
```diff
@@ -1,49 +0,0 @@
...
-  };
-}
-
-export function saveVideo(video) {
-  return dispatch => {
-    dispatch(setSaving(true));
-    dispatch(setVideo(video));
-    return VideosApi.saveVideo(video.id, video)
-      .then(res => {
-        dispatch(setSaving(false));
```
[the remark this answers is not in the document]
dblatcher: Thanks - I have included this in new   `saveVideo` function.

I've included it even after dispatches of `showError` - its not strictly necessary as "SHOW_ERROR"  also sets `state.saving` to `false` because of the extraReducer in the `saveState` slice, but it seemed best not to rely on that - we might want to consider phasing out that side effect.

public/video-ui/src/actions/VideoActions/saveVideo.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const savedVideo: Video | undefined = await VideosApi.saveVideo(video.id, video)
+    .catch(error => {
+      const message = error.status === 409
+        ? errorMessages.saveVideoConflict
+        : errorMessages.saveVideoDefault;
+      dispatch(showError(message, error));
+      return undefined;
+    });
+  if (!savedVideo) {
```
SHession, line 28: Prior to this PR we would call We previously would call `setVideo` with the returned Video object from the savedVideo request.

public/video-ui/src/actions/VideoActions/saveVideo.ts
```diff
@@ -0,0 +1,56 @@
...
+    dispatch(setSaving(false));
+    return;
+  }
+
+  const fetchUsagesAction = await dispatch(fetchUsages(video.id));
+  if (fetchUsagesAction.meta.requestStatus !== 'fulfilled') {
```
SHession, line 34: Looks like the idiomatic RTK way to handle this is using the unwrap function: 
https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions. It would make it a little more consistent with how we are handling promises in the rest of this function.

public/video-ui/src/actions/VideoActions/saveVideo.ts
```diff
@@ -0,0 +1,40 @@
...
+  return (dispatch: ThunkDispatch<RootState, void, Action>) => {
+    dispatch(setSaving(true));
+    dispatch(setVideo(video));
+    return VideosApi.saveVideo(video.id, video)
+      .then(res => {
+        return dispatch(fetchUsages(video.id))
+          .then(fetchUsagesAction => {
+            if (fetchUsagesAction.meta.requestStatus === 'rejected') {
+              throw new Error('fetchUsagesFulfilledAction rejected');
+            }
```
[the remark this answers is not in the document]
SHession: I think refactoring was a good move, it was getting pretty hard to follow. I have added some more comments. LMK what you think and if you want to pair at all.

PR #1317, opened 2025-09-16
Show which videos are loops and videos on the main search page

Fweddi:
## What does this change?

image

Show an atom's media platforms through an icon (Youtube icon for Youtube, Loop icon for Url / self-hosted). 

If an atom has multiple assets with different media platforms, show both icons, with the active asset highlighted and the inactive asset faded.

NB. The 'Loop' filter and icon is not strictly accurate, as there are rare cases where self-hosted videos don't loop (pre-Youtube content, sensitive content which we can't upload to Youtube). We are carrying out some work to model whether an atom loops by default on the atom itself, which should clean this up as we can identify a loop based on that field.

### DIFF
--- a/app/data/AtomListStore.scala
+++ b/app/data/AtomListStore.scala
@@ -95,7 +95,7 @@ class CapiBackedAtomListStore(capi: CapiAccess) extends AtomListStore {
 
       // sort media platforms so the current one is first
       val sortedMediaPlatforms = currentMediaPlatform match {
-        case Some(current) => (current :: mediaPlatforms.filter(_ != current)).distinct
+        case Some(current) => current :: mediaPlatforms.filter(_ != current)
         case None => mediaPlatforms
       }
 
@@ -156,7 +156,7 @@ class DynamoBackedAtomListStore(store: PreviewDynamoDataStore) extends AtomListS
 
     // sort media platforms so the current one is first
     val sortedMediaPlatforms = currentMediaPlatform match {
-      case Some(current) => (current :: mediaPlatforms.filter(_ != current)).distinct
+      case Some(current) => current :: mediaPlatforms.filter(_ != current)
       case None => mediaPlatforms
     }
 


public/video-ui/src/components/VideoItem/index.jsx
```diff
@@ -46,6 +45,7 @@ export default class VideoItem extends React.Component {
 
   render() {
     const video = this.props.video;
+    const mediaPlatform = VideoUtils.getMediaPlatform(video);
```
Fweddi, line 48: note the 'video' here doesn't have all the video fields, just the fields returned from the mediaAtomSummary endpoint

public/video-ui/src/components/Icon.jsx
```diff
@@ -94,6 +94,36 @@ export class SubtitlesIcon extends React.Component {
...
+  render() {
+    return (
+      <svg fill="currentColor" width="32px" height="32px" viewBox="0 0 32 32" version="1.1" xmlns="http://www.w3.org/2000/svg">
+        <path d="M12.932 20.459v-8.917l7.839 4.459zM30.368 8.735c-0.354-1.301-1.354-2.307-2.625-2.663l-0.027-0.006c-3.193-0.406-6.886-0.638-10.634-0.638-0.381 0-0.761 0.002-1.14 0.007l0.058-0.001c-0.322-0.004-0.701-0.007-1.082-0.007-3.748 0-7.443 0.232-11.070 0.681l0.434-0.044c-1.297 0.363-2.297 1.368-2.644 2.643l-0.006 0.026c-0.4 2.109-0.628 4.536-0.628 7.016 0 0.088 0 0.176 0.001 0.263l-0-0.014c-0 0.074-0.001 0.162-0.001 0.25 0 2.48 0.229 4.906 0.666 7.259l-0.038-0.244c0.354 1.301 1.354 2.307 2.625 2.663l0.027 0.006c3.193 0.406 6.886 0.638 10.634 0.638 0.38 0 0.76-0.002 1.14-0.007l-0.058 0.001c0.322 0.004 0.702 0.007 1.082 0.007 3.749 0 7.443-0.232 11.070-0.681l-0.434 0.044c1.298-0.362 2.298-1.368 2.646-2.643l0.006-0.026c0.399-2.109 0.627-4.536 0.627-7.015 0-0.088-0-0.176-0.001-0.263l0 0.013c0-0.074 0.001-0.162 0.001-0.25 0-2.48-0.229-4.906-0.666-7.259l0.038 0.244z"></path>
+      </svg>
+    );
+  }
+}
+
+export class LoopingIcon extends React.Component {
```
Fweddi, line 107: this is the same iconography as https://github.com/guardian/facia-tool/blob/main/fronts-client/src/components/icons/Icons.tsx#L475-L489

public/video-ui/src/util/video.js
```diff
@@ -144,4 +144,8 @@ export default class VideoUtils {
   static hasExpired({ contentChangeDetails }) {
     return !!contentChangeDetails.expiry && contentChangeDetails.expiry.date <= Date.now();
   }
+
+  static getMediaPlatform(atomSummary) {
+    return atomSummary?.mediaPlatform?.toLowerCase() || null;
```
Fweddi, line 149: `mediaPlatform` comes back from the server with the first letter capitalised

app/data/AtomListStore.scala
```diff
@@ -80,7 +80,16 @@ class CapiBackedAtomListStore(capi: CapiAccess) extends AtomListStore {
         (asset \ "version").as[Long]
       }
 
-      Some(MediaAtomSummary(id, title, posterImage, contentChangeDetails))
+      val currentAsset = (atom \ "assets").as[JsArray].value.find { asset =>
+        val assetVersion = (asset \ "version").as[Long]
+        activeVersion.contains(assetVersion)
```
Fweddi, line 85: effectively we want to find the asset where activeVersion == assetVersion, but since activeVersion is an `Option[Long]` and assetVersion is just a `Long`, we use `.contains`

app/data/AtomListStore.scala
```diff
@@ -80,7 +80,16 @@ class CapiBackedAtomListStore(capi: CapiAccess) extends AtomListStore {
...
+      }
+
+      val mediaPlatform = currentAsset.flatMap { asset =>
+        (asset \ "platform").asOpt[String].map(_.toLowerCase)
+      }
```
Fweddi, line 90: When we search for atoms by platform on CAPI, we check against all the assets. So if one asset in an atom has a platform of `url`, it will get returned by a query looking for `media-platform=url`.

Here instead we are checking for the latest asset. We could check all assets, in which case we could potentially show _both_ Youtube _and_ Loop if one asset is self-hosted, and the other is Youtube. This might be more confusing - or less?

Very few production atoms will contain both a Youtube and Loop asset.

domlander:
I wonder if the loop icon may become confusing when self-hosted videos are used for more than just loops. I don't have a better suggestion for a self-hosted video icon though.

Since it looks like we're already using the same icon in facia-tool, it probably makes sense to leave it as it is here in this PR, then update the icon across both places once we start using self-hosted video for non-looping videos.

app/data/AtomListStore.scala
domlander: I think it makes more sense to show both Youtube and Loop if one asset is self-hosted and another is Youtube, but I can see how this can potentially be confusing either way.

paperboyo:
> when self-hosted videos are used for more than just loops

They already are/were. Last time in February to work around some censorship from Youtube IIRC…
Currently, Loops are indeed all the rage and majority of uses, but when suddenly a high profile case (like the one in Feb) comes, it may indeed be confusing.

Beyond UI, we should absolutely make sure that self-hosted videos will continue to allow for non-loopy. non-cinemagraphy, just video-y content.

app/data/AtomListStore.scala
Fweddi: Okay, I've changed the logic so that it shows all the platforms - if the asset is inactive it shows with lower opacity, and the active asset is always the one on the right hand side.

app/data/AtomListStore.scala
```diff
@@ -131,8 +150,17 @@ class DynamoBackedAtomListStore(store: PreviewDynamoDataStore) extends AtomListS
 
   private def fromAtom(atom: MediaAtom): MediaAtomSummary = {
     val versions = atom.assets.map(_.version).toSet
+    val currentAsset = atom.assets.find(asset => asset.version == atom.activeVersion.getOrElse(versions.max))
+    val mediaPlatforms = atom.assets.map(_.platform.name.toLowerCase).distinct
+    val currentMediaPlatform = currentAsset.map(_.platform.name).map(_.toLowerCase)
+
+    // sort media platforms so the current one is first
+    val sortedMediaPlatforms = currentMediaPlatform match {
+      case Some(current) => (current :: mediaPlatforms.filter(_ != current)).distinct
```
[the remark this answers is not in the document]
Fweddi: good catch

Fweddi:
> > when self-hosted videos are used for more than just loops
> 
> They already are/were. Last time in February to work around some censorship from Youtube IIRC… Currently, Loops are indeed all the rage and majority of uses, but when suddenly a high profile case (like the one in Feb) comes, it may indeed be confusing.
> 
> Beyond UI, we should absolutely make sure that self-hosted videos will continue to allow for non-loopy. non-cinemagraphy, just video-y content.

I've just caught up with @jonathonherbert, we're thinking of adding a new field to the Atom to connote 'looping by default'. When that field is released, we can make the Loops filter search on atoms which have that field enabled. This will make the Loops filter more accurate and exclude the just video-y / sensitive video use cases you've just described @paperboyo.
 
Until then, since the vast majority of self-hosted videos are intended to be 'Loops', I think we can live with this slightly imperfect world where sensitive videos are incorrectly marked as Loops.

PR #1320, opened 2025-09-17
Create search RTK slice

SHession:
## What does this change?

This PR introduces a new `search` slice, which combines several small existing reducers: 

- searchTerm Reducer
- mediaPlatform Reducer
- shouldUseCreatedDateForSortReducer

This should move us closer to RTK adoption and make the code a little more condensed. 

## How to test

The changed reducer relate the /videos view and specifically the search, filter and sort options. Testing these continue to work as expected should be sufficient.


public/video-ui/src/actions/actions.ts
emdash-ie, line 1: What’s the reason for deleting this file? Was it just unused?

emdash-ie (approved):
Looks good to me and works as expected locally

PR #1321, opened 2025-09-17
Moves Error Key state into Error slice

SHession:
## What does this change?

This PR removes the errorKey reducer and moves the state into the error slice. This was originally intended but we didn't have enough confidence this wouldn't break the error handling. 

## How to test

Emulate errors and confirm they show in the view's top banner. When errors occur the banner should animate and be scrolled into view. 

One fewer reducer to worry about!

### DIFF
--- a/public/video-ui/src/components/ReactApp.jsx
+++ b/public/video-ui/src/components/ReactApp.jsx
@@ -85,10 +85,11 @@ class ReactApp extends React.Component {
         />
         {this.props.error.message
           ? <div
-              key={this.props.error.key}
-              className={`error-bar error-bar--animate`}
-              dangerouslySetInnerHTML={{ __html: this.props.error.message }}
-            />
+            key={this.props.error.key}
+            className={`error-bar error-bar--animate`}
+          >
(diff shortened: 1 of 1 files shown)


public/video-ui/src/components/ReactApp.jsx
```diff
@@ -83,11 +83,11 @@ class ReactApp extends React.Component {
...
           error={this.props.error}
         />
-        {this.props.error
+        {this.props.error.message
           ? <div
-              key={this.props.errorKey}
+              key={this.props.error.key}
               className={`error-bar error-bar--animate`}
-              dangerouslySetInnerHTML={{ __html: this.props.error }}
+              dangerouslySetInnerHTML={{ __html: this.props.error.message }}
```
dblatcher, line 90: Looking at the usages of `showError`, I don't see any instances of HTML being used for the error message (plain strings only) so we could render it directly - although might limit future use.

SHession: I had the same thought, I can change it

PR #1323, opened 2025-09-17
 Migrate formFieldsWarning Reducer to RTK

LAKSHMIRPILLAI:
## What does this change?

Migrate the checkedFormFields reducer to Redux Toolkit as part of an effort to modernise the codebase

[Issue](https://github.com/guardian/media-atom-maker/issues/1309)

## How to test

Test if the form data fields are getting populated foe scenarios.
Ex: It should throw an error message "This field is required" in case title is absent. Also once the title field is updated, it should change back to null.

image

image

Relatively safe


SHession (commented):
This is a nice improvement! A few comments about improving the typing. 

In general, it might be nice to combine this slice with the checkedFormFieldsReducer which seems very similar. 

As far as I can tell, it appears that the state is only being used by the Header for displaying the required Composer fields message, which assumed any attribute found is required for publishing to Composer. It is setup in conjunction with the ManagedField `isDesired` prop.

SHession (approved):

PR #1332, opened 2025-09-18
Migrate s3Upload Reducer to RTK

LAKSHMIRPILLAI:
## What does this change?

Migrate the s3Upload Reducer to Redux Toolkit as part of an effort to modernise the codebase

[Issue](https://github.com/guardian/media-atom-maker/issues/1329)

## How to test

- Create a new atom and Click on the Video Preview 
- Upload a new video to YouTube /Self-Hosted Video
image

- Check the different Redux states from 

image

- Check the state values

image

- Check if the video was uploaded successfully

Relatively safe

### DIFF
--- a/public/video-ui/src/actions/UploadActions/s3Upload.js
+++ b/public/video-ui/src/actions/UploadActions/s3Upload.js
@@ -27,7 +27,8 @@ export function startPacFileUpload({id, file}) {
     return uploadPacFile({id, file}).then(() => {
       dispatch(setS3UploadStatusToComplete());
     }).catch(err => {
-      dispatch(setS3UploadStatusToError(errorDetails(err)));
+      dispatch(showError(errorDetails(err), err));
+      dispatch(setS3UploadStatusToError());
     });
   };
 }
@@ -37,7 +38,8 @@ export function startSubtitleFileUpload({id, version, file}) {
     return uploadSubtitleFile({id, version, file}).then(() => {
       dispatch(setS3UploadStatusToComplete());
     }).catch(err => {
-      dispatch(setS3UploadStatusToError(errorDetails(err)));
+      dispatch(showError(errorDetails(err), err));
+      dispatch(setS3UploadStatusToError());
     });
   };
 }
@@ -47,7 +49,8 @@ export function deleteSubtitle({id, version}) {
     return deleteSubtitleFile({id, version}).then(() => {
       dispatch(setS3UploadStatusToComplete());
     }).catch(err => {
-      dispatch(setS3UploadStatusToError(errorDetails(err)));
+      dispatch(showError(errorDetails(err), err));
+      dispatch(setS3UploadStatusToError());
     });
   };
 }


public/video-ui/src/slices/s3Upload.ts
```diff
@@ -0,0 +1,50 @@
...
+        status: 'uploading'
+      });
+    },
+    s3UploadProgress: (state : S3UploadState, action: S3UploadAction)=>{
+      Object.assign(state, {
+        progress: action.payload.progress,
+        status: 'uploading'
+      });
+      },
+      s3UploadPostProcessing: _=> ({
```
dblatcher, line 32: Instead of having the reducer methods return a whole `S3UploadState`, they can also modify the property of the current value, which would probably be a bit clearer in these reducers that just need to change the status eg:
```
    s3UploadPostProcessing: (state) => {
      state.status = 'post-processing';
    },
```

This is just another stylistic point, but if the methods had "verb" names like "setS3UploadStatusToPostProcessing", it would make it a bit clearer what dispatching them would do

public/video-ui/src/slices/s3Upload.ts
```diff
@@ -0,0 +1,50 @@
...
+
+type S3UploadAction = PayloadAction<S3UploadState>
+
+const initialState : S3UploadState =  { id: null, progress: 0, total: 0, status: 'idle' };
+
+const s3Upload = createSlice({
+  name: 's3Upload',
+  initialState,
+  reducers: {
+    s3UploadStarted:(state :S3UploadState, action)=> {
```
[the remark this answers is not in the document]
dblatcher: I just saw we already have a definition of the `Upload` type at public/video-ui/src/components/VideoUpload/VideoAsset.tsx - I'd suggest moving that to this file for now.

public/video-ui/src/slices/s3Upload.ts
```diff
@@ -0,0 +1,50 @@
+import { createSlice,PayloadAction } from '@reduxjs/toolkit';
+
+export interface S3UploadState {
+  id: string | null;
+  progress: number;
+  total: number;
+  status: 'idle' | 'uploading' | 'post-processing' | 'complete' | 'error';
+}
+
+type S3UploadAction = PayloadAction<S3UploadState>
```
dblatcher, line 10: If you follow the path of how the reducer method that uses this action (`s3UploadProgress`) is used or log out the action in the method, the `payload` is actually a number, not an `S3UploadState` object.

It's a bit hard to to follow as the  `s3UploadProgress` gets wrapped in the `progress` function in `startVideoUpload`, then passed round as a callback.

public/video-ui/src/slices/s3Upload.ts
```diff
@@ -0,0 +1,50 @@
...
+      const total = action.payload.parts[action.payload.parts.length - 1].end;
+      Object.assign(state, {
+        id: action.payload.id,
+        total: total,
+        status: 'uploading'
+      });
+    },
+    s3UploadProgress: (state : S3UploadState, action: S3UploadAction)=>{
+      Object.assign(state, {
+        progress: action.payload.progress,
```
dblatcher, line 28: As noted above - the payload will be a number, so setting `progress` to `action.payload.progress` will be setting it to `undefined`.
```suggestion
        progress: action.payload,
```

dblatcher: it would be best to not use `Object.assign` as we loose type safety that way - I think it was the norm in the js reducer files as it meant they didn't need to track which fields where being changed that way.

dblatcher (commented):
Nice work - I can see that the uploads are still working, but I thinks there's some improvements to the typing and implementation that are worht going over.

public/video-ui/src/slices/s3Upload.ts
LAKSHMIRPILLAI: This is the point  I mentioned in the huddle last week, that I wanted to check with you, Thanks for clarifying

PR #1336, opened 2025-09-22
Migrate targeting Reducer to RTK

LAKSHMIRPILLAI:
## What does this change?

Migrate the targeting Reducer to Redux Toolkit as part of an effort to modernise the codebase

[Issue](https://github.com/guardian/media-atom-maker/issues/1334)

## How to test

- Create a new atom and Click on the Targeting Tab
- Check the different Redux states
- On moving to Targeting Tab, I must call the below actions
image

- Click on 'Add targeting rule' button 
image

image

It will call below actions 

image

- Check the state values
image

- Now delete existing targeting rule

It will call below actions 

image

- Check the state values

image

- Check if the targeting has been removed successfully

Relatively safe


public/video-ui/src/slices/targeting.ts
```diff
@@ -0,0 +1,68 @@
...
+  initialState,
+  reducers: {
+    requestUpdateTarget:(state:TargetingState, action :PayloadAction<TargetType>)=> ({
+        ...state,
+        targets: [
+          ...(state.targets || []).filter(({ id }) => id !== action.payload.id),
+          action.payload
+        ]
+    }),
+    receiveUpdateTarget:(state:TargetingState)=> ({
```
waisingyiu, line 31: This action and the one below `requestCreateTarget` do not change the state at all.  Do you think if it's worth removing them to simplify the reducer?

PR #1341, opened 2025-09-23
Checks YouTube API for live duration

SHession:
## What does this change?

This PR is a proposed fix for issues with atoms having a duration of 0 and therefore displaying a "Live" badge on our platforms. Based on conversations with users, the most common case is non-live assets having the badge, there were no reported issues with genuinely live videos missing the badge. 

### Issue hypothesis

When determining the duration for a YouTube asset, if the Video is not returned for any reason we set the Duration to `None`. 

https://github.com/guardian/media-atom-maker/blob/dc196f62a5d4da65224f6fc8139466173f0e2dd2/common/src/main/scala/com/gu/media/youtube/YouTubeVideos.scala#L39-L48

https://github.com/guardian/media-atom-maker/blob/dc196f62a5d4da65224f6fc8139466173f0e2dd2/app/model/commands/ActiveAssetCommand.scala#L36-L39

 When combined with the TypeScript atom default of 0, this would set the Atom's duration to 0 rather than None. Therefore inadvertently indicating that the video is live. 

https://github.com/guardian/media-atom-maker/blob/4838d3eea52d40dc9a5d2e6ace717f2ace2f9389/public/video-ui/src/constants/blankVideoData.ts#L8

https://github.com/guardian/media-atom-maker/blob/4838d3eea52d40dc9a5d2e6ace717f2ace2f9389/public/video-ui/src/slices/video.ts#L32-L37

### Extra context

There is existing API function to attempt to set the assets duration which behaves slightly differently, only updating the duration if the asset is found:

https://github.com/guardian/media-atom-maker/blob/dc196f62a5d4da65224f6fc8139466173f0e2dd2/app/controllers/Api.scala#L79-L87

### Fix

This PR intends to fix this issue by not setting the duration to None if the asset can't be found, instead using the existing duration value (similar to the `resetDurationFromActive` function). It also changes the default from 0 to 1, which is a more preferable non-live default. AFAIK live videos should correctly set this back to 0.

I am about 70% confident in the fix, although I have not tested it in reality and am not sure of a reliable way to do so. It seems like a reasonable change, regardless of whether it solves the bug. However, reasonable changes can have unintended consequences. 

**We could implement just the logging and see if it corresponds to production issues.** 

## How to test

Open to suggestions on how to test this! The backend has no existing tests and this is an issue which exists between the frontend and backend. 

Hopefully solving the intermittent issue with live badges on non-live videos.  

Could setting the video duration to a plausible looking but wrong duration (i.e. from one atom asset to another) cause problems of its own?

### DIFF
--- a/public/video-ui/src/constants/blankVideoData.ts
+++ b/public/video-ui/src/constants/blankVideoData.ts
@@ -5,7 +5,7 @@ export const blankVideoData: Video = {
   title: '',
   description: '',
   category: 'News',
-  duration: 1,
+  duration: 0,
   channelId: '',
   youtubeTitle: '',
   youtubeCategoryId: '',


dblatcher (commented):
This feels like a sensible approach.

My only concern is whether we need to add something to the UI to make it more obvious that new videos will default to 1s duration. I don't now the workflow, but if user posting videos which actually are live are used to the default being zero, they might not notice.

PR #1343, opened 2025-09-24
Prevents setPublishedVideo from overwriting video state acidentally

SHession:
## What does this change?

This PR fixes a pernicious bug which could cause the published video state to overwrite the draft state. This could leave to some confusing UI bugs where draft entries would sometimes appear but sometimes be overwritten by preview. 

Introduced by https://github.com/guardian/media-atom-maker/pull/1331/files#diff-ae6115d5403f4a600bbd09142ee2a78b1acf83a0e3cd13468e74149438d84c88.

## How to test

Testable locally on a published atom. Makes changes to an atom after publish, refresh the page and notice sometimes the preview state is displayed and sometimes the draft state is displayed.

Solves a nasty UI bug.

### DIFF
--- a/public/video-ui/src/slices/video.ts
+++ b/public/video-ui/src/slices/video.ts
@@ -20,6 +20,9 @@ const video = createSlice({
       state.video = { ...blankVideoData, ...payload };
     },
     setPublishedVideo: (state, { payload }: PayloadAction<Video>) => {
+      state.publishedVideo = { ...blankVideoData, ...payload };
+    },
+    setVideoAndPublishedVideo: (state, { payload }: PayloadAction<Video>) => {
       state.video = { ...blankVideoData, ...payload };
       state.publishedVideo = { ...state.video };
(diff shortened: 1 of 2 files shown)


public/video-ui/src/slices/video.ts
```diff
@@ -20,6 +20,9 @@ const video = createSlice({
       state.video = { ...blankVideoData, ...payload };
     },
     setPublishedVideo: (state, { payload }: PayloadAction<Video>) => {
+      state.publishedVideo = { ...blankVideoData, ...payload };
```
SHession, line 23: This is now only used by:

 https://github.com/guardian/media-atom-maker/blob/dba89c1357da55018a51b215262ba03b7e6a30a9/public/video-ui/src/actions/VideoActions/getPublishedVideo.ts#L9-L11.

dblatcher (approved):

PR #1346, opened 2025-09-25
add first frame jpeg output to media convert

gnm-notona:
## Accessibility

### DIFF
--- a/uploader/src/main/resources/cfn-template.yaml
+++ b/uploader/src/main/resources/cfn-template.yaml
@@ -254,6 +254,14 @@ Resources:
                         Bitrate: 160000
                         CodingMode: CODING_MODE_2_0
                         SampleRate: 44100
+              - ContainerSettings:
+                  Container: RAW
+                VideoDescription:
+                  CodecSettings:
+                    Codec: FRAME_CAPTURE
+                    FrameCaptureSettings:
+                      MaxCaptures: 1
+                      Quality: 95
             OutputGroupSettings:
               Type: FILE_GROUP_SETTINGS
               FileGroupSettings: { }


uploader/src/main/resources/cfn-template.yaml
```diff
@@ -254,6 +254,14 @@ Resources:
...
                         CodingMode: CODING_MODE_2_0
                         SampleRate: 44100
+              - ContainerSettings:
+                  Container: RAW
+                VideoDescription:
+                  CodecSettings:
+                    Codec: FRAME_CAPTURE
+                    FrameCaptureSettings:
+                      MaxCaptures: 1
+                      Quality: 95
```
Fweddi, line 264: This is in the MP4 output group, do we need one for the M3U8 as well?

app/model/commands/ActiveAssetCommand.scala
```diff
@@ -15,11 +15,14 @@ case class ActiveAssetCommand(
...
   youtube: YouTube,
   user: PandaUser,
-  awsConfig: AWSConfig
+  awsConfig: AWSConfig,
+  imageUtil: S3ImageUtil
 ) extends Command with MediaAtomImplicits with Logging {
 
   type T = MediaAtom
 
+  lazy val mediaConvertBucket: String = awsConfig.getMandatoryString("aws.mediaconvert.destinationBucket")
```
Fweddi, line 24: Is this new? Has it been added to CODE and PROD configs?

app/model/commands/ActiveAssetCommand.scala
```diff
@@ -49,4 +57,43 @@ case class ActiveAssetCommand(
       AssetVersionConflict
     }
   }
+
+  private [commands] def firstFrameImageName(mp4Name: String): String = {
```
Fweddi, line 61: TIL `[package]` = _qualified_ private

private to the `commands` package rather than just the `ActiveAssetCommand` class 👍

app/util/S3ImageUtil.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    imageUrl match {
+
+      case s3Url(httpOrigin, s3Key) =>
+        val s3ImageAsset = getS3ImageAsset(s3Bucket, s3Key, httpOrigin)
+        s3ImageAsset.map { asset =>
+          Image(
+            assets = List(asset),
+            master = Some(asset),
+            mediaId = "",
```
Fweddi, line 35: The Thrift model unfortunately expects a `mediaId` here: https://github.com/guardian/content-atom/blob/main/thrift/src/main/thrift/atoms/shared.thrift#L197. Setting to empty string seems the next best thing - and avoids confusion where we try to find this image in Grid.

app/model/commands/ActiveAssetCommand.scala
```diff
@@ -49,4 +57,43 @@ case class ActiveAssetCommand(
...
+   * use a default image generated from the first frame of video if:
+   *  - the new asset version to activate is self-hosted
+   *  AND
+   *  - a first-frame image is available
+   *  AND (
+   *    - the current posterImage is empty
+   *    OR
+   *    - the current posterImage is using a default image
+   *  )
+   */
```
Fweddi, line 76: Clever. Some implicit logic going on here for the end user but I think the right amount that we don't need to communicate it to them.

Fweddi (commented):
I can't get this to work on CODE - might be a config issue?

uploader/src/main/resources/cfn-template.yaml
Fweddi, line 264: Serving this high quality directly to users might cause issues. If the images end up in `guim` that's okay, as the Fastly resizer can handle them. Do we know where these images end up?

Of course, I imagine in most cases we'll be serving these images internally (as previews for self-hosted videos in Media Atom Maker and Fronts Tool), so it's not an issue in these cases, but if we end up using it as a 'first frame' to load for poor internet conditions we would probably want the Fastly resizer to be doing its work.

gnm-notona: The MP4 section is for creating the MP4 output, so we can tag the jpeg first frame capture on to that process.  I don't know if the M3U8 section supports it too, but we only need one jpeg

app/model/commands/ActiveAssetCommand.scala
gnm-notona: It's not a new setting but it needed to use the value here.

app/util/S3ImageUtil.scala
gnm-notona: Initially I had the atom Id, not realising it is used by the Grid popup to reference a Grid image.  So we're saying these images aren't grid images at all.   I had to tweak the grid popup in js to open the search page if the media id is empty.

They end up in the uploads-origin.guim.co.uk bucket which is the destination for media convert and the origin for https://uploads.guim.co.uk

uploader/src/main/resources/cfn-template.yaml
Fweddi: might be worth adding a comment explaining this difference in the cloudformation

common/src/main/scala/com/gu/media/model/VideoAsset.scala
```diff
@@ -19,4 +19,5 @@ object VideoAsset {
   implicit val formatYouTube: Format[YouTubeAsset] = Jsonx.formatCaseClass[YouTubeAsset]
   implicit val formatSelfHosted: Format[SelfHostedAsset] = Jsonx.formatCaseClass[SelfHostedAsset]
   implicit val format: Format[VideoAsset] = Jsonx.formatSealed[VideoAsset]
+  val firstFrameImageSuffix = ".0000000.jpg"
```
Fweddi, line 22: might want to move this into `VideoSource` when you've rebased from main (that's where the captions suffix is now stored)

Fweddi (approved):
Awesome stuff. Thanks to @andrew-nowak for setting us on this path (https://github.com/guardian/media-atom-maker/pull/1219)!

Some context for posterity as to why we're doing this.

Since the release of loops, users have stopped adding images to self-hosted videos:
Image

The reason why they're doing this is:
(1) we don't use the images on our platforms (loops autoplay)
(2) it's one less thing to do (loops are more time-sensitive than other videos)

However, this means we can't easily tell which loop is which in Media Atom Maker or the Fronts Tool (they are all blank images). This PR fixes that, and allows us in theory to provide a 'first frame' to platforms for users with slow internet.

Fweddi:
> I can't get this to work on CODE - might be a config issue?

Needed https://github.com/guardian/editorial-tools-platform/pull/984 👍

PR #1349, opened 2025-09-29
Fixes create video failure messaging and behaviour

SHession:
## What does this change?

This PR resolves an issue which would causes furniture data to become inaccessible after a save failure on atom creation:

| Before | After |
|--------|--------|
|  |  | 

The new behaviour allows a user to reattempt a save but also save the form inputs if needed. 

## How to test

Attempt to make a save when creating a new atom. Use the browser tools to cause the save to fail i.e. going offline or removing auth cookies. Observe the new functionality allows you to reattempt a save. 

User's don't lose their changes entirely and can attempt to save an atom again.

### DIFF
--- a/public/video-ui/src/actions/VideoActions/createVideo.ts
+++ b/public/video-ui/src/actions/VideoActions/createVideo.ts
@@ -14,6 +14,9 @@ export function createVideo(video: Video) {
         dispatch(setSaving(false));
         dispatch(setVideo(res));
       })
-      .catch(error => dispatch(showError('Could not create video', error)));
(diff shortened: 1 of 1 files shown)


public/video-ui/src/actions/VideoActions/createVideo.ts
SHession, line 8: Notably this is only used in one place:

https://github.com/guardian/media-atom-maker/blob/4a09d15ed86097090559165b7b279b749bfd5e4a/public/video-ui/src/pages/Video/index.jsx#L62-L67

Issue #1262, opened 2025-08-19
Reudx Toolkit


SHession:
## Problem

Media Atom Maker uses Redux to manage it's application state. This recommended best practices for Redux have moved on since it was developed and newer ways with better supported typing are now available. After exploring #1249, we would like to progress moving the codebase over to RTK where possible. 

## Solution

Using the [migration guide](https://redux.js.org/usage/migrating-to-modern-redux) we can explore moving Redux actions, reducers and thunks over to RTK.

This issue is a guide for engineers getting up to speed on the project and intended to give a starting point. We should also consider the design of parts we change and consider how and where improvements can be effectively made. 

### Good ones to start on (rough order of difficulty)

- [ ] [Config Reducer](https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/reducers/configReducer.js)
- [ ] [Path Reducer](https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/reducers/pathReducer.js)
- [ ] [Search term Reducer](https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/reducers/searchTermReducer.js)
- [ ] [Videos Reducer](https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/reducers/videosReducer.js)

### Key reducers

- [] [Video Reducer](https://github.com/guardian/media-atom-maker/blob/main/public/video-ui/src/reducers/videoReducer.js)

### Exploration work

It would be nice to consider how we might adopt [RTK Query](https://redux.js.org/usage/migrating-to-modern-redux#data-fetching-with-rtk-query) for some of the common external state management done. AFAIK this would need to be done in conjunction with a move toward functional components.

PR #1353, opened 2025-10-01
Add typing to all workflow actions

SHession:
## What does this change?

This PR continues the work of #1262, moving the Workflow related actions to TypeScript and into the workflow slice. This should make the code a little cleaner and improve TypeSafety. 

This PR also fixes a bug where a save failure to the Workflow fields would still transition from the edit state to the readonly state. 

## How to test

This PR should be a no-op. Key flows to test are adding Workflow tracking to a new atom and editing the various workflow fields on an existing one.


public/video-ui/src/slices/workflow.ts
```diff
@@ -1,85 +1,118 @@
...
-    'Cannot get Atom status in Workflow',
-    error
-  );
-}
+export const trackInWorkflow = createAsyncThunk<unknown, WorkflowDetails>(
+  'workflow/trackInWorkflow',
+  (workflowDetails, { dispatch }) =>
+    WorkflowApi.trackInWorkflow(workflowDetails).catch(err => {
+        dispatch(showError('Failed to track Atom in Workflow', err));
+        throw err;
```
SHession, line 96: Throwing an error here means we can prevent the editor from transitioning to the readonly state.

public/video-ui/src/util/setupStore.ts
```diff
@@ -1,13 +1,47 @@
...
+import uploads from '../slices/uploads';
+import path from '../slices/path';
+import pluto from '../slices/pluto';
+import workflow from '../slices/workflow';
+import targeting from '../slices/targeting';
 
 export function setupStore() {
   return configureStore({
-    reducer: reducers,
+    reducer: {
```
SHession, line 25: Moving the reducers here is a small improvement not strictly related to the goals of this PR

waisingyiu (approved):
Nice work!  Happy to approve.  Tested on CODE and it works great!

public/video-ui/src/services/WorkflowApi.ts
```diff
@@ -278,7 +279,7 @@ export default class WorkflowApi {
     });
   }
 
-  static async updateNote({ id, note }: {id: string, note: string | null}): Promise<ApiResponse<number> | null> {
```
emdash-ie, line 281: What’s the rationale for this change? I _think_ I chose `null` over `undefined` because that was what I observer in the redux store, but are you confident that the type passed is actually `undefined`?

public/video-ui/src/slices/workflow.ts
```diff
@@ -1,85 +1,118 @@
...
-        }
+export const getStatus = createAsyncThunk<
+  FlatStub<string, string> | undefined,
+  { id: string }
+>('workflow/getStatus', (video, { dispatch }) =>
+  WorkflowApi.getAtomInWorkflow(video).catch((error: Response | any) => {
+    if (error instanceof Response) {
+      if (error.status !== 404) {
+        return void dispatch(
+          showError('Cannot get Atom status in Workflow', error)
```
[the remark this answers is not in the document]
SHession: Any suggestions on what would be a more appropriate message?

public/video-ui/src/services/WorkflowApi.ts
emdash-ie: Ah I see! Thanks for the explanation :)

public/video-ui/src/slices/workflow.ts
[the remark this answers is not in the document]
emdash-ie: Hmmm, not sure. I was thinking something like “unexpected error response from workflow” here and “failed to parse workflow json” below, but since these are user-facing errors I guess that might be too much detail? (On the other hand, it’s already not a great user-facing error situation since we’re not suggesting any action for the user to take, and the more detail would probably help with debugging if these ever came up)

PR #1360, opened 2025-10-02
replace riff-raff-artifact with actions-riff-raff

andrew-nowak:
Replaces usage of sbt-riffraff-artifact with actions-riff-raff, as well as upgrading a couple of the other actions.


bryophyta (approved):
this looks good to me! 🤩

PR #1369, opened 2025-10-07
Run eslint for .jsx and .tsx files

bryophyta:
## What does this change?

I realised that the `yarn lint` command was only applying to files with `.js` and `.ts` extensions. This PR removes the list of extensions from the lint command, and adds a new list to the `.eslintrc.js` file, which now includes `.jxs` and `.tsx` as well.

I subsequently ran eslint with the `--fix` flag, which fixed a number of issues (largely formatting ones). There were then a handful of errors remaining, which I've fixed manually. I've tried to keep the manual fixes as minimal as possible, so for the deprecated methods I've just updated their names to `UNSAFE_*` rather than trying to refactor the component to not use the method anymore. (Refactoring will be required before upgrading to React 17, just fyi.)

## Accessibility

### DIFF
--- a/public/video-ui/src/components/TagSearch/TagSearch.jsx
+++ b/public/video-ui/src/components/TagSearch/TagSearch.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 class TagSearch extends React.Component {
   constructor(props) {
     super(props);
+    this.listNodeRef = React.createRef();
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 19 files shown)


public/video-ui/src/components/TagSearch/TagSearch.jsx
```diff
@@ -12,10 +13,9 @@ class TagSearch extends React.Component {
       nextProps.selectedTagIndex !== null &&
       prevProps.selectedTagIndex !== nextProps.selectedTagIndex
     ) {
-      const listNode = this.refs.list;
-      const elementHeight = listNode.children[0].offsetHeight;
-      if (listNode) {
-        listNode.scrollTop =
+      if (this.listNodeRef.current) {
```
bryophyta, line 16: This was the least trivial fix. I hadn't looked at this component before, but it seems that this logic is in place to make sure that the currently focused tag element is in view when you're navigating through the list using the arrow keys:

Image

I've tested the new code locally and the scroll behaviour seems to be the same as it was before 👍

PR #1372, opened 2025-10-08
Adds a client side architecture markdown document

SHession:
## What does this change?

Adds a client side architecture document, which hopefully speaks for itself. 

## How to test

Does the content make sense? Is there anything else we can add?

Future contributors can understand a bit about how the client is structured and how it can be improved

### DIFF
--- a/docs/11-client-side-architecture.md
+++ b/docs/11-client-side-architecture.md
@@ -77,8 +77,10 @@ significant benefits to the Developer Experience.
 
 ### Test coverage
 
-The client is sparsely tested, but some examples do exist. When rewriting components, it would be worth adding test
-coverage to ensure consistent behaviour and help document business logic.
+The client is sparsely tested, but
+some [examples](https://github.com/guardian/media-atom-maker/blob/ccfd5c194d2c76e55d0c4b468d91e252716a0878/public/video-ui/src/components/FormFields/TagPicker.test.jsx)
+do exist. When rewriting components, it would be worth adding test coverage to ensure consistent behaviour and help
+document business logic.
 
 ### Improving telemetry
 


bryophyta (approved):
😻😻

docs/11-client-side-architecture.md
```diff
@@ -0,0 +1,88 @@
...
+
+Continuing to improve the TypeScript coverage will make the project easier to work on in the future and provides
+significant benefits to the Developer Experience.
+
+[Setting TypeScript to be `strict`](https://www.typescriptlang.org/docs/handbook/2/basic-types.html#strictness) would be a sensible improvement and help identify further potential bugs.
+
+### Test coverage
+
+The client is sparsely tested, but some examples do exist. When rewriting components, it would be worth adding test
+coverage to ensure consistent behaviour and help document business logic.
```
waisingyiu, line 81: Would it be worth putting an example (such as this one https://github.com/guardian/media-atom-maker/blob/ccfd5c194d2c76e55d0c4b468d91e252716a0878/public/video-ui/src/components/FormFields/TagPicker.test.jsx) here?

docs/11-client-side-architecture.md
```diff
@@ -0,0 +1,88 @@
...
+### TypeScript Coverage
+
+The project's overall TypeScript coverage has increased significantly, roughly doubling during the course of this work.
+This has helped to identify potential bugs, improve the understanding and readability of the code, and generally make
+refactoring safer.
+
+## Future development
+
+When developing the client further, a few key areas should be considered for improvement:
+
```
waisingyiu, line 64: I remember RTK Query was suggested at some point.  Is it something useful for future developers to think about?

waisingyiu (approved):
It looks great to me!  Happy to approve.  Just a couple of non-blocking questions.

docs/11-client-side-architecture.md
waisingyiu: No, not really.  I didn't look into RTK Query.  If we don't have any specific recommendations, we may just skip it.

PR #1381, opened 2025-10-14
Add and run scalafmt

bryophyta:
## What does this change?

Runs scalafmt across the codebase, and adds a `.git-blame-ignore-revs` file with the commit hash so that git blame won't be too messed up when viewing in contexts that respect the ignore file.

## Accessibility


bryophyta:
> we should add a check step to CI though so we don't build up another big diff from scalafmt's ideals

good shout, done!

PR #1387, opened 2025-10-16
Add Iconik store and ingestion endpoints

bryophyta:
## What does this change?

Adds back end to handle requests coming from Iconik via the ingestion lambda (see https://github.com/guardian/media-atom-maker/pull/1386).

Follows the existing patterns used for storing data from Pluto (which Iconik is replacing), but adds an extra layer to the hierarchy: projects still belong to commissions, but now commissions belong to working groups, too.

- [x] **nb.** Depends on https://github.com/guardian/editorial-tools-platform/pull/992
- [x] **nb.** Also depends on https://github.com/guardian/editorial-tools-platform/pull/994

## How to test

- [x] Deploy to CODE
- [x] Run pre-existing test script (`./scripts/test-pluto-lambda-CODE.js`) to check old functionality still works
- [x] Run new test script to test new message actions: `./scripts/test-iconik-upsert-CODE.js`

## Accessibility


app/controllers/IconikController.scala
```diff
@@ -0,0 +1,101 @@
...
+    Ok(Json.toJson(iconikCommissions))
+  }
+
+  def getCommissions() = APIAuthAction {
+    val iconikCommissions = stores.iconikCommissionStore.list()
+    Ok(Json.toJson(iconikCommissions))
+  }
+
+  def getCommission(commissionId: String) = APIAuthAction {
+    stores.iconikCommissionStore.getById(commissionId) match {
```
lindseydew: We could remove one branch in the pattern match with a fold: 

def getCommission(commissionId: String) = APIAuthAction {
  stores.iconikCommissionStore.getById(commissionId)
    .fold {
      log.warn(s"no commission with ID $commissionId")
      NotFound
    } {
      case Right(commission) =>
        Ok(Json.toJson(commission))

      case Left(error) =>
        log.error(
          s"failed to get iconik commission with ID $commissionId",
          error
        )
        ServiceUnavailable
    }
}

common/src/main/scala/com/gu/media/iconik/IconikCommissionDataStore.scala
```diff
@@ -0,0 +1,62 @@
...
+      commissionId: String
+  ): Option[Either[DynamoReadError, IconikCommission]] = {
+    log.info(s"getting commission $commissionId")
+    scanamo.exec(table.get("id" === commissionId))
+  }
+
+  def list(): List[IconikCommission] = {
+    scanamo
+      .exec(table.scan())
+      .collect {
```
lindseydew: I know this is how the pluto implementation is working, but these feels a bit strange to side effect like this, especially when other methods in this class are pure. 
Another approach could be to add a sequence helper method: 
```
 def sequence[E, A](eithers: List[Either[E, A]]): Either[E, List[A]] =
    eithers.foldRight(Right(Nil): Either[E, List[A]]) {
      case (Right(a), Right(acc)) => Right(a :: acc)
      case (Left(e), _)           => Left(e)
      case (_, Left(e))           => Left(e)
    }
    ```
    and then this method would become: 
    
     def getByCommissionId(commissionId: String): Either[DynamoReadError, List[IconikProject]] = {
    sequence(scanamo
      .exec(commmissionIndex.query("commissionId" === commissionId))).map(_.sortBy(_.title))
  }

which would return a signature of [Either[DynamoReadError, List[IconikCommission]]

PR #1390, opened 2025-10-21
Catch errors from the createUpload call

bryophyta:
## What does this change?

Adds a `.catch` block for the call to `createUpload`. This function makes an API call and at the moment if that fails (e.g. if you're offline) there's no indication for the user that something has gone wrong.

## How to test

- Load the Upload page
- Set connection to offline in dev tools
- Upload a file
- In `main` nothing visibly happens; in this branch you get an error message.

## Accessibility

### DIFF
--- a/public/video-ui/src/slices/s3Upload.ts
+++ b/public/video-ui/src/slices/s3Upload.ts
@@ -46,12 +46,9 @@ export const startVideoUpload = createAsyncThunk<
       const progress = (completed: number) =>
         dispatch(s3UploadProgress(completed));
 
-      return uploadParts(upload, upload.parts, file, progress)
-        .then(() => dispatch(setS3UploadStatusToComplete()))
-        .catch(err => {
-          dispatch(showError(errorDetails(err), err));
-          dispatch(setS3UploadStatusToError());
-        });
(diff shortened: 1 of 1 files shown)


public/video-ui/src/slices/s3Upload.ts
```diff
@@ -39,19 +39,24 @@ export const startVideoUpload = createAsyncThunk<
...
-      });
-  })
+      return uploadParts(upload, upload.parts, file, progress)
+        .then(() => dispatch(setS3UploadStatusToComplete()))
+        .catch(err => {
+          dispatch(showError(errorDetails(err), err));
+          dispatch(setS3UploadStatusToError());
+        });
+    })
+    .catch(err => {
```
andrew-nowak: if we have this, can we remove the catch on the lines just above this? I think the promise rejection should fall through to here if not caught there, and we handle it in the same way

andrew-nowak (approved):

public/video-ui/src/slices/s3Upload.ts
bryophyta: thanks!

PR #1393, opened 2025-10-23
Disable delete button when processing, and fix update after upload

bryophyta:
**nb.** If you review this PR commit-by-commit, please note that I've tried to take small incremental steps in each commit, but this means it zig-zags a bit in order to get to the final destination.

## What does this change?

This branch aims to fix a couple of bugs related to deleting assets:

1. The 'delete' button for an asset is currently not disabled when it should be (i.e. when the client app is not in a position to make a successful delete request)
2. The client is not updating its state after successful upload, so that it will _never_ be in a position to delete an asset that's just been uploaded, until something else triggers a data sync (e.g. navigating away from the upload view and back, or refreshing the page)

### To replicate the bugs on `main`

- Upload a new video asset
- Notice that once the 'delete' button becomes available, it continues to be in an enabled state while the upload is processing
- Wait for a while after the upload has finished, without refreshing the page or similar, and click the 'delete' button. Observe that a failing `DELETE` request will be made.

### History

Problem (2) was fixed in PR #1284, by having the client fetch the 'video' state again once the 'post-processing' phase in the S3 uploader store is complete. However, at some point since then there has been a regression in this behaviour.

Currently on `main` the client doesn't know that there's still work in progress until the return of a `getUploads` call, and marks the post-processing as being finished before that call returns. The `getUploads` call is started at point 1 on the screenshot below, and resolved at point 3. But the post-processing state has already been entered and resolved by this point, at 2.
 
image

I haven't been able to isolate which commit introduced the regression from #1284, and there were some complications in the code due to the use of class components in VideoTrail, so it felt reasonable to rework the logic here as part of a refactor to make it a functional component.

### Solution

To address problem (1) above, I've made the delete button's enabled/disabled state dependent on the underlying state that the client needs in order to make a successful delete request. If the video store doesn't know about the asset in question, it can't construct a `DELETE` request for it, and so the button should be disabled.

To address problem (2), I've made it so that:

- every upload/delete action in the S3 store triggers a `getUploads` call
- if any uploads are marked as `processing`, and the VideoTrail component is mounted, then it will trigger further polling to `getUploads` until all uploads have resolved, at which point 
- it will trigger a `getVideo` call, at which point the back end should be in a position to return all of the updated assets, meaning the video store on the client has what it needs to construct a `DELETE` request.

This is basically the same as the strategy that was being used before, but the polling logic should be a bit more reliable now that it's in a `useEffect` hook, and I've tried to move all of the internal state transitions into the S3 upload store instead of the VideoTrail component.

## How to test

- [x] Run locally or in CODE
- [x] Upload a YouTube video, and then:
    - [x] **i.** notice that the delete button is initially disabled for the processing upload asset
    - [x] **ii.** then eventually becomes enabled
    - [x] **iii.** if you press the delete button once it's enabled, it will successfully trigger a DELETE operation
- [x] Repeat steps (i), (ii) and (iii) with self-hosted videos, and with uploading and deleting subtitles.

## Images

| Before | After |
|-|-| 
|an asset is uploading, and the 'activate' button is disabled, but the 'delete' button is enabled|an asset is uploading, and both the 'active' and 'delete' buttons are disabled|

## Accessibility

### DIFF
--- a/public/video-ui/src/components/VideoUpload/VideoAsset.tsx
+++ b/public/video-ui/src/components/VideoUpload/VideoAsset.tsx
@@ -60,7 +60,7 @@ function AssetControls({
 
   const cannotDeleteAsset =
     cannotActivateAsset ||
-    !video.assets.find(videoAsset => videoAsset.version.toString() === assetId);
+    !video.assets.some(videoAsset => videoAsset.version.toString() === assetId);
 
   const userCircle = (
     <div className="video-trail__presence_indicator" title={user}>


gnm-notona (approved):
We've been aware in Fronts & Curation team of the bugs in getting status updates during subtitle uploads, so I've tested that bit in CODE and this seems to be working.  Very happy to see this rewritten in TS.  I'm not really a frontend person, so tentatively approve, but would recommend getting someone else like @SHession to have a look.

PR #1399, opened 2025-10-31
Upgrade AWS JS SDK to v3

andrew-nowak:
## What does this change?

Convert the AWS JS SDK usage to v3. The only real usage here is for uploading chunks of the video to S3 from the user's browser. (This is a bit niche, I think we could replace the browser SDK usage/credential passing with [presigned urls](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html) but that'll be a fairly involved change, so for now I'm simply replacing like-for-like.)

## How to test

Upload some videos using this branch.

No more remaining usage of the v2 sdk, and the js bundle should be a lot smaller with this nicely modularised version of the sdk (currently 3.9MB reduced to 2.3MB!)

Missed edge cases in the way that the upload works.

## Accessibility

### DIFF
--- a/public/video-ui/src/services/UploadsApi.js
+++ b/public/video-ui/src/services/UploadsApi.js
@@ -62,16 +62,16 @@ function getS3(region, credentials) {
   const awsCredentials = {
     accessKeyId: temporaryAccessId,
     secretAccessKey: temporarySecretKey,
-    sessionToken: sessionToken,
+    sessionToken: sessionToken
   };
 
   return new S3({
     credentials: awsCredentials,
     requestHandler: XhrHttpHandler.create({
-      requestTimeout: 240_000,
+      requestTimeout: 240_000
     }),
     region: region,
-    useAccelerateEndpoint: true,
+    useAccelerateEndpoint: true
   });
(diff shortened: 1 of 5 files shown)


public/video-ui/src/services/UploadsApi.js
```diff
@@ -51,49 +50,73 @@ function getCredentials(id, key) {
...
-      Body: slice,
-      ACL: 'private',
-      Metadata: { original: file.name }
-    };
-    const request = s3.upload(params, httpOptions);
-
-    request.on('httpUploadProgress', event => {
-      progressFn(part.start + event.loaded);
+      Metadata: { original: file.name },
+      Body: body
```
lindseydew: Is there a behavioural change here in passing in the body rather than the `slice` variable above? Was that intentional?

lindseydew: Ah right ok, makes sense. Small point and I know this was how it was before, but there could be a slightly more helpful variable name than slice here too.

lindseydew (approved):

PR #1405, opened 2025-11-07
Add Iconik project read and write UI to app

bryophyta:
## What does this change?

* **Back end:** Add Iconik metadata to the models used to persist Media Atoms in the database.
* **Front end:**
  * Add API helpers for getting Iconik projects, commissions, etc.
  * Add a Redux store slice to hold Iconik info.
  * Add read-only 'Iconik' tab to the Video page, following the example for Pluto.
  * Add a new form to the upload page, allowing users to select Iconik Working Group, Commission and Project for the Atom that they're editing.
  * Each project belongs to one commission, and each commission belongs to one working group, so the logic of the selection form reflects this: if you select a working group, it'll clear any existing commissions or projects, and load the commissions for the new working group as options, etc.

## How to test

- Run locally or in CODE
- [x] Go to a video, check that the 'Iconik' tab is empty (assuming it hasn't been tested with already).
- [x] Go to the 'upload' view and set a working group, commission and project in the form.
  - [x] Save and reload the page; verify that the choices are persisted
  - [x] Select a new working group and verify that the commission and project are blanked out, and new commissions (the ones for the new working group) are loaded. Ditto for selecting a new commission.
  - [x] 'Reset' button works to restore state to the previous saved state.
  - [x] 'Remove' button removes all Iconik fields from the UI, and saves the new state.
- [x] Buttons show when appropriate:
    - [x] 'reset' only enabled when the current UI state differs from the saved state
    - [x] 'save' only enabled when the UI state differs from saved state, AND there's a project selected (otherwise it's not a valid state to save).
    - [x] 'remove' only enabled when non-blank state is saved.

https://github.com/user-attachments/assets/df737dd1-7212-422a-b078-52de5cd96c2f

## Accessibility

### DIFF
--- a/public/video-ui/src/pages/Upload/index.tsx
+++ b/public/video-ui/src/pages/Upload/index.tsx
@@ -91,9 +91,11 @@ export const VideoUpload = (props: { params: { id: string } }) => {
                 <PlutoProjectPicker video={store.video} />
               </div>
             </div>
-            <div>
-              <IconikProjectPicker video={store.video} />
-            </div>
+            {store.config.showIconik && (
+              <div>
+                <IconikProjectPicker video={store.video} />
+              </div>
+            )}
             <YoutubeUpload
               video={store.video}
               categories={store.youtube.categories}
--- a/public/video-ui/src/pages/Video/tabs/Iconik.tsx
+++ b/public/video-ui/src/pages/Video/tabs/Iconik.tsx
@@ -1,7 +1,9 @@
 import React, { useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
 import { Tab, TabPanel } from 'react-tabs';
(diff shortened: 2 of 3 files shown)


public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+      dispatch(resetCommissions());
+    }
+    if (commission) {
+      dispatch(fetchIconikProjects(commission));
+    } else {
+      dispatch(resetProjects());
+    }
+  }, [commission, dispatch, workingGroup]);
+
+  const hasBeenEdited =
```
lindseydew: Is this just computed once on load? Or should this be wrapped in a callback and have the form fields as state values?

PR #1415, opened 2025-11-21
feat: Update content bundle map

akash1810:
## What does this change?
Updates the content bundle introduced in https://github.com/guardian/media-atom-maker/pull/464, adding a new `full-story-podcast` entry as requested by Tom Duffett.


akash1810:
> I'm not sure what a bundle is in Youtube context but LGTM

From memory (read: I might be wrong!) we use a "bundle" to control what category of adverts can be shown against the YouTube videos. For example, only sport related adverts will show against sport content.

PR #1503, opened 2026-03-05
Bundling tags cleanup

andrew-nowak:
## What does this change?

Two small cleanups around the content bundling tags after #1500 and #1410.

First remove the lookup table from the Scala code, since that's now not needed or used.
Then update the `addOrDropBundlingTags` function to take named parameters, since it's both using two parameters of the same type (which is easy to confuse) and a boolean parameter (which is easy to get mixed up what `true` vs. `false` means) 

## Accessibility


bryophyta (approved):
nice, this looks good to me!

PR #1528, opened 2026-04-08
Order video preview by mime type and size

abeddow91:
## What does this change?
Orders video sources by mimetype and size, in descending order. This is because the previous implementation of finding the first mp4 and using that did not account for multiple resolutions. As a consequence, the lower resolution 480w version was being displayed in preview.

This PR ports the DCR implementation introduced here https://github.com/guardian/dotcom-rendering/commit/36f340f8d2912e7b51cee0c6212706fa789c70db

Deploying to CODE to ensure 720h videos are played when available instead of 480w

Users see the highest resolution video

This will always now show the highest quality available. This therefore hides any other quality videos which may still be representative of video quality on platform, depending on device, and so preview is now not a complete representation of the video atom. A mitigation for this would be to allow the user to view multiple video resolutions. This would be future work and is not included in this PR. 

### Before
Screenshot 2026-04-08 at 16 25 23

### After
Screenshot 2026-04-09 at 11 04 48

## Accessibility

### DIFF
--- a/public/video-ui/src/components/VideoPreview/VideoPreview.jsx
+++ b/public/video-ui/src/components/VideoPreview/VideoPreview.jsx
@@ -24,7 +24,7 @@ export default class VideoPreview extends React.Component {
     }
 
     const sources = active.map(asset => {
-      return { src: asset.id, mimeType: asset.mimeType };
+      return { src: asset.id, mimeType: asset.mimeType, height: asset.height, width: asset.width };
     });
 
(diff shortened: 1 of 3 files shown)


Georges-GNM (approved):
Can see this is up on code, looks good to me. 

Tiny thing, but in case you haven't seen - CI's complained of a missing semicolon in VideoEmbed.tsx on line R73:

Image

PR #1530, opened 2026-04-09
Add has audio flag to atom

abeddow91:
## What does this change?
Stores hasAudio boolean flag on each asset. This will allow consumers downstream to read if the atom has audio without having to detect from audio tracks. 

The hasAudio state is derived from the same state that is used to determine if audio tracks should be stripped from the asset during transcoding, which is detected in sendToTranscoderV2 using ffmpeg. 

The state machine is event-driven, rather than polling-based. It uses [TaskTokens](https://docs.aws.amazon.com/sdk-for-swift/latest/api/awssfn/documentation/awssfn/sendtaskfailureinput/tasktoken/) to pause execution whilst the transcoder step is processed. `TranscoderComplete` is then triggered and provides the output for the step. More detail on the implementation can be found here: https://github.com/guardian/media-atom-maker/pull/1487. As such, input is immutable during the state machine execution and can only be updated on output, once the MediaConvert job has completed. In order to persist the state of audio tracks discovered prior to transcoding, this is stored as user metadata on the MediaConvert job, which is carried through to the completion event and read by TranscoderComplete to set hasAudio on each asset.

- Deployed this branch to MAM to CODE
- Deployed https://github.com/guardian/content-api/pull/3336 to Porter
- Tested uploading videos with and without audio 
- Confirm that the atom has the correct output via querying the atom in Dynamodb and inspecting the step function output. 
- Confirming that the atom output json had the correct value on hasAudio in Cerebro. Accessing Cerebro requires assistance from the Articles and Publishing team.

Platforms can read if an atom has audio and render controls as appropriate.

## Images

Screenshot 2026-04-21 at 12 41 00

## Accessibility

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ import scala.sys.process.*
 val scroogeVersion = "4.12.0"
 val awsV2Version = "2.42.24"
 val pandaVersion = "13.0.0"
-val atomMakerVersion = "9.0.0"
+val atomMakerVersion = "10.0.0"
 val typesafeConfigVersion =
   "1.4.0" // to match what we get from Play transitively
 val scanamoVersion = "1.0.0-M28"
--- a/uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
@@ -139,14 +139,20 @@ class AddAssetToAtom
     val outputGroupsSettings = JobSettingsBuilder.getOutputGroups(hasAudio)
     val outputGroupsResults = completeEvent.detail.outputGroupDetails
 
-    outputGroupAssets(outputGroupsSettings, outputGroupsResults, version) ++
-      outputAssets(outputGroupsSettings, outputGroupsResults, version)
+    outputGroupAssets(
+      outputGroupsSettings,
+      outputGroupsResults,
+      version,
+      hasAudio
+    ) ++
+      outputAssets(outputGroupsSettings, outputGroupsResults, version, hasAudio)
   }
 
   private def outputGroupAssets(
       outputGroupsSettings: List[OutputGroupDefinition],
       outputGroupsResults: List[MediaConvertOutputGroupDetails],
-      version: Long
+      version: Long,
+      hasAudio: Boolean
   ) =
     for {
       (settings, results) <- outputGroupsSettings.zip(outputGroupsResults)
@@ -167,13 +173,15 @@ class AddAssetToAtom
       // HLS playlists can include multiple renditions with different resolutions, so we can't provide dimensions for the asset at this level
       dimensions = None,
       // HLS playlist videos may have different resolutions, but they should have the same aspect ratio.
-      ratio(results.outputDetails.head)
+      ratio(results.outputDetails.head),
+      hasAudio = Some(hasAudio)
     )
 
   private def outputAssets(
       outputGroupsSettings: List[OutputGroupDefinition],
       outputGroupsResults: List[MediaConvertOutputGroupDetails],
-      version: Long
+      version: Long,
+      hasAudio: Boolean
   ) =
     for {
       (outputGroupSettings, outputGroupResults) <- outputGroupsSettings.zip(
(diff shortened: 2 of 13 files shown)


build.sbt
```diff
@@ -7,7 +7,7 @@ import scala.sys.process.*
 val scroogeVersion = "4.12.0"
 val awsV2Version = "2.42.24"
 val pandaVersion = "13.0.0"
-val atomMakerVersion = "9.0.0"
+val atomMakerVersion = "10.0.0-PREVIEW.abbump-content-atom.2026-04-16T1312.5fe22378"
```
abeddow91: NB : This will need updating to the prod release before merging

uploader/src/main/scala/com/gu/media/upload/TranscoderComplete.scala
```diff
@@ -65,10 +65,15 @@ class TranscoderComplete
...
+      hasAudio = data.detail.userMetadata
+        .get("hasAudio")
+        .flatMap(_.toBooleanOption)
+
       output = upload.copy(
         metadata = upload.metadata.copy(
           asset = updatedAsset,
-          runtime = existingMetadata.copy(completeEvent = Some(data))
+          runtime = existingMetadata.copy(completeEvent = Some(data)),
+          hasAudio = hasAudio
```
SHession, line 76: After going down a bit of a rabbit hole, this makes sense to me. The step function is using a `TaskToken` to pause execution until the the transcoder step takes place. `TranscoderComplete` is then triggered and provides the output for the step. More detail here: https://github.com/guardian/media-atom-maker/pull/1487

This is not very clear from either the code or the step function UI. It would be nice to make this more obvious.

uploader/src/main/scala/com/gu/media/upload/SendToTranscoderV2.scala
```diff
@@ -58,8 +58,7 @@ class SendToTranscoderV2
       upload.metadata.copy(
         runtime = SelfHostedUploadMetadata(
           jobs = Some(List(jobs))
-        ),
-        hasAudio = Some(hasAudio)
```
SHession, line 62: We're no longer setting `hasAudio` here, presumably because it makes no difference. If that is the case, do we even need to set the other attributes?

uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
```diff
@@ -195,7 +203,8 @@ class AddAssetToAtom
       ThriftPlatform.Url,
       Some(mimeType),
       dimensions(outputResults),
-      ratio(outputResults)
+      ratio(outputResults),
+      hasAudio = Some(hasAudio)
```
SHession: In my test it looks like the subtitle asset also has the `hasAudio` property set:

Image

Furthermore, should we even been creating these files when hasAudio is false?

SHession (commented):
This appears to be working nicely now. I upload two videos using the DEV pipeline and found the hasAudio attribute set as expect on the assets. 

There is a question about subtitles we may wish to address in this PR, it might be preferable to not set the attribute to true or false as it doesn't make much sense.

uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
abeddow91: Thats a good point. Im trying to think if there would be a scenario where the video is silent with no audio tracks but an .srt file has been provided for subtitles. If that were to be true, we would still want to output a .vtt even if the underlying video asset is silent. 

In either instance though, I agree that hasAudio shouldn't exist on .vtt files, so I'll update that in this PR.

uploader/src/main/scala/com/gu/media/upload/SendToTranscoderV2.scala
abeddow91: Yep, I had wondered that. I suspect that we do not need them as they will also be dropped. I'd be happy to test removing them but I think that might be better outside of this PR.

uploader/src/main/scala/com/gu/media/upload/TranscoderComplete.scala
abeddow91: I agree, as someone not overly familiar with this state machines  (and particularly event-driven state machines) this took me a while to figure out. I'll look at formalising this into a doc for future travellers outside of this PR.

abeddow91: In the mean time I've added some more detail to the PR description to better explain this

SHession (approved):
Tested again locally, subtitles no longer have `hasAudio` set. LGTM!

PR #1534, opened 2026-04-15
Explicitly set vtt extension to avoid AWS bug

davidfurey:
Without explicitly setting the vtt extension, vtt files are written to S3 with `.vtt` extension, but the MediaConvert complete event omits the extension. 

We've previously worked around this by re-writing the filepath based on the expected mimetype which works but isn't great.

AWS support identified that if we explicitly set the `Extension` property on our `Output`, the file will be written with the extension and the event will also have the extension. This seems like a nicer fix.


abeddow91 (approved):
Much less brittle, a welcome change!

PR #1536, opened 2026-04-16
Sets self-hosted video duration when activating an asset if available

SHession:
## What does this change?

This PR attempts to improve the way durations are applied to media atoms by persisting the duration determined in the upload step and applying it when an asset is activated.

The AWS transcoder already provides metadata on outputted video's duration, this change persists that data against the assets and applies it during activation. This is similar to how YouTube duration is applied, however, that uses the YouTube API to request the duration.

To facilitate this change, we need to add the duration value to the asset model in the content model repository:

- https://github.com/guardian/content-atom/pull/208
- https://github.com/guardian/atom-maker/pull/142

These have been done as preview release but will need to be merged properly and applied throughout the content pipeline 😵‍💫

This change has been tested locally using the following steps:

1. Create a self-hosted video atom
2. Upload a video
3. Activate the video
4. Observe the atom duration has been set
5. Check CAPI returns the intended duration

Users have to think less about applying durations when switching active assets.

Given the model change will be across the pipeline, it would be sensible to ensure we're applying it consistently. 

## Accessibility

---
  - https://app.asana.com/0/0/1214024821512601

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,13 +66,7 @@ val jacksonOverrides = Seq(
   "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310",
   "com.fasterxml.jackson.module" % "jackson-module-parameter-names",
   "com.fasterxml.jackson.module" %% "jackson-module-scala"
-).map(
-  _ % jacksonVersion
-) :+ "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonAnnotationsVersion
-
-val nettyOverrides = Seq(
-  "io.netty" % "netty-codec-http2" % "4.1.132.Final"
-)
+).map(_ % jacksonVersion) :+ "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonAnnotationsVersion
 
 lazy val common = (project in file("common"))
   .settings(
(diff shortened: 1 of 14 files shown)


SHession:
Tested this on CODE and it is behaving as intended: https://video.code.dev-gutools.co.uk/videos/29adb550-b84f-4197-94fa-0d1b9742b6c2.

The attribute is not currently visible via CAPI, but I believe this is expected and not necessary for the intent of this change.

build.sbt
```diff
@@ -7,7 +7,7 @@ import scala.sys.process.*
 val scroogeVersion = "4.12.0"
 val awsV2Version = "2.42.24"
 val pandaVersion = "13.0.0"
-val atomMakerVersion = "9.0.0"
+val atomMakerVersion = "10.0.0-PREVIEW.abbump-content-atom.2026-04-16T1312.5fe22378"
```
abeddow91: Stating the obvious - this will need to be updated to the prod release before merging.

abeddow91 (approved):
Tested on CODE and can see the atom duration is being updated to the active asset's duration when the asset is activated.

This is approved assuming the atom-maker version is bumped to PROD version prior to release.

As a side note, I can't see the duration persisted on the asset itself in CAPI. Is this expected behaviour? If not, it might point to an issue in the CAPI pipeline

uploader/src/main/scala/com/gu/media/upload/AddAssetToAtom.scala
```diff
@@ -205,16 +205,14 @@ class AddAssetToAtom
...
-  private def supportsAudio(mimeType: String, hasAudio: Boolean) = {
-    mimeType match {
-      case VideoSource.mimeTypeMp4 | VideoSource.mimeTypeM3u8 => Some(hasAudio)
-      case _                                                  => None
-    }
-  }
+  private def whenVideo[T](mimeType: String, value: T): Option[T] =
+    Option.when(
+      mimeType == VideoSource.mimeTypeMp4 || mimeType == VideoSource.mimeTypeM3u8
+    )(value)
```
abeddow91, line 215: Nice! Much more reusable 👏

abeddow91:
Tested on CODE. I can see that only video assets are being stored with a duration and that the atom's duration is being derived from the active asset on activation. LGTM!

PR #1541, opened 2026-04-27
always allow editing embargo, only prevent new schedules

andrew-nowak:
## What does this change?

We've had one atom that was simultaneously published (ie. the Composer page had been published, so a usage was listed) and embargoed. Since it had been published, the UI would not allow users to enter the menu to remove the embargo. 

Update to always allow entering the scheduling/embargoing menu, and only prevent editing of the _launch schedule_. Users can always prevent publication (or remove the embargo).

Tested on CODE. Created a video page, then embargoed the atom. Then launched the composer page to produce a usage. The atom could still be unembargoed to allow publish.

Users can get themselves out of similar sticky situations.


bryophyta (approved):
tested on CODE and it works as described 👍

PR #1543, opened 2026-04-28
Include commissioning only

davidfurey:
## What does this change?

Limits the commissioning desk dropdown to only include commissioning desk tags.

~It also adds types to the tag picker.~

### Before

Note the dcroptout (platformfunctional) tag.

Screenshot 2026-04-28 at 12 30 05

### After

Note the absence of the dcroptout tag.

Screenshot 2026-04-28 at 12 30 19

### DIFF
--- a/public/video-ui/src/components/VideoData/VideoData.jsx
+++ b/public/video-ui/src/components/VideoData/VideoData.jsx
@@ -142,7 +142,7 @@ export default class VideoData extends React.Component {
               isRequired={canonicalVideoPageExists}
               inputPlaceholder="Search commissioning info (type '*' to show all)"
             >
-              <TagPicker disableTextInput />
+              <TagPicker disableTextInput tagSubType="commissioningdesk" />
             </ManagedField>
         }
         {
--- a/public/video-ui/src/components/FormFields/TagPicker.jsx
+++ b/public/video-ui/src/components/FormFields/TagPicker.jsx
@@ -94,7 +94,7 @@ class TagPicker extends React.Component {
         searchResultTags: []
       });
     } else {
-      getTagsByType(this.props.tagManagerUrl, searchText, tagTypes)
+      getTagsByType(this.props.tagManagerUrl, searchText, tagTypes, this.props.tagSubType)
         .then(response => {
 
           const tags = response.data.reduce((tags, {data}) => {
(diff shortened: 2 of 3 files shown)


public/video-ui/src/components/FormFields/TagPicker.tsx
```diff
@@ -8,16 +8,43 @@ import TextInputTagPicker from './TextInputTagPicker';
...
+  };
+  hasWarning: (props: TagPickerProps) => boolean;
+  hasError: (props: TagPickerProps) => boolean;
+  onUpdateField: (newValue: string) => Promise<void>;
+  fieldName: string;
+  editable: boolean;
+  updateSideEffects?: () => void;
+}
+
+class TagPicker extends React.Component<TagPickerProps, any> {
```
SHession: Setting `any` here means we don't getting typing on our state. If we apply it we get a type error: 

Image

public/video-ui/src/components/VideoData/VideoData.jsx
```diff
@@ -142,7 +142,7 @@ export default class VideoData extends React.Component {
               isRequired={canonicalVideoPageExists}
               inputPlaceholder="Search commissioning info (type '*' to show all)"
             >
-              <TagPicker disableTextInput />
+              <TagPicker disableTextInput tagSubType="commissioningdesk" />
```
SHession: Looks like the convention would to add the prop to `ManagedField`, which would require a change there too. I don't have an opinon on whether this is worth doing.

SHession (commented):
I've tested the functionality locally and it looks good. One comment on the TypeScript conversion worth considering.

public/video-ui/src/components/ManagedForm/ManagedField.tsx
```diff
@@ -142,7 +146,7 @@ export class ManagedField extends React.Component {
       className += ' form-element--hidden';
     }
 
-    const hydratedChildren = React.Children.map(this.props.children, child => {
+    const hydratedChildren = React.Children.map(this.props.children, (child: React.ReactElement) => {
```
SHession: Some exceptions being thrown on the atom page, which renders a blank page. Attributed to this function:
Image

SHession (approved):
Tested locally, this is working nicely on the "Commissioning Desks" and "Composer Keywords" inputs

PR #1546, opened 2026-05-06
allow adding videos from some specified "external" channels as imported urls

andrew-nowak:
## What does this change?

New bit of config: allow nominating some channels as "external" - they're not part of the guardian content management group of channels, so we can't manage them through the API, but we can allow them to be imported into an atom.

Tested on CODE - adding the channel ID for the new channel to this new list, then importing to an atom in CODE and doing the work to publish it. All worked fine, despite the channel not yet being linked to the Content Manager in Youtube.


bryophyta (approved):
lgtm!

PR #1548, opened 2026-05-07
Support tagging video atoms

waisingyiu:
## What does this change?

We are looking to add tagging functionality to video atoms as part of our work on supporting experiments that could help us understand how different types of videos on fronts are received by readers. Tags allows data team to distinguish different types of videos in data lake and to draw insights with respect to different video characteristics.

We added a new property for tags to the content atom thrift model in guardian/content-atom#211 and updated the atom maker libraries in guardian/atom-maker#143.

This pull request adds UI to the media atom maker behind a feature switch to allow tagging videos. Tags are mandatory for default videos but optional for other video types.

It makes the following changes specifically:

1. add a new property `AtomTagIds` to the `Video` model in the frontend.
2. update the content atom thrift model by bumping `atom-maker`, add `AtomTagIds` to the `MediaAtom` model in the backend and update the conversion classes on the model.
3. support a query parameter `supportTagging` as a feature flag. The UI change made in this PR is visible only when this parameter is present.
4. build a React component `StandTagPicker` to wrap over Stand's `TagAutocomplete` and `TagTable` with a dark colour theme, and make it work with the `ManagedField` component.
5. add the `StandTagPicker` to the `VideoData` panel in furniture tab behind the feature tag.
6. make atom tag field mandatory for default videos (i.e. Non-Youtube videos).

To test the change, we should manually append the query parameter `?supportTagging` to the end of the URL.
1. Create a new Non-Youtube video. On furniture tab, we should see the field "Tags" between "Commissioning Desks" and "Composer Keywords". On the publish bar, the "Publish" button is disabled with a label "Add at least one tag before publishing non-youtube video".
2. Click "Edit". We should see the tag picker field under "Tags" with a yellow label "This field is required for publishing default videos". As we input text, the tag picker should give a list of tags that match the text. It supports wildcard at the start of the search word.
3. Select tags from the tag picker.  We should see the selected tags in the table below the tag picker. We could reorder the tags or delete any of them.
4. Click "Save". We should see the selected tag paths in the data model returned from the backend API (URL: https://video.local.dev-gutools.co.uk/api/atoms/<atom ID>). The Publish button should be enabled now.
5. Open the same video in MAM on another tab. We should see the tags field populated with the selected tags correctly.
6. Create a new Loop (or Cinemagraph or Youtube) video. We should see the Tags field. The Publish button is enabled.
7. Click "Edit". We should see the tag picker field but without the yellow label ("This field is required for publishing default videos").

To make sure the current behaviour is not affected when the feature flag is off, open MAM without the query parameter `supportTagging`.
1. The Tags field is not visible whether in preview or edit mode.
2. The publish button is enabled for "Non-Youtube" videos with or without tags.

Atoms can be tagged, which makes it possible to include tags in data lake and improve data analysis (future work).

The UI change and the validation are behind a query parameter. 

This PR should not bring any user-facing changes. The risk should be low.

## Images

| Screen |
| --- | 
| Preview Mode |
| Screenshot 2026-05-11 at 12 48 39 |
| Edit Mode - tag picker |
| Screenshot 2026-05-11 at 15 00 10|
| Show selected tags |
| Screenshot 2026-05-11 at 13 16 06 |
| Disabled published button |
| Screenshot 2026-05-11 at 15 00 22 |

## Accessibility

---
  - https://app.asana.com/0/0/1214430137603646


package.json
```diff
@@ -73,6 +73,8 @@
     "@dnd-kit/core": "^6.0.7",
     "@dnd-kit/sortable": "^7.0.2",
     "@dnd-kit/utilities": "^3.2.1",
+		"@emotion/react": "^11.11.4",
+    "@guardian/stand": "^0.0.31",
```
SHession, line 77: 🎉

common/src/main/scala/com/gu/media/model/MediaAtom.scala
```diff
@@ -50,6 +50,7 @@ abstract class MediaAtomBase {
   val tags: List[String]
   val byline: List[String]
   val commissioningDesks: List[String]
+  val atomTagIds: List[String]
```
SHession: Small thing but this attribute is being added under the header `composer metadata`. it may make sense higher up, under `generic metadata`.

As previously mentioned, `atomTagIds` is not an ideal name, and I think easily confused with `tags` or `keywords`.  It appears `keywords` are for tags to be applied to the canonical Composer page, whilst `tags` are for YouTube. Ideally these names would be better, but at least we could add some comments to the model to make it more obvious for future engineers.

public/video-ui/src/constants/requiredForDefaultVideo.ts
```diff
@@ -0,0 +1,13 @@
+export default class RequiredForDefaultVideo {
+  static get warning() {
+    return 'This field is required for publishing default videos';
```
SHession: For a warning only we might want to avoid the term required.

public/video-ui/src/util/validateField.ts
SHession, line 1: I expect it be trivial to use co-pilot to rewrite this file as TypeScript. Might be worth doing to help improve our TS coverage, even slightly.

public/video-ui/src/util/videoPublishCheck.ts
SHession, line 1: `nit` I think this file should be located in the util directory and have a lowercase V for the title.

public/video-ui/src/components/ManagedForm/ManagedField.jsx
```diff
@@ -73,13 +76,14 @@ export class ManagedField extends React.Component {
     }
 
     if (this.props.updateWarnings) {
-
-      if (notification && notification.type === FieldNotification.warning) {
+      // despite its general function name "updateWarnings", this function "updateWarnings" updates
+      // the slice "formFieldWarnings" in the store, which is used in the "Header" component to decide
+      // whether a composer page can be created. It assumes that all warnings in the slice come from
+      // composer related fields.
+      if (notification && notification.type === FieldNotification.warning && composerValidation) {
```
SHession, line 83: Useful comment 🙏

public/video-ui/src/components/VideoData/VideoData.jsx
```diff
@@ -145,6 +148,19 @@ export default class VideoData extends React.Component {
...
             </ManagedField>
         }
+        {
+          isTaggingSupported() &&
+          <ManagedField
+            fieldLocation="atomTagIds"
+            name="Tags"
+            formRowClass="form__row__byline"
+            isDesired={mustHaveTags}
+            isRequired={false}
```
SHession, line 158: This doesn't look like what I was expecting it to be required for default video and optional for all others?

> A) Can have the ‘tag’ option as mandatory for default videos in MAM? Obviously we’d need this behind a feature flag until ready to release so we don’t impact anything in the current workflow. 
> 
> B) How much effort would it be to add it as optional field for Loops and Cinemagraphs?

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.jsx
SHession, line 1: This component also looks like it would easily written in TypeScript, using modern React style. However, happy for that to be considered separate.

public/video-ui/src/services/tagmanager.ts
```diff
@@ -1,10 +1,39 @@
...
   return apiRequest({
     url: `${tagManagerUrl}/hyper/tags?query=${query}&limit=150&type=${types.join(",")}${subTypeQuery}`
-  });
+  }) as Promise<TagManagerApiResponse>;
 }
 
+export function getTagByPath(tagManagerUrl: string, path: string): Promise<Tag | undefined> {
+  const apiResponse = apiRequest({
+    url: `${tagManagerUrl}/hyper/tags?query=${path}&limit=150&searchField=path`
+  }) as Promise<TagManagerApiResponse>;
```
SHession: We can avoid these two type assertions by passing the type parameter to the `apiRequest` call.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -0,0 +1,323 @@
...
+    type: data.type,
+    sectionName: data.section.name
+  };
+};
+
+const videoTagsToStringList = (tags: VideoTag[]) => {
+  return tags.map(tag => tag.path);
+};
+
+let pseudoIdCounter = -100; // For generating unique IDs for tags that don't exist in the tag manager
```
SHession, line 33: We are effectively using `path` as ID  (which has made me realise TagIds is a bit confusing in of itself). Do we need pseudo IDs? Do we care about IDs at all when we have paths?

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -0,0 +1,323 @@
...
+            : ''}
+        </>
+      )}
+      {!editable && renderReadOnly()}
+    </>
+  );
+ };
+
+//REDUX CONNECTIONS
+import { connect } from 'react-redux';
```
SHession: This can be avoided when hooks are available: https://redux.js.org/usage/migrating-to-modern-redux#migrating-connect-to-hooks

SHession (commented):
I've reviewed but not yet tested the functionally. I have submitted the comments to unblock any changes. 

It is unfortunate how muddied our naming is around tags and Ids, but it seems hard to avoid at this point.

common/src/main/scala/com/gu/media/model/MediaAtom.scala
waisingyiu: Great idea! I've moved the attribute and added comments for clarity.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: Great idea!

public/video-ui/src/services/tagmanager.ts
waisingyiu: That's great!

public/video-ui/src/util/videoPublishCheck.ts
waisingyiu: It makes sense to me!

public/video-ui/src/util/validateField.ts
waisingyiu: Sounds good, I've changed it to Typescript.

public/video-ui/src/components/VideoData/VideoData.jsx
waisingyiu: So we want to make it a required field for default video. I think the exact behaviour we may want is that there must be at least a tag when we hit the "publish" button to publish it, but it should be fine to allow users to save it with empty tags while they are editing the furniture halfway through.

If we set `isRequired` attribute to true, it means the furniture data is not valid without tags and does not allow users to save what they have got even if the atom is still in "draft". That's why 'isRequired` is always set to false.

The 'isDesired' attribute controls whether to display the "_warning_" of the field. Currently it is used, for example, to display the label "This field is required for creating composer pages" under trail text field. So we set it to be true when `mustHaveTags` returns true so that we display it only for default videos.

To make it mandatory, I've added a function `checkVideoReadyToPublish` which is called to decide whether to enable the publish button. In this function, we make sure there is at least a tag for default videos.

We don't have a mechanism for feature flags in MAM so in this PR, I make use of a query parameter to turn on or off this feature. By default, we don't show the tag field or require tags on default videos. When there is the query parameter `supportTagging` on the URL, the page shows the tag field on all types of video formation and make it mandatory for default videos.

Please let me know if you have any thoughts.

public/video-ui/src/constants/requiredForDefaultVideo.ts
waisingyiu: Despite its name "warning", this message is shown when the `isDesired` property is set on the field. As mentioned in another [reply](https://github.com/guardian/media-atom-maker/pull/1548#discussion_r3235057668), I follow the same pattern as we are using for trail text field.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: We need a number property for ID mainly to meet the requirement of Stand's `TagAutocomplete` and `TagTable` components:

https://github.com/guardian/stand/blob/cd18a014ea1afa11fab9603f6d153f42cb4ac5e0/src/components/tag-picker/types.ts#L2 

Alternatively we may generate a numerical value from a hash of the path but we need to define one for ourselves unless we import from a (not-commonly-used) third party library.

Great if you have other ideas!

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.jsx
waisingyiu: This pull request is relatively big so it may be more manageable if the change is in a separate PR.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+  return (
+    <>
+      <div className="form__row">
+        <div className="form__label__layout">
+          <label className="form__label">{fieldName}</label>
+        </div>
+      </div>
+      {editable && (
+        <>
```
SHession, line 265: It would be valuable to have a bit of explanation text to explain what the field is for. We should discuss some sensible wording with stakeholders. This can follow-on separately from this PR.

public/video-ui/src/components/VideoData/VideoData.jsx
SHession: Thanks for the explanation that does make sense. 

Thinking on it, blocking publication without tags does seem fairly severe, but it may be the most practical way to ensure atoms are being tagged.

public/video-ui/src/constants/requiredForDefaultVideo.ts
SHession: In the publish bar, we refer to this video type as "non-youtube". We should also do the same here to ensure a consistent UX.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
SHession: That makes sense, thanks for explaining. Happy to keep it as is. 

Is there any specific reason we start at -100?

public/video-ui/src/components/VideoData/VideoData.jsx
```diff
@@ -145,6 +148,19 @@ export default class VideoData extends React.Component {
...
+            inputPlaceholder="Search tags (type '*' to show all)"
+          >
+            <StandTagPicker tagTypes={[TagTypes.keyword]} />
```
SHession, line 161: It would be nice to follow up on this, setting all MAM tag pickers to use the stand version.

image

SHession (commented):
Thank you for the detailed comments. The explanations make sense and the functionality appears to be working nicely. I have added a few more comments, all non-blocking, except the "default" / "non-youtube" discrepancy.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: No, not really. It should work as long as it won't clash with real tag IDs from tagmanager.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: A good point.

public/video-ui/src/constants/requiredForDefaultVideo.ts
waisingyiu: A great spot!

SHession (approved):

PR #1560, opened 2026-05-19
More primitive aspect ratio calculation

davidfurey:
## What does this change?

The existing aspect ratio solution is quite complicated. It attempts to express the aspect ratio in the standard form x:y, allowing a supplied pixel tolerance to match to pre-defined ratios.

For our purposes, it is more helpful to assume that the video creator was trying to create a video that matched one of our ratios, but with the constraint that the width (or height) had to be fixed to a specific size. This more closely matches how our tool is used.

A side effect is that in most cases a tolerance of zero should be fine.

I believe this will make https://github.com/guardian/media-atom-maker/pull/1558 unnecessary

### DIFF
--- a/common/src/test/scala/com/gu/media/util/AspectRatioTest.scala
+++ b/common/src/test/scala/com/gu/media/util/AspectRatioTest.scala
@@ -7,25 +7,36 @@ class AspectRatioTest extends AnyFlatSpec with Matchers {
 
   "calculate" should "produce an aspect ratio from dimensions" in {
     AspectRatio.calculate(1280, 720).map(_.name) should contain("16:9")
+    AspectRatio.calculate(1280, 720).map(_.name) should contain("16:9")
     AspectRatio.calculate(720, 1280).map(_.name) should contain("9:16")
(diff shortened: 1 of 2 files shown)


common/src/test/scala/com/gu/media/util/AspectRatioTest.scala
```diff
@@ -6,22 +6,27 @@ import org.scalatest.matchers.should.Matchers
 class AspectRatioTest extends AnyFlatSpec with Matchers {
 
   "calculate" should "produce an aspect ratio from dimensions" in {
+    AspectRatio.calculate(1280, 720).map(_.name) should contain("16:9")
```
abeddow91: This line looks to be a duplicate - could it be removed?

abeddow91 (approved):

common/src/test/scala/com/gu/media/util/AspectRatioTest.scala
davidfurey: Oops. Removed

PR #1561, opened 2026-05-19
Extend tag picker to video tracking tags and other allowed tags

waisingyiu:
## What does this change?

This pull request follows up our work on supporting tags on atoms in #1548, extending the tag picker to allow more types of tags to apply to video atoms.

It allows the following tags to be applied to video atoms:
- all keyword tags
- all tracking/video tags
- two tone tags: `tone/news` and `tone/features`
- one content type tag: `podcasts`

Specifically it makes the following changes:
- query tag manager API with "keyword", "tracking", "tone" and "contenttype" types by default 
- add a filter function to refine the search result
- add a dropdown list in the tag picker to allow searching in one type only 

To test the change, we should manually append the query parameter ?supportTagging to the end of the URL.

1. Create a new Non-Youtube video. 
2. Click "Edit" on the Furniture tab. We should see the tag picker field under "Tags" and it has a dropdown box to select a particular tag type for searching. It should have the options "keyword", "tracking-video", "tone" and "contenttype".
3. Select tags from the tag picker. We could limit the type of tags to search by the tag picker. We should see the selected tags with their type in the table below the tag picker.
4. Click "Save". We should see the selected tag paths in the data model returned from the backend API (URL: https://video.local.dev-gutools.co.uk/api/atoms/<atom id>).
5. Open the video in a separate tab. We should see the selected tags correctly.

It makes progress with improving data analysis on video content.

The tags field is currently behind a query parameter.

The UI change of this PR should not be visible to users. The risk should be low.

## Images

| Screenshots |
| --- |
| Tag picker with no selected tags |
| Screenshot 2026-05-20 at 14 36 57 |
| Filter by tag type |
| Screenshot 2026-05-20 at 14 37 07 |
| With selected tags |
| Screenshot 2026-05-20 at 12 28 07 |
| In non-edit mode |
| Screenshot 2026-05-20 at 14 37 26 |

## Accessibility

-   [X] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)

---
  - https://app.asana.com/0/0/1214904384250550

### DIFF
--- a/public/video-ui/src/components/FormFields/StandTagPicker.tsx
+++ b/public/video-ui/src/components/FormFields/StandTagPicker.tsx
@@ -176,7 +176,15 @@ const StandTagPicker = ({ tagTypes, allowTags, filters, fieldName, fieldValue, e
   const [isLoading, setIsLoading] = useState(false);
   const [tagSearchError, setTagSearchError] = useState<boolean>(false);
   const tagManagerUrl = useSelector((state: {config: AppConfig}) => state.config.tagManagerUrl);
-
+  const tagPickerContainerStyle = { display: 'flex', alignItems: 'stretch', gap: '8px', marginBottom: '8px' };
+  const tagPickerDropdownStyle = {
+    border: "1px solid #BDBDBD",
+    backgroundColor: "#393939",
+    color: "inherit",
+    font: "inherit",
+    paddingLeft: "8px",
+    paddingRight: "8px"
+  };
   const searchTags = useCallback((inputText: string, selectedTags: VideoTag[], filter?: StandTagPickerFilter) => {
     if (tagManagerUrl) {
       setIsLoading(true);
@@ -295,7 +303,7 @@ const StandTagPicker = ({ tagTypes, allowTags, filters, fieldName, fieldValue, e
               Tags are currently unavailable
             </div>
           )}
-          <div style={{ display: 'flex', alignItems: 'flex-start', gap: '8px' }}>
+          <div style={tagPickerContainerStyle}>
             <div style={{ flex: 1 }}>
               <TagAutocomplete
                 onTextInputChange={onTextInputChange}
@@ -310,7 +318,7 @@ const StandTagPicker = ({ tagTypes, allowTags, filters, fieldName, fieldValue, e
               />
             </div>
             {filters && filters.length > 0 && (
-              <select onChange={onFilterChange}>
+              <select style={tagPickerDropdownStyle} onChange={onFilterChange}>
                 {filters.map((filter, index) => (
                   <option key={filter.displayLabel} value={index}>{filter.displayLabel}</option>
                 ))}
--- a/public/video-ui/src/components/FormFields/StandTagPicker.tsx
+++ b/public/video-ui/src/components/FormFields/StandTagPicker.tsx
@@ -8,11 +8,18 @@ import FieldNotification from '../../constants/FieldNotification';
 import debounce from "lodash/debounce";
 import type { AppConfig } from '../../slices/config';
 
-type VideoTag = {
+export type StandTagPickerFilter = {
+  displayLabel: string;
+  tagTypes: string[];
+  tagSubType?: string;
+}
+
+export type VideoTag = {
   id: number;
   path: string;
   name: string;
   type: string;
+  subtype?: string;
   sectionName: string;
 };
 
@@ -22,7 +29,8 @@ const videoTagFromTagManager = (data: Tag): VideoTag => {
     path: data.path,
     name: data.internalName,
     type: data.type,
-    sectionName: data.section.name
+    sectionName: data.section.name,
+    subtype: data.subType
   };
 };
 
@@ -130,6 +138,8 @@ const editableUiTheme = {
 
 interface StandTagPickerProps {
   tagTypes: string[];
+  allowTags?: (tag: VideoTag) => boolean;
+  filters?: StandTagPickerFilter[];
 
   // the following properties are passed down from the parent ManagedField
   fieldName: string;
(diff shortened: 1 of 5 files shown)


public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -157,25 +167,39 @@ const isFieldValueChanged = (fieldValue: string[], selectedTags: VideoTag[]) =>
...
   const [options, setOptions] = useState<VideoTag[]>([]);
   const [value, setValue] = useState('');
+  const [selectedFilter, setSelectedFilter] = useState<StandTagPickerFilter | undefined>(filters?.[0]);
   const [isLoading, setIsLoading] = useState(false);
   const [tagSearchError, setTagSearchError] = useState<boolean>(false);
   const tagManagerUrl = useSelector((state: {config: AppConfig}) => state.config.tagManagerUrl);
-
-  const searchTags = useCallback((inputText: string, selectedTags: VideoTag[]) => {
+  const tagPickerContainerStyle = { display: 'flex', alignItems: 'stretch', gap: '8px', marginBottom: '8px' };
+  const tagPickerDropdownStyle = {
```
SHession: I would tend to declare styles outside of the component unless there is a specific reason to declare them within the function?

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -130,6 +138,8 @@ const editableUiTheme = {
 
 interface StandTagPickerProps {
   tagTypes: string[];
+  allowTags?: (tag: VideoTag) => boolean;
+  filters?: StandTagPickerFilter[];
```
SHession: `nit` For me, `filterOptions` would be a more intuitive name for this prop.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -196,7 +220,23 @@ const StandTagPicker = ({ tagTypes, fieldName, fieldValue, editable, onUpdateFie
...
+      }
+      debouncedSearchTags(inputText, selectedTags, selectedFilter);
+    }
+  };
+
+  const onFilterChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    const newFilter = filters?.[Number(e.target.value)];
+    setSelectedFilter(newFilter);
+    if (value !== '') {
+      debouncedSearchTags(value, selectedTags, newFilter);
```
SHession: Do we want to to `setIsLoading` in this case?

public/video-ui/src/components/FormFields/tags/allowTags.ts
```diff
@@ -0,0 +1,37 @@
...
+    tagSubType: 'video'
+  },
+  {
+    displayLabel: 'Tone',
+    tagTypes: [TagTypes.tone]
+  },
+  {
+    displayLabel: 'Content type',
+    tagTypes: [TagTypes.contentType]
+  }
```
SHession, line 36: I don't have a solution for this but given we only allow two tones and one content type, the UX seems a little strange to me. I.e. I select Content Type, the only option I will be able to find is `podcast`.

SHession (commented):
A couple of minor suggestions. This is looking good!

public/video-ui/src/components/FormFields/StandTagPicker.tsx
```diff
@@ -130,6 +138,8 @@ const editableUiTheme = {
 
 interface StandTagPickerProps {
   tagTypes: string[];
```
SHession, line 152: `nit` it seems to me `tagTypes` is essentially unused at the moment and could be removed the existing picker without affecting the functionality. We may be expecting to use this for other parts of the UI, but essentially `tagTypes` and `filters` are mutually exclusive.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: No specific reasons really. Moving it out of the component sounds good to me - they are constants and we don't really need to create these constants whenever the rendering happens.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: Good spot!

public/video-ui/src/components/FormFields/tags/allowTags.ts
waisingyiu: Yes that's true. But unless we don't provide any filtering function, it may be good to use consistent UI in tag pickers, which means filtering by tag type. It may seem a little strange, but it should not cause any inconvenience to how they use the tag picker, given that they are internal users.

Hopefully I will run it through in a team demo session to get what they think about it.

public/video-ui/src/components/FormFields/StandTagPicker.tsx
waisingyiu: Yes, you are right - essentially it uses either `tagTypes` or `filters` but not both. The 'tagTypes' is a required prop to make sure that at least one of them is supplied, but if `filters` is provided, the `tagTypes` is essentially not used.

I hope this `StandTagPicker` can be more generic and it can be used on some other fields like "Commissioning Desks" and "Composer Keywords". It appears that we may not need to have the filter dropdown list for commissioning desks. In this case, we may want to use `tagTypes` instead of `filters`.

public/video-ui/src/components/VideoData/VideoData.jsx
```diff
@@ -162,7 +162,7 @@ export default class VideoData extends React.Component {
             <StandTagPicker
               tagTypes={[TagTypes.keyword, TagTypes.tracking, TagTypes.tone]}
```
SHession: > The 'tagTypes' is a required prop to make sure that at least one of them is supplied, but if filters is provided, the tagTypes is essentially not used.

It would be worth setting this to an empty array to avoid confusion.

SHession (approved):
Tested locally and reviewed the latest changes. LGTM!

public/video-ui/src/components/VideoData/VideoData.jsx
waisingyiu: sounds good!

PR #1562, opened 2026-05-20
Narrower typing for config, parsed on ingestion

bryophyta:
## What does this change?

The `AppConfig` type in the client code was typed more loosely than it needed to be.

Most of the fields are injected into the page HTML by the server, and on the server the type treats almost all of the fields as mandatory:

https://github.com/guardian/media-atom-maker/blob/f5056ce7d6d27030608261e5db895f001b7d2af1/app/model/ClientConfig.scala#L21-L40

The client was treating them as optional, but it feels reasonable to say that if any of the fields are missing then something must have gone wrong. So let's assert that all of the fields are present when the app's first loaded. This way, it should fail early if anything's actually missing, and then when we use the values in the codebase we can be confident of what's available.

To implement this, I've created a zod schema for the ClientConfig and used it to parse the data from the HTML template before we load it into config.

- [x] Deploy to CODE and verify that the app loads and works as expected.
- [x] Check that config-dependent functionality (e.g. Sentry integration) still works.

## Accessibility


public/video-ui/src/components/FormFields/TagPicker.test.jsx
```diff
@@ -27,11 +27,6 @@ const defaultProps = {
 };
 
 const store = setupStore();
-store.dispatch(
```
bryophyta, line 30: as far as I can tell, this is no longer needed because the config slice now initialises with dummy config if running in test.

public/video-ui/src/util/config.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export function extractEmbeddedModeFromUrl(): boolean {
+  const maybeEmbeddedMode = new URLSearchParams(location.search).get(
+    'embeddedMode'
+  );
+  if (!maybeEmbeddedMode) {
+    return false;
+  }
+  return maybeEmbeddedMode === 'true';
```
andrew-nowak: Composer appears to embed MAM with `embeddedMode=live`, I can see a couple of other modes but I'm not sure if they're similarly used (https://github.com/search?q=org%3Aguardian%20embeddedmode&type=code)
As is, this branch doesn't appear to allow embedding videos into articles (as noted in a comment in chat which I now can't find...)

bryophyta: ooh, good call!

I think the updated logic should work for this case now? cd568a8b

andrew-nowak: That commit looks good to me, but it seems to have gone onto a different branch than this PR?

bryophyta: Sorry, I'm trying to do too many things at once today! I've pushed it to this branch now 👍

andrew-nowak (approved):

PR #1563, opened 2026-05-20
try increasing youtube uploader memory

andrew-nowak:
## What does this change?

Try the straightforward thing to try speeding up the lambdas - increase the memory of the yt upload lambda. Going 2GB to 8GB sped up the upload of a 3.5GB vid from 8:30 to 5:55. A large increase for a comparatively small speedup, but every little helps.

Ran on CODE, an unscientific test showed it reduce the overall upload time for a 3.5GB vid from 8m30s to 5m55s. This may or may not hold (a subsequent test which increased the memory further then took 7m30s to run, which is not what we'd have expected), but feels a fairly safe way to increase upload speed a bit.

## Accessibility

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -216,7 +216,8 @@ lazy val uploader = (project in file("uploader"))
           "Checks to see if a chunk of video has been uploaded to S3"
       ),
       "UploadChunkToYouTube" -> LambdaConfig(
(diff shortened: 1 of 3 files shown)


lindseydew (approved):
Noice 💯

PR #1566, opened 2026-05-21
Call the grid api for crop options and store in memory

lindseydew:
## What does this change?

This calls a grid API endpoint to find out what crop options are supported and uses these to configure the grid crop endpoint

- [x] Update CODE config
- [x] Update PROD config

## Accessibility


app/schedule/GridAPI.scala
```diff
@@ -0,0 +1,59 @@
...
+    actorSystem: ActorSystem,
+    ws: WSClient,
+    gridApiUrl: String
+)(implicit
+    ec: ExecutionContext
+) extends Logging {
+
+  actorSystem.scheduler.scheduleWithFixedDelay(
+    initialDelay = 0.seconds,
+    delay = 24.hours
```
andrew-nowak: I'm not certain we'd need this cache, I think we'd be fine if we just proxied all requests directly to the Grid endpoint (it's just returning a constant response, so should scale quite well).
If we did keep the cache, we should make it handle errors a bit better - if the first fetch fails, it'll hold an empty list for 24 hours until the scheduler runs again. An easy way might be to reduce the delay to something like 5 mins, or we could retry `runJob` a few times in the failure case?

andrew-nowak (approved):
LGTM! I think worth thinking a little bit more about the caching behaviour, but all is working well for me on CODE!

app/schedule/GridAPI.scala
lindseydew: Yeah, that's a good point. I guess my feeling was the fallback behaviour is still acceptable in case of a failure because without the crop options MAM will still create a functional url to GRID. 

But agree it is probably a bit of hidden fail which might be confusing. I think reducing to 5 minutes sounds like a sensible compromise 👍

PR #1567, opened 2026-05-26
Create JobMetricsPublisher Lambda 

lindseydew:
## What does this change?
This adds a `JobMetricsPublisher` lambda, which computes a set of metrics from the job and posts to the telemetry server. 

## Accessibility


common/src/main/scala/com/gu/media/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+  }
+  implicit val tagMapValueWrite: Writes[Map[String, TagValue]] =
+    new Writes[Map[String, TagValue]] {
+
+      override def writes(tagMap: Map[String, TagValue]): JsValue = {
+        JsObject(tagMap.map({ case (k, v) => k -> Json.toJson(v) }))
+      }
+    }
```
andrew-nowak: nit: So long as you have the above teaching play-json how to write `TagValue`s  you shouldn't need to also teach it about `Map[String, TagValue]`, it should be able to do that by itself

```suggestion
```

common/src/main/scala/com/gu/media/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,158 @@
...
+    .builder()
+    .credentialsProvider(
+      AwsCredentialsProviderChain
+        .builder()
+        .credentialsProviders(
+          DefaultCredentialsProvider.builder().build()
+        )
+        .build()
+    )
+    .build()
```
andrew-nowak, line 52: nit: Since this is setting the credentials provider to be the default, we can create the client directly

```suggestion
  val secretsManagerClient = SecretsManagerClient.create()
```

common/src/main/scala/com/gu/media/util/JobCompletedMetrics.scala
```diff
@@ -0,0 +1,175 @@
...
+  DescribeExecutionResponse,
+  GetExecutionHistoryRequest,
+  HistoryEvent,
+  HistoryEventType
+}
+
+import scala.jdk.CollectionConverters.IterableHasAsScala
+
+class JobCompletedMetrics(stepFunctionsClient: SfnClient) {
+  val LAMBA_WARM_UP = "LAMBA_WARM_UP"
```
andrew-nowak: nit:

```suggestion
  val LAMBDA_WARM_UP = "LAMBDA_WARM_UP"
```

common/src/main/scala/com/gu/media/util/JobCompletedMetrics.scala
```diff
@@ -0,0 +1,175 @@
...
+          (enteredX, exitedX) <- enteredEvents.zip(exitedEvents)
+          if enteredX.id() < exitedX.id()
+        } yield {
+          exitedX.timestamp().toEpochMilli - enteredX.timestamp().toEpochMilli
+        }
+
+        val name = enteredEvents.headOption
+          .flatMap(h =>
+            h.`type`() match {
+              case HistoryEventType.LAMBDA_FUNCTION_STARTED =>
```
bryophyta: I've created a test case which seems to show a problem with the logic for calculating the lambda warm up time: https://github.com/guardian/media-atom-maker/pull/1570

If I've understood the intended functionality correctly, then should this line be `LAMBDA_FUNCTION_SCHEDULED` instead of `LAMBDA_FUNCTION_STARTED`?

Alternatively, it looks like we could use the `id` that's in scope here (the key of the `taskEnteredMap`) as the `name`. If it is indeed equivalent then that might be simpler than re-calculating the name here?

common/src/main/scala/com/gu/media/util/JobCompletedMetrics.scala
```diff
@@ -0,0 +1,175 @@
...
+  def getMetricsForJobRun(executionArn: String) = {
+    val describeExecution = getExecutionResponse(executionArn)
+    val runTime = getRunTime(describeExecution)
+    val stepId = extractStepId(executionArn)
+
+    val historyEvents = getAllHistoryEvents(executionArn)
+    val metricsFromHistory =
+      getMetricsFromHistory(historyEvents, runTime)
+
+    val uploadOpt = Json.parse(describeExecution.input()).validate[Upload].asOpt
```
bryophyta: I think that this `input` value is passed directly to the handler function as part of the step function orchestration, which might save us a network call? 

I don't think it's a big deal though

common/src/main/scala/com/gu/media/util/JobCompletedMetrics.scala
lindseydew: Oh yeah good spot thanks 👍

common/src/main/scala/com/gu/media/util/MediaAtomHelpers.scala
```diff
@@ -31,6 +31,12 @@ object MediaAtomHelpers {
     )
   }
 
+  def extractAtomIdAndVersion(stepFunctionId: String): (String, String) = {
+    val atomId = stepFunctionId.take(stepFunctionId.length - 2)
+    val version = stepFunctionId.drop(stepFunctionId.length - 1)
```
andrew-nowak: These will break if we get to a version number with more than 1 digit

common/src/main/scala/com/gu/media/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,158 @@
...
+}
+
+class HMACClient(secretArn: String) extends HMACHeaders {
+  lazy val secret =
+    SecretsManager.getSecret(secretArn) getOrElse (throw new Exception(
+      s"Could not retrieve $secretArn from secrets manager"
+    ))
+
+  override def createHMACHeaderValues(uri: URI): HMACHeaderValues =
+    super.createHMACHeaderValues(uri)
```
andrew-nowak: nit: This override is only calling the overridden function, so I think we can skip the override entirely?


```suggestion
```

common/src/main/scala/com/gu/media/util/JobCompletedMetrics.scala
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  def getMetricsFromHistory(
+      historyEvents: List[HistoryEvent],
+      runTime: Long
+  ): Map[String, TagValue] = {
+    val durationsMap = computeDurations(historyEvents)
+    val totalSubTimes = durationsMap.foldLeft(0L)({ case (acc, (_, v)) =>
+      acc + v
+    })
```
andrew-nowak: nit:

```suggestion
    val totalSubTimes = durationsMap.values.sum
```

andrew-nowak (approved):
lgtm, I have a few comments which are mostly optional nits, but also one bigger point about parsing out the version number

common/src/main/scala/com/gu/media/util/MediaAtomHelpers.scala
lindseydew: Oh gosh, good spot thank you!

PR #1568, opened 2026-06-01
try upgrading atom-maker lib to new v12 with updated reindexing logic

andrew-nowak:
## What does this change?

Updates atom-workshop to v12. Implement the updated atom reindexing classes (https://github.com/guardian/atom-maker/pull/146) which reindex atoms in batches, so we no longer need to fit the entire atom database into memory all at once.

Tested on CODE 

With a successful reindex of atoms.

### DIFF
--- a/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
+++ b/common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
@@ -1,6 +1,5 @@
 package com.gu.media.model
 
-import com.amazonaws.services.s3.model.PutObjectRequest
 import com.gu.media.aws.{AwsAccess, UploadAccess}
 import com.gu.media.upload.CompleteUploadKey
 import com.gu.ai.x.play.json.Jsonx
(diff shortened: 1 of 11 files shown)


common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
```diff
@@ -47,29 +46,6 @@ object AtomAssignedProjectMessage {
...
-  def build(atom: MediaAtom, putRequest: PutObjectRequest): PacFileMessage = {
-    PacFileMessage(
-      "pac-file-upload",
-      atom.id,
-      putRequest.getBucketName,
-      putRequest.getKey
-    )
-  }
-}
-
```
JustinPinner, line 72: I'm not familiar with these operations but it appears we just don't do this anymore and we're sweeping out old code now.

JustinPinner (approved):
👍  One Q re. removal of `PacFileMessage` but if that's just a tidy up this looks good to go.

common/src/main/scala/com/gu/media/model/PlutoIntegrationData.scala
andrew-nowak: Yep exactly - upgrading the atom-maker lib removed the final transitive dependencies on the AWS SDK v1, so flagged the `PutObjectRequest` import here. Rather than migrate it to the v2 equivalent, I removed it as it was marked as unused

PR #1573, opened 2026-06-04
Track YouTube uploads

lindseydew:
## What does this change?
This adds a telemetry log to track when videos are uploaded using the tool and when they are uploaded to youtube directly and added to the tool later

- [x] Upload latest DEV config to s3
- [x] Add secrets config to CODE creds
- [x] Add secrets config to PROD creds

## Accessibility


app/model/commands/AddAssetCommand.scala
```diff
@@ -47,7 +49,15 @@ case class AddAssetCommand(
         AssetAlreadyAdded
 
       case YouTubeId(videoId) =>
-        addAsset(atom, mediaAtom, currentAssets, videoId)
+        val asset = addAsset(atom, mediaAtom, currentAssets, videoId)
+        telemetry.sendTelemetryEvent(
+          "VIDEO_UPLOADED_VIA_YOUTUBE",
+          Map(
+            "atomId" -> TagString(atomId),
+            "duration" -> TagLong(asset.duration.getOrElse(0L))
```
andrew-nowak: I think it'll be worth distinguishing at a glance if the imported video was a livestream, since there's no other legitimate way of importing those so we don't need to consider them as trying to work around the upload system.

I think the simplest way of doing that is something like this

```suggestion
            "duration" -> TagLong(asset.duration.getOrElse(0L)),
            "isLivestream" -> videoUri.contains("/live/"),
```

but I don't know if it's guaranteed that the url will always have a live component. The safest then is probably to call the youtube API (like in `getYoutubeChannel` and check the `snippet.liveBroadcastContent` field https://developers.google.com/youtube/v3/docs/videos#snippet.liveBroadcastContent

andrew-nowak (approved):

app/model/commands/AddAssetCommand.scala
lindseydew: Ok, I think I'll go for the simplest thing for now, and we can always tweak the metrics if we need more robustness 👍

PR #1575, opened 2026-06-08
Add multiple resolutions to HLS streaming video output

waisingyiu:
## What does this change?

Now that we decided to use HLS streaming videos for long form videos, we wanted to improve user experience by adding more resolutions to the HLS output so that the video player could switch to lower bitrate video if the network bandwidth is low according to the HLS specification.  

Currently we transcode the uploaded video to HLS with the output dimension set to 720px in height. This pull request adds a new output with 480px in width, which makes HLS in line with the two resolutions of MP4 video outputs.

I've deployed the branch to `CODE`.
1. On MAM, create a Non-Youtube video and upload a video.
2. Fill in required fields, create a video page and then publish.
3. Preview the video page and check the content of those m3u8 files via the Inspector tool.
4. The root m3u8 should have two resolutions of the videos (in another m3u8 file). They should show the resolutions of (1) 720px in height and (2) 480px in width.
5. The child m3u8 files should list the ts segment videos.
6. We should be able to play the video in the preview.

Improve the user experience and it may potentially lead to more people viewing the video to completion.

It may increase the traffic to our S3 bucket for m3u8's videos but the impact should not be concerning because most of the cost actually goes to Fastly serving videos to clients.

## Images

| Screenshots |
| --- |
| Root m3u8 |
| Screenshot 2026-06-09 at 09 22 39 |
| m3u8 - 720px in height |
| Screenshot 2026-06-09 at 09 23 21 |
| m3u8 - 480px in width |
| Screenshot 2026-06-09 at 09 23 14 |
| Output from MediaConvert |
| Screenshot 2026-06-09 at 09 28 02 |

## Accessibility


uploader/src/main/scala/com/gu/media/upload/mediaconvert/Resolution.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.media.upload.mediaconvert
+
+case class Dimensions(width: Option[Int], height: Option[Int])
```
waisingyiu, line 3: I extracted all these classes from "MP4Output.scala" as they are shared between MP4Output and VideoOutput after this PR.

davidfurey (approved):

PR #1578, opened 2026-06-10
Require users to have permission to publish default non-youtube videos

davidfurey:
## What does this change?

The non-youtube default player should only be used in limited circumstances when Youtube is not suitable. While we work out why the non-youtube player is being used more than we expect, we will limit the number of people who can use this feature by putting it behind a permission.

Instead of just hiding the option, the option is still displayed but with a note saying to contact Central Production if you need to use this feature. CP will be able to assess the need and if appropriate give the user the necessary permission.

Since this change is merely to discourage use of a generally safe feature, this PR does not actually enforce this limitation server side.

Tested locally

Fewer non-Youtube videos are created.

People who need to create these videos will be unable to. This is mitigated by referring them to central production.

## Images
Screenshot 2026-06-10 at 10 14 20

## Accessibility

### DIFF
--- a/public/video-ui/src/components/Create.tsx
+++ b/public/video-ui/src/components/Create.tsx
@@ -12,13 +12,19 @@ import Youtube from "../../images/youtube.svg?react";
 import Cinemagraph from "../../images/cinemagraph.svg?react";
 import NonYoutube from "../../images/nonyoutube.svg?react";
 import {fieldLengths} from "../constants/videoEditValidation";
+import { getStore } from '../util/storeAccessor';
 
 export default class Create extends React.Component {
   props: React.PropsWithChildren<{
-    createVideo: (video: VideoWithoutId) => (dispatch: AppDispatch) => Promise<void>
-    inModal: boolean
+    createVideo: (video: VideoWithoutId) => (dispatch: AppDispatch) => Promise<void>;
+    inModal: boolean;
     closeCreateModal?: () => void;
   }>;
+
+  // this permission is not validated on the server-side because we are only hiding the option
+  // to discourage use of this feature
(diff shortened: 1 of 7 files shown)


public/video-ui/src/util/config.ts
```diff
@@ -10,7 +10,8 @@ const PresenceSchema = z.object({
 const PermissionsSchema = z.object({
```
SHession, line 10: A small improvement would be to use this for typing the Create.tsx file.

public/video-ui/src/components/Create.tsx
```diff
@@ -82,9 +92,10 @@ export default class Create extends React.Component {
             </div>
             <div className="create-form__option-radio-description">
               {videoCreateOptionDetails.description}
+              { disabled ? '. Contact Central Production if you need to use this video format.' : ''}
```
SHession, line 95: As this element is faded out, this message is easy to miss.

SHession (approved):
Tested locally, this is behaving nicely. A couple of small comments.

public/video-ui/src/util/config.ts
davidfurey: I'm not entirely sure what you mean, but I've pushed a commit that might resolve this.

public/video-ui/src/components/Create.tsx
davidfurey: Fair point, although ideally the whole option is now easy to miss 🤣

public/video-ui/src/util/config.ts
SHession: Looks great!

PR #1580, opened 2026-06-11
Add additional lower quality encodings to HLS/M3U8

davidfurey:
## What does this change?

This adds 2 lower quality encodings to the HLS/M3U8 output, each at the 2 existing resolutions.

I am not anticipating that we will want to produce these 6 versions of a video on an ongoing basis. I'd like to uses this for a few videos and then review the logs to decide which ones are worth continuing to create. If a resolution/quality combination isn't requested much by users there is little point in us creating it.

About to test in CODE.

A reasonable number of requests (say 10%) are for lower quality videos, reducing our bandwidth bill.

The videos could look really bad.


waisingyiu (approved):
👍

PR #1585, opened 2026-06-11
try running all lambdas, including ffmpeg calling lambdas, under arm64

andrew-nowak:
## What does this change?

Alternative to #1584, make _all_ lambdas run on arm64. The prebuilt ffmpeg binaries are also available for the arm64 architecture, so download and use them instead of x86_64, and we can have all the lambdas running happily on arm64.

## How was this tested?

Deployed to CODE, uploaded a self-hosted loop with subtitles, and the upload completed successfully. We've not checked for any change to youtube upload speed, we can monitor ongoing metrics in PROD with realistic throughputs.


davidfurey (approved):
Lovely!

PR #1593, opened 2026-06-16
Add AV1 encoding to HLS/M3U8 output

davidfurey:
## What does this change?

This PR adds AV1 alongside H.264 in the CMAF HLS output group, and generates separate manifests for each codec so clients that don't support AV1 can fall back to the H.264-only playlist. AV1 encoding parameters are based on Mateusz's testing.

AV1 produces visually similar outputs with significantly less bandwidth.

Before we release this, we need to release https://github.com/guardian/media-atom-maker/pull/1598 and ensure that:

1. DCR lists the AV1 manifest first, and the H.264 manifest second
2. MAPI returns only the H.264 manifest 

Later we can explore how MAPI can expose the AV1 and H.264 manifests to the native app so that devices that support AV1 can use it.

#### Codec tracking on `OutputDefinition` (`model.scala`, all output objects)

- Adds a `codec: Codec` field to `OutputDefinition`, represented by a sealed trait with three wrappers: `VideoCodecWrapper`, `AudioCodecWrapper`, and `CaptionsCodecWrapper`.
- All existing output objects (`VideoOutput`, `AudioOutput`, `CaptionsOutput`, `MP4Output`, `JPEGOutput`, `WebVTTOutput`) are updated to populate this field.
- This enables filtering outputs by codec at the group level, which is needed to generate per-codec manifests.

#### AV1 encoding config and settings (`EncodingConfig.scala`, `SharedCodecSettings.scala`)

- Adds `AV1EncodingConfig(dimensions, qualityLevel)` alongside `H264EncodingConfig`, implementing `codecSettings` with `VideoCodec.AV1` and QVBR rate control.
- AV1 settings: 10-bit depth, `HIGH` adaptive quantization, spatial adaptive quantization enabled, max bitrate inherited from the existing shared bitrate config.
- The `nameModifier` derived from `EncodingConfig` now includes the codec name (e.g. `_480w_q8_av1`, `_720h_q4_h264`), making codec identifiable in logs and MediaConvert output paths.

#### Dual-codec HLS output group (`HLSOutputGroup.scala`)

- `outputDefinitions` now contains both `H264EncodingConfig` and `AV1EncodingConfig` across the same two dimensions (`480w`, `720h`) and two quality levels (`q8`, `q4`), producing 8 video outputs in total (4 H.264 + 4 AV1).
- Adds two additional CMAF manifests via `additionalManifests`:
  - `_av1`: AV1-only playlist (excludes H.264 outputs) — for clients that support AV1.
  - `_h264`: H.264-only playlist (excludes AV1 outputs) — for clients that don't support AV1.
- The main manifest continues to include all outputs which might be helpful for debugging but probably will not be used in practice.
- `manifestWithoutCodec` is a helper that builds a `CmafAdditionalManifest` by filtering outputs whose codec doesn't match the excluded codec.

#### `AddAssetToAtom.scala`

`playlistFilePaths` can now include multiple playlists, corresponding to the additional manifests explained above.

tbd

## Accessibility

### DIFF
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
@@ -3,12 +3,15 @@ package com.gu.media.upload.mediaconvert.hls
 import com.gu.media.model.VideoSource
 import com.gu.media.upload.mediaconvert.SharedCodecSettings.h264Settings
 import com.gu.media.upload.mediaconvert.{EncodingConfig, OutputDefinition}
+import com.gu.media.upload.mediaconvert.VideoCodecWrapper
+
 import software.amazon.awssdk.services.mediaconvert.model._
 
 object VideoOutput {
   def apply(config: EncodingConfig): OutputDefinition =
     OutputDefinition(
       mimeType = Some(VideoSource.mimeTypeM3u8),
+      codec = VideoCodecWrapper(config.codecSettings.codec),
       assetType =
         None, // Not currently used as an asset, as the m3u8 playlist which combines this and subtitles is used instead
       output = () => {
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
@@ -1,31 +1,27 @@
 package com.gu.media.upload.mediaconvert.hls
 
 import com.gu.media.model.VideoSource
-import com.gu.media.upload.mediaconvert.OutputDefinition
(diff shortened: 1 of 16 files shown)


SHession (commented):
These changes look nice and straightforward. I checked the CODE output of a job and saw AV1 outputs being produced. 

Based on the description, I expected the AV1 HLS segments to be a smaller size than their H264 counterparts, but AWS S3 show them as the same. I may be missing something, how do we expect these changes to save bandwidth?

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
SHession, line 47: Should this be: `.codecSettings(config.codecSettings)`?

image

PR #1594, opened 2026-06-16
Hide tags if YouTube video

Fweddi:
## What does this change?
We shouldn't show self-hosted fields on YouTube videos.

This can be changed later, if YouTube videos need the tags field, but for now this will stop users filling in the wrong field by accident.

I still think there needs to be a clearer distinction between 'Composer keywords' and 'Tags' (it's quite confusing to users currently).


abeddow91 (approved):
Checked on CODE. Works as expected.

PR #1595, opened 2026-06-16
Use VTT subtitle output file instead of the uploaded one

waisingyiu:
## What does this change?

This pull request addresses an issue with uploaded subtitle files having a longer duration than their corresponding videos.

After a subtitle file is uploaded for a video, the "AddSubtitleToMP4" process in the video pipeline adds the subtitle stream from the uploaded subtitle file to the output MP4 files of MediaConvert. If the subtitle stream has a longer duration than the video stream, some video players may use this longer duration for the whole video, showing the last frame until the subtitle stream finishes.

Therefore, this pull request aims to have the subtitle stream truncated in the output MP4 files.

### Solution

We set the MediaConvert job to produce a subtitle file in VTT in the `FileOutputGroup` together with MP4 files (in two resolutions) and a thumbnail image. The output subtitle file is truncated so that it doesn't have subtitle text that goes beyond the length of the video track.

Instead of using the original subtitle file uploaded by users, this PR adds that output subtitle track to the output MP4 files, both of which come from the same output group of the MediaConvert job.

However, the MediaConvert job produces a VTT file even when there are no subtitles uploaded. So we add the subtitle track only if an uploaded subtitle file is present.
 
### Alternative approach

An alternative approach is to modify a subtitle track using `ffmpeg` before adding it to those MP4 files. I tried this approach, running `ffmpeg` with [`-t`](https://ffmpeg.org/ffmpeg.html#Subtitle-options:~:text=%2Dt%20duration%20(input/output)) option over a subtitle file. It successfully removed subtitle blocks that are shown after the duration, but it does not modify the subtitle block that appears before the duration ends and remains past the end of the duration. For example, if the subtitle file has this block:

```
3
00:00:10,000 --> 00:00:18,500
Subtitle from 10.00 to 19.00
```

Using `ffmpeg` to truncate it to 15 seconds will still leave this block unmodified in the output file, producing a subtitle file that is still 18.5 seconds long. 

The MediaConvert job used in our solution can handle this case without this issue.

In addition, we need to find out the length of a video track before truncating the subtitle track, which adds more complexity to this approach.

The `ffmpeg` supports an option `-shortest`, which tells it to "finish encoding when the shortest output stream ends". But as expected, it truncates the video track (that is undesirable) if the subtitle track doesn't have subtitle near the end of the video.

Deploy the main branch to CODE
- upload a video in MAM with a subtitle file that has longer duration.
- download the output MP4 file produced by MediaConvert job.
- open the MP4 file using VLC locally.
- we should see that the video has the same duration as the subtitle file, not the video we've uploaded.
- we should see the subtitle track in VLC.

Deploy this branch to CODE
- upload the same video in MAM with the same subtitle file (that has longer duration).
- download the output MP4 file produced by MediaConvert job.
- open the MP4 file using VLC locally.
- we should see that the video has the same duration as the original video we've uploaded.
- we should see the subtitle track in VLC.

Users are not required to manually trim subtitle files before uploading them to MAM.

Rely more on the output of MediaConvert job. But it is in the same output group as those MP4 files and as we already rely on the transcoding of MediaConvert, it should not be much of a concern.

### DIFF
--- a/uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
@@ -58,10 +58,8 @@ class AddSubtitlesToMP4
   private def getVttOutput(
       outputGroupDetails: MediaConvertOutputGroupDetails
   ): Option[String] = {
-    val vttOutputDetails = outputGroupDetails.outputDetails.find(
-      _.outputFilePaths.exists(_.endsWith(".vtt"))
-    )
-    vttOutputDetails.flatMap(_.outputFilePaths.find(_.endsWith(".vtt")))
+    // todo: use the mime type from the OutputDefinition instead of checking the file extension
+    outputGroupDetails.outputDetails.view.flatMap(_.outputFilePaths.find(_.endsWith(".vtt"))).headOption
   }
 
   override def handle(upload: Upload): Upload = {
@@ -72,10 +70,8 @@ class AddSubtitlesToMP4
       }.toList;
       event <- selfHostedUploadMetadata.completeEvent.toList;
       outputGroupDetails <- event.detail.outputGroupDetails;
-      outputVttPath <- getVttOutput(outputGroupDetails);
+      outputVttPath <- getVttOutput(outputGroupDetails) if upload.metadata.subtitleSource.isDefined;
       outputDetails <- outputGroupDetails.outputDetails;
-      // we don't need the subtitle source, but just meant to check if it exists
-      subtitleSource <- upload.metadata.subtitleSource;
       // todo: use the mime type from the OutputDefinition instead of checking the file extension
       path <- outputDetails.outputFilePaths.headOption if path.endsWith(".mp4")
     ) {
--- a/uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
@@ -59,7 +59,9 @@ class AddSubtitlesToMP4
       outputGroupDetails: MediaConvertOutputGroupDetails
   ): Option[String] = {
     // todo: use the mime type from the OutputDefinition instead of checking the file extension
-    outputGroupDetails.outputDetails.view.flatMap(_.outputFilePaths.find(_.endsWith(".vtt"))).headOption
+    outputGroupDetails.outputDetails.view
+      .flatMap(_.outputFilePaths.find(_.endsWith(".vtt")))
+      .headOption
   }
 
   override def handle(upload: Upload): Upload = {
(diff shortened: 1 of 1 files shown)


uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
```diff
@@ -58,7 +72,9 @@ class AddSubtitlesToMP4
       }.toList;
       event <- selfHostedUploadMetadata.completeEvent.toList;
       outputGroupDetails <- event.detail.outputGroupDetails;
+      outputVttPath <- getVttOutput(outputGroupDetails);
       outputDetails <- outputGroupDetails.outputDetails;
+      // we don't need the subtitle source, but just meant to check if it exists
       subtitleSource <- upload.metadata.subtitleSource;
```
davidfurey: You could make this more explicit
```suggestion
      outputVttPath <- getVttOutput(outputGroupDetails) if upload.metadata.subtitleSource.isDefined;
```

uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
```diff
@@ -50,6 +55,15 @@ class AddSubtitlesToMP4
     s3Client.putObject(putObjectRequest, path)
   }
 
+  private def getVttOutput(
+      outputGroupDetails: MediaConvertOutputGroupDetails
+  ): Option[String] = {
+    val vttOutputDetails = outputGroupDetails.outputDetails.find(
+      _.outputFilePaths.exists(_.endsWith(".vtt"))
+    )
+    vttOutputDetails.flatMap(_.outputFilePaths.find(_.endsWith(".vtt")))
```
davidfurey:
```suggestion
outputGroupDetails.outputDetails.view.flatMap(_.outputFilePaths.find(_.endsWith(".vtt"))).headOption
```

davidfurey (commented):
Looks reasonable. I've made a few minor suggestions

uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
davidfurey: Can you add the same todo as below:
`// todo: use the mime type from the OutputDefinition instead of checking the file extension`

waisingyiu: It looks great. Applied in the latest commit.

uploader/src/main/scala/com/gu/media/upload/AddSubtitlesToMP4.scala
waisingyiu: Great suggestion!

davidfurey (approved):

PR #1596, opened 2026-06-17
Add single source of truth for asset versioning

bryophyta:
## What does this change?

**Depends on https://github.com/guardian/editorial-tools-platform/pull/1067.**

### Background

Media Atoms can have 'assets' (videos), and assets are mapped to atoms many-to-one.

We keep track of an atom's assets via 'asset version' numbers, incrementing each time a new asset is added. The problem is that there are two paths for adding a new asset to an atom which have inconsistent logic for assigning asset version numbers, leading to potential overwriting of data or false joins between assets and asset metadata.

#### Method 1: Adding via a pre-existing URL

If the user has already uploaded a video directly to YouTube, they can add the video to an atom using the YouTube URL.

When adding via URL, the `AddAssetCommand` looks at the existing assets array in the Atom model, finds the highest existing asset version number, and then increments it by 1.

image

#### Method 2: Adding via the upload pipeline

Most assets are added via the upload pipeline, which is run via an AWS Step Function. To initially calculate a version number, the `UploadController.create` method follows the same logic as Method 1 above. However, it also adds an additional check: if a Step Function execution already exists with the name `${atomId}-${assetVersionNumber}` then it will increment the proposed version number and then try again.

#### Consequences

* Race conditions: if `AddAssetCommand` and `UploadController.create` are invoked at roughly the same time, they can end up trying to claim the same asset version number. For instance, if you kick off an upload, and then add via URL before the upload pipeline has updated the atom model, then the URL asset will be added but then its data will be overwritten in the atom model by the upload pipeline once the latter has finished.
* False joins: if you add an asset via the upload pipeline, a record of the upload is kept in the uploads cache table. We use this table to provide metadata in the UI about when an asset was added and by whom. This upload metadata is joined to assets via the `${atomId}-${assetVersionNumber}` key. If you then delete the asset from the atom model, its version number is left available for the URL method to claim, but the UI will still join the new asset with the upload metadata for the previous asset because they share a version number. (Upload, delete, upload doesn't face this issue because it has the extra step function execution name check.)

### Proposed fix

Both asset-upload methods should use a shared version number claiming mechanism, instead.

```mermaid
flowchart TD
    subgraph UploadRoute[Upload pipeline]
        UploadPipeline[UploadController.create] --> UPCC[Claim version]
        UPCC --> OW1[Other work]
    end

    subgraph UrlRoute[URL route]
        URL[AddAssetCommand] --> URLCC[Claim version]
        URLCC --> OW2[Other work]
    end

    UPCC <--> VCC[VersionClaimChecker]
    VCC <--> URLCC

    OW1 --> Atom[Atom Model]
    OW2 --> Atom
```

The proposed implementation is to [create a new DynamoDB table](https://github.com/guardian/editorial-tools-platform/pull/1067) which both methods can use to conditionally write a 'claim' for a given asset version number for a given atom. If the conditional write fails, then we increment the proposed number and try again (similar to how the pipeline currently does with step function execution names). This makes the dynamo records the source of truth for both methods, and guarantees that the version id is already claimed before the atom model itself gets updated with the new version.

As a side benefit, we can store the upload metadata fields (who added it, when, and what was the original filename) in this new table. This means that we can display this metadata for assets added via upload pipeline (as we are already) and also for assets added via URL.

### Deployment plan

The new table only effectively becomes the source of truth once it has data for the most recent asset version in it.

If we just roll this code change out without backfilling the new table, it is expected to work as it currently does when calculating version numbers. So the failure case there is not catastrophic.

But to make it more likely that we'll avoid any future clashes due to the old logic, we can backfill the new table using the upload metadata cache. There is a backfill script in this branch which has been tested against CODE. Given that we're trying to sync two DynamoDB tables, I don't think there's an effective way to guarantee consistency, so there's a possibility of people adding new assets between the time we start running of the backfill script and the deployment of the new code. But again, the worst-case scenario here is effectively a regression to the existing behaviour until the new asset versions overtake the old ones.

Steps:

- [x] Merge the PR to create the new table (https://github.com/guardian/editorial-tools-platform/pull/1067)
- [x] Run the backfill script against PROD
- [x] Merge this PR
- [x] Check that everything is working as expected:
    - [x] Adding via Upload and URL still work
    - [x] Race conditions no longer cause overwrites
    - [x] Double check that we don't see any kind of runaway claiming process arising from the tail-recursive claim function

### What about subtitles?

Subtitles are added and deleted by re-running a pre-existing upload payload, to reprocess the video with new subtitle file. This edits the asset version 'in place', pointing it to a new hosted video, rather than adding a new asset version.

This runs a bit counter to the current PR, but the mechanism still works with the new code. Adding or removing subtitles doesn't involve making a new version claim; it just re-uses the existing version number for the asset in question. The upload pipeline executions which make the changes are created using a different schema from the original upload (`${atomId}-${assetVersionNumber}-${subtitleVersionNumber}` instead of `${atomId}-${assetVersionNumber}`) so they are not in competition for the same space.

That said, the same sort of race condition bugs exist for subtitle versions for the same reason as asset versions, so it would be nice to bring them into the same sort of claim system at some point if we have the chance.

- [x] Run backfill script
- [x] Deploy branch to CODE
- [x] Try to replicate the race condition described above, verify that it *doesn't* produce a conflict/overwrite anymore.

## Accessibility

### DIFF
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -21,7 +21,10 @@ import com.gu.pandahmac.HMACAuthActions
 import data.{DataStores, UnpackedDataStores}
 import com.gu.ai.x.play.json.Jsonx
 import com.gu.media.telemetry.{TagLong, TagString, Telemetry}
-import model.commands.CommandExceptions.commandExceptionAsResult
+import model.commands.CommandExceptions.{
+  AssetVersionClaimFailed,
+  commandExceptionAsResult
+}
 import model.commands.{SubtitleFileDeleteCommand, SubtitleFileUploadCommand}
 import org.scanamo.{ConditionNotMet, Table}
 import org.scanamo.generic.auto._
@@ -127,7 +130,12 @@ class UploadController(
         version = MediaAtomHelpers.getNextAssetVersion(thriftAtom.tdata),
         userEmail = userEmail,
         originalFilename = Some(req.filename)
-      )
+      ) match {
+        case Right(claimedVersion) => claimedVersion
+        case Left(error) =>
+          log.error(error.message)
+          AssetVersionClaimFailed(error.message)
+      }
 
       val upload = start(atom, userEmail, req, assetVersion)
 
@@ -253,17 +261,18 @@ class UploadController(
       log.warn(
         s"Execution already exists for atom ${atom.id} version ${assetVersion}. Trying next available version. NB. This should not happen unless the backfill script has failed, or an upload was started during the backfill process."
       )
-      start(
-        atom,
-        email,
-        req,
-        assetVersionManager.claimThisOrNextAvailableVersion(
-          atom.id,
-          assetVersion,
-          userEmail = email,
-          originalFilename = Some(req.filename)
-        )
-      )
+      assetVersionManager.claimThisOrNextAvailableVersion(
+        atom.id,
+        assetVersion,
+        userEmail = email,
+        originalFilename = Some(req.filename)
+      ) match {
+        case Right(nextVersion) =>
+          start(atom, email, req, nextVersion)
+        case Left(error) =>
+          log.error(error.message)
+          AssetVersionClaimFailed(error.message)
+      }
     }
   }
 
--- a/app/controllers/UploadController.scala
+++ b/app/controllers/UploadController.scala
@@ -1,10 +1,7 @@
 package controllers
 
 import com.gu.ai.x.play.json.Encoders._
-import software.amazon.awssdk.services.sfn.model.{
-  ExecutionAlreadyExistsException,
-  ExecutionListItem
-}
+import software.amazon.awssdk.services.sfn.model.ExecutionListItem
 import com.gu.media.MediaAtomMakerPermissionsProvider
 import com.gu.media.logging.Logging
 import com.gu.media.model.{
@@ -23,8 +20,9 @@ import com.gu.ai.x.play.json.Jsonx
 import com.gu.media.telemetry.{TagLong, TagString, Telemetry}
 import model.commands.CommandExceptions.commandExceptionAsResult
 import model.commands.{SubtitleFileDeleteCommand, SubtitleFileUploadCommand}
-import org.scanamo.Table
+import org.scanamo.{ConditionNotMet, Table}
 import org.scanamo.generic.auto._
+import org.scanamo.syntax._
 import play.api.libs.Files
 import play.api.libs.json.{Format, Json}
 import play.api.mvc.{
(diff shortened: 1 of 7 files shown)


app/util/AssetVersionManager.scala
```diff
@@ -0,0 +1,95 @@
...
+          userEmail = userEmail,
+          originalFilename = originalFilename
+        )
+      case Left(_) =>
+        log.warn(
+          s"Unexpected error claiming version $version for atom $atomId. Retrying with next version."
+        )
+        claimThisOrNextAvailableVersion(
+          atomId = atomId,
+          version = version + 1,
```
lindseydew: Would we want to increment the version in the case where dynamo errors for a different reason other than the condition not met? Could we end up in an infinite retry loop if dynamo was down?

app/controllers/UploadController.scala
```diff
@@ -233,8 +243,28 @@ class UploadController(
...
-      start(atom, email, req, assetVersion + 1)
+    /** @todo
+      *   once the new asset version claim mechanism has been live long enough
+      *   for any existing executions to have been flushed through, we should be
+      *   able to remove this path, and then `start()` will no longer need to be
+      *   recursive.
+      */
+    case _: ExecutionAlreadyExistsException => {
+      log.warn(
+        s"Execution already exists for atom ${atom.id} version ${assetVersion}. Trying next available version. NB. This should not happen unless the backfill script has failed, or an upload was started during the backfill process."
```
lindseydew: I wonder whether in this case it would be safer to fail the upload given it's a state the new behaviour shouldn't allow us into?

scripts/backfill-asset-versions.sh
```diff
@@ -0,0 +1,236 @@
...
+  process_page "$SCAN_RESULT"
+
+  NEXT_TOKEN=$(echo "$SCAN_RESULT" | jq -r '.NextToken // empty')
+  if [[ -z "$NEXT_TOKEN" ]]; then
+    break
+  fi
+
+  PAGE=$((PAGE + 1))
+done
+
```
lindseydew, line 234: I think this makes sense, though it's a little hard to follow using bash. It might be a bit clearer to make a scala script? though appreciate it's just for a one off job so may not be worth a rewrite

scripts/backfill-asset-versions.sh
```diff
@@ -0,0 +1,236 @@
...
+    # Build item JSON safely with jq to prevent injection
+    local item_json
+    item_json=$(jq -n \
+      --arg id "$claim_id" \
+      --argjson ts "$claim_timestamp" \
+      --arg fn "$original_filename" \
+      '{
+        PutRequest: {
+          Item: (
+            {id:{S:$id}, claimSource:{S:"UploadPipeline"}, claimedAtTimestamp:{N:($ts|tostring)}, claimedByUser:{S:"backfill"}}
```
lindseydew: could it be worth marking this a backfill type?

lindseydew (commented):
This is looking great. I do feel slightly uneasy about a backfill for keeping a source of truth up to date as it can rely on execution timings and be a bit tricky to ascertain if anything has gone out of sync. Though I feel I would need to understand in a little more detail how the cache table is tracking things in case there are alternatives we could explore. Perhaps we could have a discussion?

Looking great though, this is a much more robust way of allocating version ids

app/util/AssetVersionManager.scala
bryophyta: good point, thanks! Yeah we probably want to just exit and fail the upload in this case, right?

app/controllers/UploadController.scala
lindseydew: Having had a discussion, it makes sense to try the database for a new version in this case. I think it would be helpful to update the logging statement as a catch all case rather than some specific to a one off back fill

scripts/backfill-asset-versions.sh
bryophyta: ah, I didn't address this! I did consider that, but the plan would be to use this table as a source of asset metadata later, in which case we want to preserve the fact that the original upload was via the pipeline..

I could add a separate boolean backfill field or something like that -- what do you think?

lindseydew (approved):
LGTM 👍  just a note about the logging statement

scripts/backfill-asset-versions.sh
lindseydew: yeah, that may end up being useful for debugging down the line 👍

PR #1597, opened 2026-06-18
Update video options for non-youtube (default) videos

abeddow91:
## What does this changes
Updates the list of supported / unsupported features for non-youtube videos.

Deployed to CODE and checked the list of supported / unsupported features was correct. 

List is up-to-date
N/A

### Before
Screenshot 2026-06-18 at 13 55 33

### After
Screenshot 2026-06-18 at 15 51 54


waisingyiu (approved):
💯

PR #1599, opened 2026-06-18
Make UploadMetadata.version non-optional

bryophyta:
## What does this change?

Converts the `UploadMetadata.version` field from `Option[Long]` to `Long`.

As far as I can tell, we would *like* the version number to be required, rather than optional. It would also make it easier to make the way we version assets more consistent.

### Existing writes

There are no code paths in MAM where we had been constructing a `UploadMetadata` instance without providing a `Some(Long)` value.

### Existing reads

We read UploadMetadata from two places:

1. The `Upload` payload that's passed around as input for the step function steps.
2. The upload cache DynamoDB table.

The step functions consume inputs that are generated by other parts of this codebase. As mentioned above, it was never being set to `None` in the first instance. And on top of that, the step functions shouldn't be editing the version number.

The upload cache table is read in order to enhance the assets view in the MAM UI (populated by `GET /api/uploads    controllers.UploadController.list(atomId)`). If the `metadata.version` field is missing in the table, then Scanamo would fail to validate the item as an instance of `UploadMetadata`.

#### What happens if `metadata.version` is missing in the DynamoDB table?

The entry for a given asset version `n` for a given atom `A` will only be read on demand when hitting `/api/uploads/A`. If `A` currently has an asset with version `1`, it will look up `A-1` in the table. If the item is malformed, Scanamo will treat it the same as if the item wasn't present. In this case, the application code falls back to querying the step function history. If it finds a valid record in either case then it will add some metadata to `A-1` in the response, which populates the original filename and upload time in the UI.

image

If it doesn't find a valid record, it defaults to a plain record without this extra metadata:

image

So, the fallback behaviour is sensible -- the worst case scenario is that we miss this bit of metadata.

#### Are there any records in the Dynamo table without this property?

I scanned the table and the only items I found in the DynamoDB table that were missing the `metadata.version` field had id keys of an unexpected format, e.g. `094f599d-609c-4d4a-9afd-70f388f290e2--56045e10-b84a-4940-8953-d7c97254ee47` instead of `094f599d-609c-4d4a-9afd-70f388f290e2-1`.

As a result, the uploads endpoint doesn't find these items anyway, so it doesn't show the extra metadata in the UI for these assets. It will continue to miss these items, so they won't actually fail to validate because they won't be read in at all. However, if they were read in and failed to validate then as described above the fallback behaviour would be equivalent to now: no extra metadata in the UI.

We considered backfilling these upload records to the recognised format, but we can't say with confidence which asset versions they each would apply to, so it feels more sensible for now to leave them as they are rather than potentially fabricating which belongs to which asset. The multimedia folks have expressed an interest in keeping this data though, so we're leaving it in the database table.

All of the problematic assets uploads are from 2017, and I believe they are all entries that were added before https://github.com/guardian/media-atom-maker/pull/618

- [x] Deploy to CODE
- [x] Find an old atom whose Upload pipeline runs won't be stored any more (e.g. `6e82d086-e34a-4a88-95ab-10d32821de11`)
- [x] Delete the `version` field from the `media-atom-pipeline-cache-CODE` for an asset associated with that atom
- [x] Refresh the page and note that the upload time and original filename no longer show, but other than that everything works as usual.

## Accessibility

### DIFF
--- a/app/util/UploadBuilder.scala
+++ b/app/util/UploadBuilder.scala
@@ -42,7 +42,7 @@ object UploadBuilder {
         startTimestamp = Instant.ofEpochMilli(currentTimestamp)
       ),
       originalFilename = Some(request.filename),
-      version = Some(assetVersion),
+      version = assetVersion,
       startTimestamp = currentTimestamp,
       hasAudio = None
     )
@@ -69,7 +69,7 @@ object UploadBuilder {
       upload: Upload,
       newSubtitleSource: Option[VideoSource]
   ): Upload = {
-    val assetVersion = upload.metadata.version.getOrElse(1L)
+    val assetVersion = upload.metadata.version
     val subtitleVersion = Upload.getNextSubtitleVersion(upload)
     val updatedAsset = getAsset(
       upload.metadata.selfHost,
--- a/common/src/main/scala/com/gu/media/upload/model/UploadMetadata.scala
+++ b/common/src/main/scala/com/gu/media/upload/model/UploadMetadata.scala
@@ -18,7 +18,7 @@ case class UploadMetadata(
     pluto: PlutoSyncMetadataMessage,
     iconikData: Option[IconikData],
     runtime: RuntimeUploadMetadata,
-    version: Option[Long] = None,
+    version: Long,
     selfHost: Boolean = false,
     asset: Option[VideoAsset] = None,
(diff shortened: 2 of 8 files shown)


lindseydew (approved):
LGTM 👍

PR #1600, opened 2026-06-19
Switch HLS output to CMAF (in preparation for AV1)

davidfurey:
### Summary

This PR migrates the MediaConvert HLS container format from M3_U8 (MPEG-TS HLS) to CMFC (CMAF/fMP4), as  a prerequisite for future AV1 support. It also includes two bug fixes discovered during development.

The AV1 change requires changing container type because AWS do not allow M3U8 to include AV1. CMFC is very similar to M3U8 but allows AV1. I believe that rendering devices that support M3U8/HLS will seamlessly support CMAF. The switch from M3_U8 to CMFC also causes a (hopefully incidental) change of the segment characteristics from TS to CMAF/fMP4. Instead of the segments being lots of small files, there is one file per stream, and the segments are byte-ranges referenced by the manifest.

#### CMAF migration (`HLSOutputGroup`, `VideoOutput`, `CaptionsOutput`, `AudioOutput`)

- Switches the HLS output group from `HLS_GROUP_SETTINGS` / `HlsGroupSettings` to `CMAF_GROUP_SETTINGS` / `CmafGroupSettings`.
- HLS outputs now use `ContainerType.CMFC` with `CmfcSettings` instead of `ContainerType.M3_U8` with `M3u8Settings`.
- Extracts audio into a dedicated `AudioOutput` object with its own CMFC container, so audio/video are in separate tracks — required for CMAF.
- The `hasAudio` flag is moved from `VideoOutput` to the output group level: audio-only outputs are appended conditionally (`AudioOutput()`) rather than baked into video outputs.
- Output group renamed from `"Apple HLS"` to `"CMAF HLS"`.
- Name modifiers updated: `"captions"` → `"_captions"` (consistent with `"_audio"`).
- `writeDashManifest` explicitly disabled (CMAF can generate both HLS and DASH manifests; we only want HLS).
- CMAF and M3U8 are closely related formats and the output should remain compatible with iOS.

#### `outputFilePaths` is now optional (`MediaConvertEvent`)

- `outputFilePaths` in `MediaConvertOutputDetails` changed from `List[String]` to `Option[List[String]]`.
- CMAF outputs do not include `outputFilePaths` in MediaConvert completion events (the field is absent from the JSON), which was causing deserialisation failures.
- `AddSubtitlesToMP4` and `TranscoderComplete` updated to handle the `Option`.
- Comprehensive new test case added covering a real CMAF completion event with a mix of file-group outputs (with paths) and CMAF outputs (without paths).

#### Null-safe error handling in `StepFunctions`

- The `cause` field in a Step Functions execution failure event can be `null`, which previously caused a `NullPointerException` before the `JsonParseException` / `JsResultException` catch could fire.
- Added an explicit `NullPointerException` case returning a friendly fallback message (`"Execution failed but no error message was provided"`).

## Testing

We need to verify that this renders correctly on Android, iOS and web

### DIFF
--- a/app/util/StepFunctions.scala
+++ b/app/util/StepFunctions.scala
@@ -46,7 +46,10 @@ class StepFunctions(awsConfig: AWSConfig) {
         try {
           Some((Json.parse(cause) \ "errorMessage").as[String])
         } catch {
-          case _: JsonParseException | _: JsResultException =>
+          case _: NullPointerException =>
+            Some("Execution failed but no error message was provided")
+          case _: JsonParseException | _: JsResultException |
+              _: NullPointerException =>
             Some(cause)
         }
     }
--- a/app/util/StepFunctions.scala
+++ b/app/util/StepFunctions.scala
@@ -48,8 +48,7 @@ class StepFunctions(awsConfig: AWSConfig) {
         } catch {
           case _: NullPointerException =>
             Some("Execution failed but no error message was provided")
-          case _: JsonParseException | _: JsResultException |
-              _: NullPointerException =>
+          case _: JsonParseException | _: JsResultException =>
             Some(cause)
         }
     }
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
@@ -1,0 +1,31 @@
+package com.gu.media.upload.mediaconvert.hls
+
+import com.gu.media.model.VideoSource
+import com.gu.media.upload.mediaconvert.OutputDefinition
+import com.gu.media.upload.mediaconvert.SharedCodecSettings.aacAudioDescription
+import software.amazon.awssdk.services.mediaconvert.model._
+
+object AudioOutput {
+  def apply(): OutputDefinition =
+    OutputDefinition(
+      mimeType = Some(VideoSource.mimeTypeM3u8),
+      assetType =
+        None, // Not currently used as an asset, as the m3u8 playlist is used instead
+      output = () => {
+        val outputBuilder = Output
+          .builder()
+          .containerSettings(
+            ContainerSettings
+              .builder()
+              .container(ContainerType.CMFC)
+              .cmfcSettings(CmfcSettings.builder().build())
+              .build()
+          )
+          .nameModifier("_audio")
+          .audioDescriptions(aacAudioDescription)
+
+        outputBuilder.build()
+      }
+    )
+
+}
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
@@ -12,7 +12,7 @@ object AudioOutput {
       assetType =
         None, // Not currently used as an asset, as the m3u8 playlist is used instead
       output = () => {
-        val outputBuilder = Output
+        Output
           .builder()
           .containerSettings(
             ContainerSettings
@@ -23,8 +23,7 @@ object AudioOutput {
           )
           .nameModifier("_audio")
           .audioDescriptions(aacAudioDescription)
-
-        outputBuilder.build()
+          .build()
       }
     )
 
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
@@ -8,7 +8,7 @@ import software.amazon.awssdk.services.mediaconvert.model._
 object AudioOutput {
   def apply(): OutputDefinition =
     OutputDefinition(
-      mimeType = Some(VideoSource.mimeTypeM3u8),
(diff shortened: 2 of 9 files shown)


app/util/StepFunctions.scala
```diff
@@ -46,7 +46,10 @@ class StepFunctions(awsConfig: AWSConfig) {
         try {
           Some((Json.parse(cause) \ "errorMessage").as[String])
         } catch {
-          case _: JsonParseException | _: JsResultException =>
+          case _: NullPointerException =>
+            Some("Execution failed but no error message was provided")
+          case _: JsonParseException | _: JsResultException |
+              _: NullPointerException =>
```
abeddow91: Should either this `NullPointerException` case or the one on line 49 be removed?

davidfurey: Good spot

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.gu.media.model.VideoSource
+import com.gu.media.upload.mediaconvert.OutputDefinition
+import com.gu.media.upload.mediaconvert.SharedCodecSettings.aacAudioDescription
+import software.amazon.awssdk.services.mediaconvert.model._
+
+object AudioOutput {
+  def apply(): OutputDefinition =
+    OutputDefinition(
+      mimeType = Some(VideoSource.mimeTypeM3u8),
```
abeddow91: Is this the correct mimeType for audio output?

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
```diff
@@ -0,0 +1,31 @@
...
+              .builder()
+              .container(ContainerType.CMFC)
+              .cmfcSettings(CmfcSettings.builder().build())
+              .build()
+          )
+          .nameModifier("_audio")
+          .audioDescriptions(aacAudioDescription)
+
+        outputBuilder.build()
+      }
```
abeddow91, line 27:
```suggestion
      output = () => {
        Output
          .builder()
          .containerSettings(
            ContainerSettings
              .builder()
              .container(ContainerType.CMFC)
              .cmfcSettings(CmfcSettings.builder().build())
              .build()
          )
          .nameModifier("_audio")
          .audioDescriptions(aacAudioDescription)
          .build()
      }
```

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/HLSOutputGroup.scala
```diff
@@ -30,17 +31,18 @@ object HLSOutputGroup {
...
             OutputGroupSettings
               .builder()
-              .`type`(OutputGroupType.HLS_GROUP_SETTINGS)
-              .hlsGroupSettings(
-                HlsGroupSettings
+              .`type`(OutputGroupType.CMAF_GROUP_SETTINGS)
+              .cmafGroupSettings(
+                CmafGroupSettings
                   .builder()
+                  .writeDashManifest(CmafWriteDASHManifest.DISABLED)
```
abeddow91, line 43: Im assuming that, without this, CMAF outputs a  DASH manifest automatically (which we have no use for)?

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
```diff
@@ -1,31 +1,27 @@
...
           .containerSettings(
             ContainerSettings
               .builder()
-              .container(ContainerType.M3_U8)
-              .m3u8Settings(M3u8Settings.builder().build())
+              .container(ContainerType.CMFC)
+              .cmfcSettings(CmfcSettings.builder().build())
               .build()
           )
-          .nameModifier(s"hls${config.nameModifier}")
```
abeddow91, line 28: Why are we OK to drop `hls` from the file name?

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/AudioOutput.scala
davidfurey: No, it is not. The mimeType here should be `None`. This asset isn't used directly so shouldn't have an assetType or a mimeType

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/HLSOutputGroup.scala
davidfurey: Exactly! I thought it it would be useful for Firefox, but it turns out FF doesn't support DASH or HLS. I don't know of anything that supports DASH but not HLS, so I don't see any point in generating DASH.

uploader/src/main/scala/com/gu/media/upload/mediaconvert/hls/VideoOutput.scala
davidfurey: I'm not really sure why we ever needed `hls` in the name, since it is already obvious from the file extension

abeddow91 (approved):

PR #1601, opened 2026-06-19
Define encoding config less verbosely

davidfurey:
### Summary

Refactor of `EncodingConfig` and `HLSOutputGroup` to eliminate the repetitive hand-enumerated encoding config objects, replacing them with a data-driven approach. The previous approach is not sustainable when we add AV1 as an additional codec. Also adds a quality suffix to the q8 outputs so they are identifiable in log analysis.

#### `EncodingConfig.scala`

- `EncodingConfig` is now an abstract type with a derived `nameModifier` implementation — the name is computed from `dimensions` and `qualityLevel` rather than hardcoded per variant (e.g. `_480w_q8`, `_720h_q6`).
- Adds a concrete subclass `H264EncodingConfig(dimensions, qualityLevel)` which carries the `VideoCodecSettings` for H.264, making it straightforward to add other codecs (e.g. AV1) later by adding further subclasses.
- `EncodingConfigs` is simplified to just `Default` and `MobileWidth` as named aliases used by the FileOutputGrouo. The full matrix of quality variants used in HLS/M3U8 is now generated in `HLSOutputGroup`.
- The q8 outputs now include the quality in their name modifier (e.g. `_480w_q8` instead of `_480w`), making naming consistent and visible across all outputs in logs.

#### `HLSOutputGroup.scala`

- The explicit list of six `VideoOutput(...)` calls is replaced by a for-comprehension over `(codec factory, dimensions, quality levels)` tuples, currently containing a single entry for H.264 across `[480w, 720h]` × `[q8, q6, q4]`.
- Adding a new codec (e.g. AV1) will require adding one line to `outputDefinitions` rather than duplicating six more `VideoOutput` calls.
- Audio/captions output construction is unchanged in behaviour.

### DIFF
--- a/uploader/src/main/scala/com/gu/media/upload/mediaconvert/EncodingConfig.scala
+++ b/uploader/src/main/scala/com/gu/media/upload/mediaconvert/EncodingConfig.scala
@@ -1,47 +1,40 @@
 package com.gu.media.upload.mediaconvert
 
+import com.gu.media.upload.mediaconvert.SharedCodecSettings.{h264Settings}
+import software.amazon.awssdk.services.mediaconvert.model.{
+  VideoCodec,
+  VideoCodecSettings
+}
+
 case class Dimensions(width: Option[Int], height: Option[Int])
 
 sealed trait EncodingConfig {
   def dimensions: Dimensions
-  def nameModifier: String
+  def nameModifier: String = {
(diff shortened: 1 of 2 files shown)


waisingyiu (approved):
Nice work!

PR #1605, opened 2026-06-29
Define encoding config less verbosely

abeddow91:
This PR restores https://github.com/guardian/media-atom-maker/pull/1601 so that it can be merged into main following an accidental early merge. 

More details can be found on the original PR. No changes have been made from PR1601


waisingyiu (approved):
👍

PR #1612, opened 2026-06-30
Revert "Revert "Switch HLS output to CMAF (in preparation for AV1)""

abeddow91:
Reverts guardian/media-atom-maker#1607

Re-introduces https://github.com/guardian/media-atom-maker/pull/1600 which was reverted in https://github.com/guardian/media-atom-maker/pull/1607 whilst an error with uploading atoms was investigated.

This branch also introduces change to `hasAudio`, detailed below.

**Fix audio detection logic so that it default to false for hasAudio on detection failure.**

hasAudio decides whether an AudioOutput is added to the HLSOutputGroup (see HLSOutputGroup.scala). Defaulting to true would adds an empty audio output to the HLS outputs and produce a broken transcode, so false is the safe fallback when we can't determine whether audio is present.

The caveat here is that if audio detection check fails for a video that actually does have audio, the outputted video would be outputted without audio.

### DIFF
--- a/app/util/StepFunctions.scala
+++ b/app/util/StepFunctions.scala
@@ -46,6 +46,8 @@ class StepFunctions(awsConfig: AWSConfig) {
         try {
           Some((Json.parse(cause) \ "errorMessage").as[String])
         } catch {
+          case _: NullPointerException =>
+            Some("Execution failed but no error message was provided")
           case _: JsonParseException | _: JsResultException =>
             Some(cause)
         }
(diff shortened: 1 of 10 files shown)


waisingyiu (approved):
Agree that it is more sensible to set the `hasAudio` to false with cmaf package if the detection fails.

It is out of the scope of this ticket, but I think it may be good to look at why the audio detection fails in future, in case a video with audio is uploaded but the output video has no audio.

PR #1615, opened 2026-07-02
Fix first frame image name mismatch

abeddow91:
## What does this change?
Add an optional quality suffix match to the regex for image name stripper along with a test. 

This is required as the image name is created on dynamically from the mp4 video name. As the mp4 can be appended with various suffixes, these need to be stripped out to ensure the names match. Without this, poster images aren't stored on the atom as the image can't be found. 

As the legacy comment states, the process should be made less brittle by using the MediaConvert event for the jpeg name, rather than dynamically building it.


waisingyiu (approved):
👍

PR #1618, opened 2026-07-07
Convert DeleteButton to ts

junyuanxue:
## What does this change?

As in title :)

* Go to a video, e.g. 'DF test' in the video grid
* Click on 'Edit assets' CTA under 'Video Preview' section
* You should see the 'Delete' button under video trail

Screenshot 2026-07-07 at 17 05 25


public/video-ui/src/components/DeleteButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+export default function DeleteButton({
+  tooltip,
+  onDelete,
+  disabled,
+  tooltipWhenDisabled
+}: DeleteButtonProps) {
+  const [confirmDelete, setConfirmDelete] = useState(false);
+  const resetTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const changeState = () => {
```
bryophyta: I know that `changeState` is the name in the original file, but I wonder if this is a good opportunity to give this function a slightly more descriptive name?

public/video-ui/src/components/DeleteButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Icon from './Icon';
+
+type DeleteButtonProps = {
+  tooltip: string;
+  onDelete: () => void;
+  disabled?: boolean;
+  tooltipWhenDisabled?: string;
+};
+
+export default function DeleteButton({
```
bryophyta: not essential, but: shall we convert this to a named export instead of default? (this is a rule that we enforce via a shared eslint config in many of our TS projects, although we haven't added those linting rules to this repo yet)

bryophyta (commented):
This looks great! I've just added a couple of small suggestions for coding style but feel free to push back on these if you disagree!

public/video-ui/src/components/DeleteButton.tsx
junyuanxue: Sure thing!

public/video-ui/src/components/DeleteButton.tsx
junyuanxue: Sounds good - personally I'm generally in favour of named exports too :)

bryophyta (approved):

PR #1619, opened 2026-07-07
Convert EditSaveCancel to ts

junyuanxue:
## What does this change?

* As in title
* I've updated the interface around `canSave` and `canCancel` to keep the data flow simple as there aren't any strong reasons why they have to be functions, but please let me know if I've missed something 

* Go to any video
* Click on the 'Edit' button, and you should see 'Save changes' and 'Cancel' buttons

Screenshot 2026-07-07 at 17 10 06
Screenshot 2026-07-08 at 12 10 40


public/video-ui/src/components/EditSaveCancel.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import Icon from './Icon';
+
+type EditSaveCancelProps = {
+  editing: boolean;
+  onEdit: () => void;
+  onSave: () => void;
+  onCancel: () => void;
+  canSave: boolean;
+  canCancel?: boolean;
```
bryophyta, line 10: In the `propTypes` definition in the original JS code, `canCancel` is marked as a function, rather than a boolean. From what I can see, it looks as though we are indeed passing a `() => boolean` function as the prop value for this component e.g. via the FurnitureTabPanel: https://github.com/guardian/media-atom-maker/blob/467a80c17b5ab608918917d0a8569b02ff19d524/public/video-ui/src/pages/Video/index.jsx#L345

But a function which _returns_ false (e.g. `() => false`) will itself be truthy, so `!(() => false) === false`, which I don't think we want.

```js
const canCancel = () => false;

!canCancel === false; // evaluates to true, though it should be false
```

Possibly I'm missing something, so please do let me know if so! But otherwise I think we want to either keep this as a function, or else update the call sites to pass a boolean?

bryophyta: sorry the rest of the diff didn't load for me at first -- I can see now that you've addressed this already!!

public/video-ui/src/pages/Video/tabs/Workflow.jsx
```diff
@@ -21,7 +21,8 @@ export class WorkflowTabPanel extends React.Component {
     onEdit: PropTypes.func.isRequired,
     onSave: PropTypes.func.isRequired,
     onCancel: PropTypes.func.isRequired,
```
bryophyta, line 23: canCancel is duplicated

```suggestion
```

junyuanxue: One is `onCancel` and one is `canCancel` - unless I've missed something? 👀

bryophyta (approved):
lgtm, and great shout to convert the `() => boolean` props to just `boolean`; I think it's much easier to read 👍

public/video-ui/src/pages/Video/tabs/Workflow.jsx
bryophyta: ah yep, totally! 🤦

PR #1624, opened 2026-07-09
Only fetch youtube category and channels for youtube videos

junyuanxue:
## What does this change?
[Asana ticket for context](https://app.asana.com/1/1210045093164357/project/1213625402425263/task/1215374293358916?focus=true)
We shouldn't need to fetch Youtube categories and channels for non-Youtube videos

* Go to a non-YouTube video in MAM
* Click on 'Edit Assets' CTA under 'Video Preview'
* You should no longer see the 'Could not get YouTube channels' error banner

| Before   | After   |
| :----- | :---------------------------------------------------------------------- |
| Screenshot 2026-07-09 at 14 13 45 | Screenshot 2026-07-09 at 14 13 30 |


bryophyta (approved):
nice! this looks great and works as expected 👍

PR #1625, opened 2026-07-09
09 07 update removeStringTagDuplicates to ts

AlexAscherson:
## What does this change?

Updates TagParsers and Remove String Duplicates to TS

Defines two new shared tag types, DisplayTag and ParsedTag

2 individual types currently not in shared types file, CapiApiResponse and StringTag.  These could potentially be moved into the shared space if they need to be used anywhere else.

Unit tests

## Accessibility


public/video-ui/src/util/removeStringTagDuplicates.ts
```diff
@@ -1,31 +1,39 @@
-export default function removeStringTagDuplicates(tag, tagValue) {
+import { DisplayTag } from '../types/tags';
+
+type ExistingTagValue = string | { id: string };
+
+export default function removeStringTagDuplicates(
```
hk-15: optional: nice opportunity to convert this to a named export?

public/video-ui/src/util/tagParsers.ts
```diff
@@ -0,0 +1,66 @@
+import ContentApi from '../services/capi';
+import TagTypes from '../constants/TagTypes';
+import { DisplayTag, ParsedTag } from '../types/tags';
+
+type CapiTagResponse = {
```
hk-15: Question: is it worth putting it in shared space so it's easier to find in future? Or would that be "polluting" the shared space?

hk-15 (approved):
Looks good and works as expected when tested locally. Just an optional suggested change and a quick Q from me.

public/video-ui/src/util/removeStringTagDuplicates.ts
AlexAscherson: Done!

public/video-ui/src/util/tagParsers.ts
AlexAscherson: I agree with this, at the moment I've only come across one place where its being used, but was thinking I would move it to the TagTypes(?) file when there is a second use of it.  

I've gone ahead and done it now, since it seems sensible.

PR #1628, opened 2026-07-10
Fix clientside state around scheduling and embargo

junyuanxue:
## What does this change?
[Asana for context](https://app.asana.com/1/1210045093164357/project/1213625402425263/task/1213633225640663?focus=true)
* When a video is opened in multiple tabs, the UI state around scheduling and embargo isn't synced with BE response if the API call fails
* Currently, we set an optimistic update in the state, but that state isn't reverted when the API returns an error. Ideally, without a robust rollback mechanism, we should avoid optimistic update and point to BE as source of truth and only update the state if the API request is successful

To solve this particular problem: 
* I've removed the optimistic state update when saving the video. In a downstack PR (https://github.com/guardian/media-atom-maker/pull/1629) I've added some loading states to make up for that.
* I've updated the error message in case of this 409 error to `Could not save video as another user (or tab) has made changes to this video. Please refresh the page and try again.` to prompt user to refresh and sync to the latest video state. Alternatively we could also refetch the video for the user and update the state, but this seems to be a simple and clear solution.

fix fe state for scheduling and embargo

### DIFF
--- a/public/video-ui/src/actions/VideoActions/saveVideo.ts
+++ b/public/video-ui/src/actions/VideoActions/saveVideo.ts
@@ -11,14 +11,13 @@ import { AppDispatch } from '../../util/setupStore';
 const errorMessages = {
   saveVideoDefault: 'Could not save video',
   saveVideoConflict:
-    'Could not save video as another user (or tab) is currently editing this video',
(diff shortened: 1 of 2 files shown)


bryophyta (approved):
nice! I agree that it seems sensible to remove the potentially misleading optimistic update here 👍

PR #1629, opened 2026-07-10
Refactor components for schedule launch and embargo

junyuanxue:
## What does this change?
* A follow-up to https://github.com/guardian/media-atom-maker/pull/1628
* The logic within component `ScheduledLaunch` felt quite hard to follow. There are wan too many states controlling the UI and business logic. This PR aims to simplify that while moving the component over to TS
  * I have largely reduced the number of states involved in handling the display logic and business logic. So now we'll simply have: 1) menu control and which menu option (`DelayPublicationOption`) is selected; 2) selected date; 3) error and loading states
  * Ideally, with the help of a component library with built-in A11y best practices we could delegate menu control state away entirely. But that'd be a separate bigger project
* The current scheduling UI looks broken. This PR fixes it as well as addressing HTML complaints by wrapping `DatePicker` elements within a `fieldset`
Screenshot 2026-07-13 at 10 54 11

Jul-13-2026 12-35-25

## Accessibility

-   [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)

---
  - https://app.asana.com/0/0/1216434586453464


public/video-ui/src/components/ScheduledLaunch.tsx
```diff
@@ -0,0 +1,296 @@
...
+    delayPublicationOption: DelayPublicationOption
+  ) => {
+    setSelectedDelayPublicationOption(delayPublicationOption);
+    setShowMenu(false);
+  };
+
+  const handleIndefiniteEmbargoClick = () => {
+    setIsLoading(true);
+
+    saveVideo({
```
bryophyta: I believe that the actual `saveVideo` function is asynchronous, so if we call it without awaiting then `setIsLoading(false)` will be immediately invoked , meaning that the loading state won't actually show

public/video-ui/src/components/ScheduledLaunch.tsx
```diff
@@ -0,0 +1,296 @@
...
+import { isAfter, isFutureDate, isSameOrAfter } from '../util/dateHelpers';
+import { impossiblyDistantDate } from '../constants/dates';
+import VideoUtils from '../util/video';
+import type { Video } from '../services/VideosApi';
+
+type DelayPublicationOption = 'scheduledLaunch' | 'embargo';
+
+type ScheduledLaunchProps = {
+  video: Video;
+  saveVideo: (video: Video) => void;
```
bryophyta:
```suggestion
  saveVideo: (video: Video) => Promise<void>;
```

I think this should be `Promise<void>`

public/video-ui/src/components/ScheduledLaunch.tsx
```diff
@@ -0,0 +1,296 @@
...
+          date: impossiblyDistantDate
+        }
+      }
+    });
+
+    setIsLoading(false);
+    setShowMenu(false);
+  };
+
+  const handleDateChange = (date: number) => {
```
bryophyta: It looks like the DatePicker component might invoke this callback with `null`, so I think we should update the type to be `(date: number | null)`?

https://github.com/guardian/media-atom-maker/blob/a59cb0b3f968d7f0349a2fe64bb2fc58c204802f/public/video-ui/src/components/FormFields/DatePicker.jsx#L89-L91

bryophyta: oh wait, maybe this intervening wrapper handles that actually..?

https://github.com/guardian/media-atom-maker/blob/a59cb0b3f968d7f0349a2fe64bb2fc58c204802f/public/video-ui/src/components/FormFields/DatePicker.jsx#L163-L169

public/video-ui/src/components/ScheduledLaunch.tsx
```diff
@@ -0,0 +1,296 @@
...
+      setInvalidDateError('Date must be in the future!');
+    } else {
+      setInvalidDateError(null);
+    }
+  };
+
+  const handleSelectOption = (
+    delayPublicationOption: DelayPublicationOption
+  ) => {
+    setSelectedDelayPublicationOption(delayPublicationOption);
```
bryophyta, line 66: It looks like the old component used to call `validateDate` in the `onSelectOption` callback; is this an intentional change?

public/video-ui/src/components/ScheduledLaunch.tsx
junyuanxue: Ah yes good spot! 🙏

public/video-ui/src/components/ScheduledLaunch.tsx
junyuanxue: Ah actually I think I should allow `null` to allow for the date picker to be cleared

public/video-ui/src/components/ScheduledLaunch.tsx
junyuanxue: It is indeed - previously `validateDate` was called every time because date selection was persisted when you click on `cancel`. In https://github.com/guardian/media-atom-maker/pull/1629/changes/4b910d30ea7a99be073e48dda60823ba4de80c77 I've just added a state reset function which hopefully will make everything clearer. This should ensure that when we open up the menu and select a delay option there is no state persisted from previous unsaved date selection

bryophyta (approved):
nice! the code is a lot easier to reason about, and from testing locally I can see that it fixes a number of problems in the rendered UI 🥳

PR #1631, opened 2026-07-13
13 07 update TagFieldValue component to ts

AlexAscherson:
## What does this change?

Adds types to this component.  There may be an argument for refactoring this component as its quite messy.

## Accessibility


public/video-ui/src/components/Tags/TagFieldValue.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </span>
+      );
+    } else if (typeof value !== 'string' && value.webTitle) {
+      // In a YouTube Furniture tab with non-launched changes, the `keyword` field
+      // passes an object with webTitle to this component
+      const titleToRender = value.webTitle.concat(shouldAddComma ? ',' : '');
+      return titleToRender;
+    }
+
+    if (typeof value !== 'string') {
```
bryophyta, line 52: this looks like a change in behaviour from the previous implementation, is that correct?

public/video-ui/src/components/Tags/TagFieldValue.tsx
```diff
@@ -0,0 +1,62 @@
...
+type TagFieldValueItem =
+  | string
+  | {
+      id: string;
+      detailedTitle?: string;
+      webTitle?: string;
+    };
+
+type Props = {
+  tagType?: string;
```
bryophyta: is it possible for us to type `tagType` more narrowly here? It looks like `youtube` is one possible value, judging from line 24 -- is there a fixed set of other values it could be, e.g. `tagType?: 'youtube' | 'tone' | ...`?

AlexAscherson: I have updated the type with the values I could find

public/video-ui/src/components/Tags/TagFieldValue.tsx
bryophyta: We discussed this offline and it looks like there is a change of behaviour, but it looks like a desirable one; before, if the value was `{ id: number }`, which it could be according to the types, then it would return `" [objectObject]`. Returning the number (i.e. `value.id`) as this branch does seems a bit more helpful in this context 👍

public/video-ui/src/components/Tags/TagFieldValue.tsx
```diff
@@ -0,0 +1,72 @@
...
+      const titleToRender = value.webTitle.concat(shouldAddComma ? ',' : '');
+      return titleToRender;
+    }
+
+    if (typeof value !== 'string') {
+      return value.id;
+    }
+
+    if (index === 0 || value === ',') {
+      return value;
```
lindseydew, line 57: These cases also feel like it would be worth adding unit tests to document the desired behaviour, but not a blocker 👍

lindseydew (approved):
LGTM 👍 ✨

PR #1634, opened 2026-07-13
Update VideoIcon to ts

AlexAscherson:
## What does this change?

Converts file to ts
Adds IconProps type
Adds return types for react nodes
Adds returns for null where previously would return undefined

## Accessibility


public/video-ui/src/components/Icon.tsx
```diff
@@ -94,8 +103,8 @@ export class SubtitlesIcon extends React.Component {
   }
 }
 
-export default class Icon extends React.Component {
-  renderText() {
+export default class Icon extends React.Component<IconProps> {
```
hk-15, line 106: I wonder if it's worth converting this (and the other class components in this file) to a function component? Might help simplify code a bit and is the recommended pattern now (TIL while reading the React docs!)

hk-15 (approved):

public/video-ui/src/components/Icon.tsx
AlexAscherson: Done! 

I put it in a separate pr here:
https://github.com/guardian/media-atom-maker/compare/update-video-icon-to-ts-part-2?expand=1

PR #1638, opened 2026-07-14
Convert GridImage to Typescript

AlexAscherson:
## What does this change?

Another small one, imports image type from videos API and defines a prop type that uses it

## Accessibility


public/video-ui/src/components/GridImage/GridImage.tsx
```diff
@@ -1,12 +1,12 @@
...
 
-export default class GridImage extends React.Component {
-  static propTypes = {
-    image: PropTypes.object.isRequired
-  };
+type Props = {
+  image: Image;
+};
 
+export default class GridImage extends React.Component<Props> {
```
bryophyta: could be nice to change this to a named export and functional component while we're here, but not blocking -- can always be done separately in a follow-up

bryophyta (approved):
I've left a suggestion, but not blocking, lgtm!

public/video-ui/src/components/GridImage/GridImage.tsx
AlexAscherson: Done!

PR #1639, opened 2026-07-14
Convert VideoSearch Component to TS

AlexAscherson:
Another quick one, brings in Search type and adds a local prop type

## What does this change?
Another quick change, import search type and use it as a prop

Updates the onSearch event to use HTMLImputElement type

## Accessibility


public/video-ui/src/components/VideoSearch/VideoSearch.tsx
```diff
@@ -1,8 +1,15 @@
...
 
-export default class VideoSearch extends React.Component {
-  onSearch = e => {
+type Props = {
+  saving: boolean;
+  search: Search;
+  updateSearchTerm: (term: string) => void;
+};
+
+export default class VideoSearch extends React.Component<Props> {
```
bryophyta: Looks like there's only one call site for this component, in Header.jsx, and it's spreading a whole load of props that this component doesn't need. Shall we update the call site so we're only passing in what we need according to the new types?

AlexAscherson: Done

bryophyta (approved):
🥳

PR #1640, opened 2026-07-14
Convert WorkflowForm to typescript

AlexAscherson:
## What does this change?

Imports video type
Defines select value type
Defines workflow status form data type

## Accessibility


public/video-ui/src/components/Workflow/WorkflowForm.tsx
```diff
@@ -1,21 +1,36 @@
...
+
+type Props = {
+  editable: boolean;
+  video: Video;
+  workflowSections: SelectValue[];
+  workflowStatuses: SelectValue[];
+  workflowPriorities: SelectValue[];
+  workflowStatus: WorkflowStatusFormData;
+  workflowProductionOffices: SelectValue[];
+  updateData: (data: WorkflowStatusFormData) => Promise<unknown> | unknown;
```
bryophyta: `Promise<unknown> | unknown` is quite loose.. Can't we narrow the return type to `Promise<WorkflowStatusFormData>`? https://github.com/guardian/media-atom-maker/blob/f56eace94dcbdf4659d744ce540cc936589373aa/public/video-ui/src/components/Workflow/Workflow.jsx#L35-L38

public/video-ui/src/components/Workflow/WorkflowForm.tsx
```diff
@@ -1,21 +1,36 @@
...
+  note?: string;
+  status?: string;
+  priority?: string | number;
+  [key: string]: unknown;
+};
+
+type Props = {
+  editable: boolean;
+  video: Video;
+  workflowSections: SelectValue[];
```
bryophyta: The definition of `SelectValue` doesn't seem to match the one we've got in the store for `WorkflowState` (which seems to be where the values for these props ultimately come from): https://github.com/guardian/media-atom-maker/blob/f56eace94dcbdf4659d744ce540cc936589373aa/public/video-ui/src/slices/workflow.ts#L12-L17

I don't know which one is incorrect but I think it would be good to resolve it rather than creating an inconsistency between the two types.

public/video-ui/src/components/Workflow/WorkflowForm.tsx
```diff
@@ -1,21 +1,36 @@
...
-    workflowStatus: PropTypes.object.isRequired,
-    workflowProductionOffices: PropTypes.array.isRequired,
-    updateData: PropTypes.func.isRequired
-  };
+type SelectValue = {
+  id: string | number;
+  title: string;
+};
 
+type WorkflowStatusFormData = {
```
bryophyta: It looks to me as though we could draw on the `WorkflowState` type from the store definition here instead of needing to define a new type?

public/video-ui/src/components/Workflow/WorkflowForm.tsx
AlexAscherson: Done

public/video-ui/src/components/Workflow/WorkflowForm.tsx
```diff
@@ -1,22 +1,36 @@
...
+  editable: boolean;
+  video: Video;
+  workflowSections: SelectValue[];
+  workflowStatuses: SelectValue[];
+  workflowPriorities: SelectValue[];
+  workflowStatus: WorkflowStatusFormData;
+  workflowProductionOffices: SelectValue[];
+  updateData: (
+    data: WorkflowStatusFormData
+  ) => Promise<WorkflowStatusFormData> | unknown;
```
AlexAscherson: remove this unknown since we have typed it  and undermines the typing

public/video-ui/src/components/Workflow/WorkflowForm.tsx
AlexAscherson: Change to the correct select type and work out which of the title / name field is correct
- Comes out of the api as 'name'

public/video-ui/src/components/Workflow/WorkflowForm.tsx
AlexAscherson: Done

lindseydew (approved):
LGTM 👍  Nice work ✨

PR #1642, opened 2026-07-15
Add year field to IconikCommission table

junyuanxue:
## What does this change?

[Asana for context](https://app.asana.com/1/1210045093164357/project/1213625402425263/task/1213633053967633?focus=true)

Add an optional `year` field to `IconikCommission` table, this will come from the `yearName` in the event stream payload when an Iconik project is created.

There are some thoughts around: 1) whether we'd want to keep this optional; 2) how strictly we want to parse and store this - and I've left them as comments below so we can 🧵 

I have tested by calling the endpoint to `/api/iconik/projects` locally, and can see the `year` field popping up in the new rows under `IconikCommission` table:

Screenshot 2026-07-16 at 16 14 20

---
  - https://app.asana.com/0/0/1216599896468928

### DIFF
--- a/common/src/main/scala/com/gu/media/iconik/IconikModels.scala
+++ b/common/src/main/scala/com/gu/media/iconik/IconikModels.scala
@@ -44,7 +44,8 @@ object IconikWorkingGroup {
 case class IconikCommission(
     workingGroupId: String,
     id: String,
-    title: String
+    title: String,
+    year: Option[String] = None
 ) extends IconikItemWithParentId {
   val parentId: String = workingGroupId
 }
@@ -60,7 +61,8 @@ object IconikCommission {
     IconikCommission(
       id = req.commissionId,
       title = req.commissionTitle,
-      workingGroupId = req.workingGroupId
+      workingGroupId = req.workingGroupId,
+      year = req.yearName
     )
   }
 }
@@ -126,7 +128,8 @@ case class IconikUpsertRequest(
     commissionTitle: String,
     workingGroupId: String,
     workingGroupTitle: String,
-    masterPlaceholderId: Option[String]
+    masterPlaceholderId: Option[String],
+    yearName: Option[String]
 //    productionOffice: String, // don't know if we need this in Iconik world?
 )
 
--- a/pluto-message-ingestion/src/process-record.test.ts
+++ b/pluto-message-ingestion/src/process-record.test.ts
@@ -122,7 +122,8 @@ describe('processRecord', () => {
       commissionTitle: 'Iconik Commission',
       workingGroupId: 'wg-1',
       workingGroupTitle: 'Working Group 1',
-      masterPlaceholderId: 'mp-1'
+      masterPlaceholderId: 'mp-1',
+      yearName: '2026'
     };
 
     const record = {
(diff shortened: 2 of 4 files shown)


common/src/main/scala/com/gu/media/iconik/IconikModels.scala
```diff
@@ -44,12 +47,34 @@ object IconikWorkingGroup {
 case class IconikCommission(
     workingGroupId: String,
     id: String,
-    title: String
+    title: String,
+    year: Option[Year] = None,
```
junyuanxue: I assume even after backfilling, we might want to keep this optional if this mainly comes from file organisation? Happy to be stricter and make it required though

common/src/main/scala/com/gu/media/iconik/IconikModels.scala
```diff
@@ -44,12 +47,34 @@ object IconikWorkingGroup {
...
+      }
+    },
+    Writes[Year](year => JsString(year.toString))
+  )
+
+  implicit val yearDynamoFormat: DynamoFormat[Year] =
+    DynamoFormat.coercedXmap[Year, Int, IllegalArgumentException](
+      Year.of,
+      _.getValue
+    )
```
junyuanxue: I found storing this as a `yearName` string a bit weird and thought I might be stricter about typing at the domain level at least. Though not strongly tied to this, as mentioned above, this seems to be mainly for file organisation purpose, and happy to also just store this as a `yearName` string to be consistent with where it comes from and not confuse people

bryophyta: I keep going back and forth on this, too! I lean towards wanting stricter typing here too. But as you said, we don't have many guarantees about what the value will look like, and I don't think we need to do any number- or year-specific operations with it, so I think on balance I'd opt for storing it as a string?

junyuanxue: Thanks for the sense-check. In that case perhaps we store as a `year` string (still find 'yearName' really weird tbh 😂 )

bryophyta: > In that case perhaps we store as a year string (still find 'yearName' really weird tbh 😂 )

sounds good to me!

bryophyta (approved):
nice, this looks great!

I think we need to remove entries from the DB which were added when the year was being modelled as a number though:

```
{
  "@timestamp": "2026-07-20T12:50:52.395399+01:00",
  "@version": "1",
  "message": "failed to get iconik commissions for working group ID a0640e30-e7bf-11f0-acef-3640eaca9ef5: InvalidPropertiesError(NonEmptyList((year,NoPropertyOfType(S,DynNum(2026)))))",
  "logger_name": "controllers.IconikController",
  "thread_name": "application-pekko.actor.default-dispatcher-7",
  "level": "ERROR",
  "level_value": 40000,
  "application.home": "/Users/pete_faulconbridge/code/media-atom-maker"
}
```

There shouldn't be any in the PROD db though, so this doesn't block merging afaict.

lindseydew (approved):
LGTM 👍

PR #1646, opened 2026-07-16
Add clientside filtering by commission year

junyuanxue:
## What does this change?
Now that we've added `year` field on `IconikCommission`, we can add clientside support for filtering commissions by year:

* Given that in the current implementation, we're always returning all commissions under a selected working group, and barring one working group that has ~200 commissions, most don't have a large amount, so I've opted for clientside filtering as payloads are reasonably small and for simplicity.
* This adds a simple select for commission year under a working group, in descending order, along with a 'View all' option
* If a project is already linked to a video, to show the correct commission year, we need to make an extra request to fetch the existing commission.

commission year

---
  - https://app.asana.com/0/0/1216599896468934

### DIFF
--- a/public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
+++ b/public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
@@ -38,15 +38,21 @@ export const IconikProjectPicker = ({ video }: Props) => {
   const [workingGroup, setWorkingGroup] = React.useState<string | undefined>(
     () => video.iconikData?.workingGroupId
   );
-  const [commissionYear, setCommissionYear] = React.useState<
+
+  const [selectedCommissionYear, setSelectedCommissionYear] = React.useState<
     string | undefined
-  >(existingCommission?.year ?? commissionYearOptions[0]?.id);
+  >();
   const [commission, setCommission] = React.useState<string | undefined>(
     () => video.iconikData?.commissionId
   );
   const [project, setProject] = React.useState<string | undefined>(
     () => video.iconikData?.projectId
   );
+
+  const commissionYear =
+    selectedCommissionYear ??
+    existingCommission?.year ??
+    commissionYearOptions[0]?.id;
 
   const filteredCommissions =
     commissionYear && commissionYear !== ALL_COMMISSION_YEARS_OPTION
@@ -59,12 +65,15 @@ export const IconikProjectPicker = ({ video }: Props) => {
     } else {
       dispatch(resetCommissions());
     }
+  }, [dispatch, workingGroup]);
+
+  useEffect(() => {
     if (commission) {
       dispatch(fetchIconikProjects(commission));
     } else {
       dispatch(resetProjects());
     }
-  }, [commission, dispatch, workingGroup]);
+  }, [dispatch, commission]);
 
   const hasBeenEdited =
     video.iconikData?.workingGroupId !== workingGroup ||
@@ -79,14 +88,10 @@ export const IconikProjectPicker = ({ video }: Props) => {
     [dispatch, video]
   );
 
-  useEffect(() => {
-    setCommissionYear(commissionYearOptions[0]?.id);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [commissions]);
-
   const onWorkingGroupChange = useCallback(
     (selectedWorkingGroupId: string | undefined) => {
       setWorkingGroup(selectedWorkingGroupId);
+      setSelectedCommissionYear(undefined);
       setCommission(undefined);
       setProject(undefined);
     },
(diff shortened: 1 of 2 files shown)


bryophyta (commented):
Filtering by year seems to work as expected 👍 

But the logic for loading and displaying an existing year doesn't seem to be working for me. Two points that I noticed:

1. The `commissionYear` state is initialised using the value of `existingCommission?.year`, but the existingCommission isn't fetched until after the state is initialised, so it seems to always get initialised as `undefined`.
2. When you change the Working Group, the Commission and Project dropdown values reset (and changing Commission resets Project, too), but we aren't resetting the Year dropdown. I _think_ that we probably want to reset it following the same logic as the others? 

From what I can see, I'm not sure we actually need the extra fetch in (1), because we should already have info on the existing commission via the commission-list call that we do already:

Image

public/video-ui/src/services/IconikApi.ts
```diff
@@ -37,6 +38,16 @@ export function getIconikCommissionsForWorkingGroup({
   });
 }
 
+export function getIconikCommission({
```
bryophyta: this function is imported into `iconik.ts` but doesn't seem to be used anymore so I think it can be removed?

public/video-ui/src/slices/iconik.ts
```diff
@@ -1,5 +1,6 @@
 import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
 import {
+  getIconikCommission,
```
bryophyta: It doesn't look like we're using `getIconikCommission` or `existingCommission` from this store slice anymore so I think we can drop the changes to this file?

public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
```diff
@@ -66,6 +79,10 @@ export const IconikProjectPicker = ({ video }: Props) => {
     [dispatch, video]
   );
 
+  useEffect(() => {
+    setCommissionYear(commissionYearOptions[0]?.id);
+  }, [commissions]);
```
bryophyta: given that `commissionYearOptions` is a pure function of `commissions`, which is already in the dependencies array, it doesn't feel strictly necessary to add `commissionYearOptions` to the dependencies, but might be worth doing just to remove the linter warning?

bryophyta: although, having thought about it a bit more, I guess adding `commissionYearOptions` to the dependencies will mean we run the useEffect unnecessarily as it will be assigned a new array on each render, even if the contents are the same. So could wrap in a `useMemo` or just ignore/suppress the warning?

public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
```diff
@@ -30,16 +30,29 @@ export const IconikProjectPicker = ({ video }: Props) => {
...
+  >(existingCommission?.year ?? commissionYearOptions[0]?.id);
   const [commission, setCommission] = React.useState<string | undefined>(
     () => video.iconikData?.commissionId
   );
   const [project, setProject] = React.useState<string | undefined>(
     () => video.iconikData?.projectId
   );
 
+  const filteredCommissions =
+    commissionYear && commissionYear !== 'all-commission-years'
```
bryophyta:
```suggestion
    commissionYear && commissionYear !== ALL_COMMISSION_YEARS_OPTION
```

Should we use the shared constant here? (though it's currently defined below this in the code)

public/video-ui/src/services/IconikApi.ts
junyuanxue: Ah good catch!

public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
junyuanxue: I think a memo would be an overkill here, so I'll ignore the warning!

public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
junyuanxue: Ah absolutely!

public/video-ui/src/components/Iconik/IconikProjectPicker.tsx
bryophyta: that's fair! 

although I think I just spotted a slightly different issue with this useEffect..

if you select a year value that's not the top of the list, then this unsets `commission`, which triggers the `useEffect` on line 56, which triggers `dispatch(fetchIconikCommissions(workingGroup))` which mutates `commissions` and then triggers _this_ `useEffect`, setting the year to `commissionYearOptions[0]?.id`.

e.g. if I manually change the selection from `2026` to `2025` it will reset to `2026`

junyuanxue: Ah good spot! I have [de-coupled things](https://github.com/guardian/media-atom-maker/pull/1646/changes/3e328d9b527072165cd741bbc4cd30ca85435799) from the previous one big `useEffect` where changing commission would cause commissions to be refetched which doesn't seem right lol

bryophyta (approved):
nice! this looks good to me, and great that it removes some unnecessary API calls in the process too 🥳

PR #1651, opened 2026-07-20
Convert remove tag duplicates to TS

AlexAscherson:
## What does this change?

Adds stub tag type and param types

## Accessibility


public/video-ui/src/util/removeTagDuplicates.ts
```diff
@@ -0,0 +1,8 @@
+type Tag = { id: string };
+
+export default function removeTagDuplicates<T extends Tag>(
+  tag: Tag,
+  tagValue: T[]
```
hk-15, line 5: tagValue is typed more narrowly in removeStringTagDuplicates.ts. Should they be the same?

public/video-ui/src/util/removeTagDuplicates.ts
```diff
@@ -0,0 +1,8 @@
+type Tag = { id: string };
+
+export default function removeTagDuplicates<T extends Tag>(
```
hk-15: Named export?

hk-15 (approved):
Just two small Qs, but I don't want to block merging :)

public/video-ui/src/util/removeTagDuplicates.ts
AlexAscherson: Done

public/video-ui/src/util/removeTagDuplicates.ts
AlexAscherson: So i think the issue here is that removeTagDuplicates / tagValue can get string values passed in if the CAPI lookup fails.

This happens in TagPicker.jsx on line61 in the componentDidMount function

PR #1656, opened 2026-07-21
Centralise server date conversion in VideosApi

davidfurey:
## What does this change?

The API returns the atom's `contentChangeDetails` date fields (`scheduledLaunch`, `embargo`, `expiry`) as ISO 8601 strings, but the client and downstream consumers expect UNIX millisecond timestamps (integers). Until now that conversion only happened in the `getVideo` action, so every other endpoint that returns an atom — `createVideo`, `saveVideo`, `publishVideo`, `fetchPublishedVideo`, the asset mutations (`createAsset`, `revertAsset`, `deleteAsset`, `deleteAssetList`), `resetDurationFromActive` and the `fetchVideos` list — returned raw string dates even though the TypeScript types claimed they were `number`. This is a latent runtime bug.

This change moves the conversion into the `VideosApi` service layer so it is applied consistently to every response:

- Introduces `ContentChangeDetailsBuilder<T>` so the server shape (string dates) and client shape (number dates) share a single definition; `ContentChangeDetails` is now `ContentChangeDetailsBuilder<number>`.
- Adds `VideoServer` / `MediaAtomSummaryServer` types describing the raw API payloads.
- Adds a generic `clientVideoFromServerVideo` helper that converts the date strings to millis, applied via `.then()` on every request that returns a video or atom summary (including the list from `fetchVideos`).
- Guards against unparseable dates (`NaN` → `undefined`) and preserves the epoch value `0`.
- Removes the now-redundant conversion and the unused `moment` import from the `getVideo` action.
- Updates test fixtures to use string dates matching the server shape.

- [ ] Deployed to `CODE` and confirmed launch/embargo/expiry dates render correctly on the video detail page and in the video list, and that scheduling/publishing still behaves as expected.

- No functional/visual change is expected — dates should continue to display and sort correctly across the detail view and list view, which previously worked only because of the single conversion in `getVideo`.
- The main win is correctness and type safety: endpoints other than `getVideo` now return the millisecond timestamps their types promise, removing a class of "string where a number is expected" bugs. This is best measured by the absence of date-related rendering/sorting issues and by the types now matching runtime behaviour.

- The behavioural surface is limited to how atom date fields are shaped before reaching the UI/store. The main risk is a subtle change for endpoints that previously returned string dates: any consumer that was (incorrectly) relying on the string value will now receive a number. This is the intended correction, but worth watching after deploy.
- Edge cases are handled explicitly: unparseable dates become `undefined` rather than `NaN`, and the epoch value `0` is preserved rather than dropped.


bryophyta (approved):
looks great to me! I checked on CODE and the issue flagged in the earlier comment is fixed now 👍

public/video-ui/src/util/parseGridMetadata.ts
```diff
@@ -117,31 +117,43 @@ export function parseComposerDataFromImage(
...
+      return undefined;
+    }
   }
 
-  function getComposerMasterAsset(asset: ParsedAsset): ComposerAsset {
-    const composerAsset = getComposerAsset(asset);
-    composerAsset.isMaster = 'true';
-    return composerAsset;
+  function getComposerMasterAsset(
+    asset?: ImageAsset
```
junyuanxue: Nit: if `asset` doesn't exist, then does it make sense to call this function in the first place? I feel like this check should happen at callsite

junyuanxue (commented):
Image

Thanks for this! While testing I've spotted an issue around the video status:
* A 'draft' video appears to be 'published' on the video page, with the publish button disabled 
* I keep getting 'Could not get published video' error on the video page

I'm wondering if some of these states are driven by dates which doesn't feel quite right 🤔

public/video-ui/src/util/parseGridMetadata.ts
davidfurey: Thanks, fixed.

davidfurey:
> Image
> Thanks for this! While testing I've spotted an issue around the video status:
> 
> * A 'draft' video appears to be 'published' on the video page, with the publish button disabled
> * I keep getting 'Could not get published video' error on the video page
> 
> I'm wondering if some of these states are driven by dates which doesn't feel quite right 🤔

Thanks for spotting this. I've fixed. It was caused by `clientVideoFromServerVideo` failing when the server responds with an empty (`{}`) published video.

junyuanxue (approved):
Thanks for fixing! LGTM :)

PR #1658, opened 2026-07-21
Add missing Iconik data warning to the video publish bar

hk-15:
## What does this change?

Ideally a Video Atom is linked to an Iconik project so we know where it comes from. Currently, a video can be published without this link, and a user does get an email if they don't set an Iconik project on a Video Atom, but it'd be good if we surface a warning in the UI too to prompt them to add this information.

This PR adds a warning message to UI to alert a user to add Iconik data. In a demo of the Iconik and MAM workflow, the Multimedia team mentioned that adding Iconik data is not mandatory because it is preferable not to block publishing. So this warning message is not linked to the existing errors that block publishing.

Tested locally on this branch.

Fewer videos have no Iconik data associated with them. 

## Images
Screenshot 2026-07-28 at 10 29 50

## Accessibility

### DIFF
--- a/public/video-ui/src/components/ReactApp.tsx
+++ b/public/video-ui/src/components/ReactApp.tsx
@@ -1,12 +1,10 @@
 import React, { useEffect, useState } from 'react';
 
 import Header from './Header';
-import { useDispatch, useSelector, useStore } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { AppDispatch, RootState } from '../util/setupStore';
 import {
-  selectIsActivatingAssetNumber,
   selectIsPublishing,
-  selectIsSaving,
   selectPublishedVideo,
   selectVideo
 } from '../slices/video';
@@ -28,6 +26,7 @@ import { getPublishedVideo } from '../actions/VideoActions/getPublishedVideo';
 import { getUploads } from '../slices/uploads';
 import { createVideo } from '../actions/VideoActions/createVideo';
 import { fetchCropOptions } from '../slices/gridMetadata';
+import { hasIconikOrPlutoProject } from '../util/hasIconikOrPlutoProject';
 
 export const ReactApp = (
   props: React.PropsWithChildren<{
@@ -62,6 +61,15 @@ export const ReactApp = (
   const video = useSelector(selectVideo);
   const publishedVideo = useSelector(selectPublishedVideo);
   const isPublishing = useSelector(selectIsPublishing);
+  const currentPath = String(props.location.pathname);
+  const isVideoDetailPage =
+    Boolean(props.params.id) && !currentPath.endsWith('/upload');
+  const hasLoadedCurrentVideo =
+    Boolean(props.params.id) && video.id === props.params.id;
+  const showMissingIconikDataWarning =
+    isVideoDetailPage &&
+    hasLoadedCurrentVideo &&
+    !hasIconikOrPlutoProject(video);
 
   useEffect(() => {
     if (
@@ -148,6 +156,11 @@ export const ReactApp = (
           className={'error-bar--warning error-bar--animate'}
         >
           {store.error.warningMessage}
+        </div>
+      ) : null}
+      {showMissingIconikDataWarning ? (
+        <div className={'error-bar--warning error-bar--animate'}>
+          Click &quot;Edit Assets&quot; to add missing Iconik data
         </div>
       ) : null}
       <div>{props.children}</div>
--- a/public/video-ui/src/components/ReactApp.tsx
+++ b/public/video-ui/src/components/ReactApp.tsx
@@ -26,7 +26,6 @@ import { getPublishedVideo } from '../actions/VideoActions/getPublishedVideo';
 import { getUploads } from '../slices/uploads';
 import { createVideo } from '../actions/VideoActions/createVideo';
 import { fetchCropOptions } from '../slices/gridMetadata';
-import { hasIconikOrPlutoProject } from '../util/hasIconikOrPlutoProject';
 
 export const ReactApp = (
   props: React.PropsWithChildren<{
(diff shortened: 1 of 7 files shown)


public/video-ui/src/components/VideoPublishBar/VideoPublishBar.jsx
```diff
@@ -101,6 +103,20 @@ export default class VideoPublishBar extends React.Component {
...
 
+  renderIconikProjectMissingWarning() {
+    const videoHasIconikProject = hasIconikProject(this.props.video);
+
+    if (videoHasIconikProject) {
+      return null;
+    } else {
+      return (
+        <div className="header__fields__missing__warning">
+          Click &quot;Edit Assets&quot; to add missing Iconik data
```
hk-15: In the Iconik tab, when data is missing it instructs the user that `Associations can be added in the "upload" view.` but I was unsure about using that wording since the button to be clicked is `Edit Assets`. Most users probably do know that button takes them to the 'upload' view, so perhaps it doesn't matter either way.

public/video-ui/src/util/hasIconikProject.ts
```diff
@@ -0,0 +1,5 @@
+import { Video } from '../services/VideosApi';
+
+export const hasIconikProject = (video: Video): boolean => {
```
junyuanxue: I think this might need to be a `hasIconikOrPlutoProject` check as the old ones might be linked to Pluto

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.jsx
junyuanxue: I agree the wording is quite unclear! I definitely did not know clicking on 'Edit asset' would take me to the 'upload' view initially 😂  but I think that's a wider problem to sort 😅

public/video-ui/src/components/VideoPublishBar/VideoPublishBar.jsx
```diff
@@ -81,6 +82,7 @@ export default class VideoPublishBar extends React.Component {
           {this.renderPublishButtonText()}
         </button>
         {this.renderRequiredFieldsMissingWarning()}
+        {this.renderIconikProjectMissingWarning()}
```
junyuanxue: I was thinking about using the error banner space to surface this warning initially for post-publishing (which I should have added to the Asana ticket - my bad! 🙈) But I think adding this non-blocking warning pre-publishing is definitely very useful! After the video is published, I do find it a bit weird to find this text next to a disabled 'Published' button and wonder if a banner might work better. I'm not too attached to that idea though. Would be keen to hear your thoughts! 

Image

hk-15: Yes, this is a much better way to surface the warning. Having a differently styled error message before and after publishing seems like too much (at least, it felt weird as a user going through the workflow), so I've opted just to have the banner pre- and post-publishing. Amended in [a529444](https://github.com/guardian/media-atom-maker/pull/1658/commits/a52944412617b588d5ba132e8837c0fba9e2d56c)

public/video-ui/src/components/ReactApp.tsx
```diff
@@ -1,12 +1,10 @@
 import React, { useEffect, useState } from 'react';
 
 import Header from './Header';
-import { useDispatch, useSelector, useStore } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { AppDispatch, RootState } from '../util/setupStore';
 import {
-  selectIsActivatingAssetNumber,
   selectIsPublishing,
-  selectIsSaving,
```
hk-15, line 9: Small clean-ups of unused imports while I'm here

public/video-ui/src/components/ReactApp.tsx
```diff
@@ -150,6 +158,11 @@ export const ReactApp = (
           {store.error.warningMessage}
         </div>
       ) : null}
+      {showMissingIconikDataWarning ? (
+        <div className={'error-bar--warning error-bar--animate'}>
+          Click &quot;Edit Assets&quot; to add missing Iconik data
+        </div>
+      ) : null}
```
junyuanxue: This works, but the downside is when there are already a warning and an error, you'll end up with 3 banners 🙈 I think the global error store is set up so there will only ever be one global error and one global warning showing, so we should use that mechanism to surface this banner.

There is a `showError` action in error slices, and a `clearError` one that clears both the error and warning which is confusing lol. We probably want to add a `showWarning` action there and make the clear action a bit cleaner. Then we could call this in the `Video` component when it's mounted, so you don't need the path check which you've done here.

Happy to chat about this or pair on it if I wasn't clear enough!

public/video-ui/src/util/hasIconikOrPlutoProject.ts
```diff
@@ -0,0 +1,7 @@
+import { Video } from '../services/VideosApi';
+
+export const hasIconikOrPlutoProject = (video: Video): boolean => {
+  return (
+    Boolean(video.iconikData?.projectId) || Boolean(video.plutoData?.projectId)
```
junyuanxue: Super nit - you can do ` Boolean(video.iconikData?.projectId || video.plutoData?.projectId)` to simplify a little bit

public/video-ui/src/pages/Video/index.jsx
```diff
@@ -46,6 +47,28 @@ class VideoDisplay extends React.Component {
...
+        : undefined;
+
+    const prevVideo =
+      prevProps.video && prevProps.params.id === prevProps.video.id
+        ? prevProps.video
+        : undefined;
+
+    if (video === prevVideo || !video) {
+      return;
+    }
```
junyuanxue: I think this will need to be narrowed down to comparison between `hasIconikOrPlutoProject(video)` and `hasIconikOrPlutoProject(prevVideo)`, otherwise every time the video updates, the warning banner will be re-triggered and you'll see the animation flashing 😅

hk-15: Yes, good point! I think we'd have to change `hasIconikOrPlutoProject` to return `true` when passed an undefined parameter? (Since either could be undefined.)

hk-15: Actually let me just change the flow of logic to keep `hasIconikOrPlutoProject` as is

junyuanxue (approved):
Looks lovely! 💛

PR #1659, opened 2026-07-22
Backfill script for IconikCommission year

junyuanxue:
## What does this change?

Add a Scala script that backfills IconikCommission table items without a `year` field

I have run this backfill against CODE and all seems to be okay. To test this:

* Ensure you have AWS credentials set
* Under repo root, run `sbt "app/runMain backfill.IconikCommissionYearBackfill --stage=CODE --dry-run"`. You should see:
Screenshot 2026-07-22 at 17 19 45
* If you dry-run in PROD: `sbt "app/runMain backfill.IconikCommissionYearBackfill --stage=PROD --dry-run"`, you should see:
Screenshot 2026-07-22 at 17 18 41

---
  - https://app.asana.com/0/0/1216599896468935

### DIFF
--- a/app/backfill/IconikCommissionYearBackfill.scala
+++ b/app/backfill/IconikCommissionYearBackfill.scala
@@ -1,0 +1,117 @@
+package backfill
+
+import com.gu.media.aws.{AwsCredentials, CredentialsForAws}
+import com.gu.media.iconik.{
+  IconikCommission,
+  IconikDynamoStoreWithParentIndex,
+  IndexConfig
+}
+import com.typesafe.config.ConfigFactory
+import util.AWSConfig
+
+object IconikCommissionYearBackfill {
+
+  private val DefaultYear = "2026"
+  private val ValidStages = Set("CODE", "PROD")
+
+  private val profileName = sys.env.getOrElse("AWS_PROFILE", "media-service")
+
+  final case class CliArgs(stage: String, dryRun: Boolean)
+
+  def main(args: Array[String]): Unit = {
+    parseArgs(args.toList) match {
+      case Left(error) =>
+        Console.err.println(error)
+        printUsage()
+      case Right(cliArgs) =>
+        run(cliArgs)
+    }
+  }
+
+  private def run(cliArgs: CliArgs): Unit = {
+    val config = ConfigFactory.load()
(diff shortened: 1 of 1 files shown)


hk-15 (approved):
Looks good to me

app/backfill/IconikCommissionYearBackfill.scala
```diff
@@ -0,0 +1,117 @@
...
+        IndexConfig("working-group-index", "workingGroupId")
+      )
+
+    commissionStore.list match {
+      case Left(err) =>
+        Console.err.println(s"[error] failed to list commissions: $err")
+
+      case Right(commissions) =>
+        val missingYear = commissions.filter(_.year.isEmpty)
+
```
lindseydew, line 82: Could be helpful to log out how many records you are expecting to update

lindseydew: I see you've got it further down, so may be not very important, I'm just thinking it could be good to track how far through this is

app/backfill/IconikCommissionYearBackfill.scala
```diff
@@ -0,0 +1,117 @@
...
+        missingYear.foreach { commission =>
+          if (cliArgs.dryRun) {
+            println(
+              s"[dry-run] would update commissionId=${commission.id} year=$DefaultYear"
+            )
+          } else {
+            commissionStore.upsert(commission.copy(year = Some(DefaultYear)))
```
lindseydew: I believe in dynamo there is an option to make a projection and just update a field, rather than reinsert the whole data item. 

It might be a bit overkill in this case, but if there is any chance the record may be modified in a different way whilst the script is running, there is a slight risk of rolling back a change that has been made by overwriting the document

lindseydew (commented):
Looking great, just a suggestion about restricting the update to only a field

app/backfill/IconikCommissionYearBackfill.scala
junyuanxue: Thanks! Both good points - updated here: https://github.com/guardian/media-atom-maker/pull/1659/changes/d546f91d50187b3be44dcbf06059e64275c8d3ee

lindseydew (approved):
Looks great, very clear ✨

PR #1665, opened 2026-07-27
Update Sentry from 2020 version (raven)

AlexAscherson:
The sentry integration in MAM is currently out of date. (2025)

The raven library is deprecated, and only is only covering the JS parts of the app and currently not capturing Scala / backend errors.

This pr:

- Removes the raven library and updates the project to use sentry/browser v10
- Adds coverage for the Scala parts of the project.
- Adds a tsconfig file to allow for more modern packages since the existing one is locked to using common.js, rather than the modern ESM first resolution approach.
- Adds request metadata to Sentry reports for additional telemetry.
- Adds session replay capture for errors

An example of the new error reporting is available here:

[https://the-guardian.sentry.io/issues/7658449555/?environment=code&project=129691&query=is%3Aunresolved%20issue.category%3A%5Berror%2Coutage%5D&referrer=issue-stream
(](https://the-guardian.sentry.io/issues/7658449555/?environment=code&project=129691&query=is%3Aunresolved%20issue.category%3A%5Berror%2Coutage%5D&referrer=issue-stream)

Frontend and backend performance reporting can be found here:
https://the-guardian.sentry.io/insights/frontend/?environment=code&project=129691&statsPeriod=14d

and here:

https://the-guardian.sentry.io/insights/backend/?environment=code&project=129691&statsPeriod=14d

To Dos
- [ ] Review requirements in terms of what features we want to use.  
- [ ] Try pair down this PR / make sure we are only doing what we need to.
- [ ] Currently frontend console errors are triggering a post to sentry, but they are not showing up in the sentry error panels.  It would be good to capture these but potentially not essential at this stage. Decide which way to go on this.
- [ ] We should update the references to raven in the env configs of each app
- [ ] Github will need to be updated to point the sourcemapping token at the correct project in sentry for MAM prod when / if it is deployed there. See note 4 below.

NOTES:  

1. This setup will currently report errors from all envs except dev

2. The request metadata may potentially include some information about the user (name, email).  There is obfuscation in sentry but wanted to note in case there are any DP implications.  

3. The environmental config in prod + code refers to the sentry DSN value as ravelUrl / RAVEN_URL(?).  

This PR has updated references to ravenURL to SentryDsn but the config needs to be updated in via s3 / aws. There is a config file there where the keys can be updated.

4. This PR adds sourcemapping, to allow for accurate error / line reporting.  

This is currently enabled for code and tied to the mam-dev project in sentry. This uses a github auth token, which means that it will work as is in production, **but will report to mam-dev in sentry** rather than the correct project.

IMO, it would be best to close the mam-dev project in sentry and report to a single MAM project, using the env of the reports to distinguish errors in different envs rather than having separate dedicated sentry projects for different envs.

At present,  the git config used for Sourcemapping points at the mam-dev project, and so it will need to be updated for sourcemapping to work in production.

5. It would be good to agree on the set of sentry features we want to use, as there are other apps (eg. Grid) that are in the same position in regards to using the deprecated Raven setup, so if we know what we want I can update them as well with the same config.

This has been tested locally and in the CODE testing env.

An exception can be triggered with a curl request to:
curl -i -X OPTIONS {app_url}/api/workflow/atoms

This will cause a "Header with name Access-Control-Request-Headers not found!" exception to be reported here:

https://the-guardian.sentry.io/issues/?project=129691&query=&referrer=issue-list&sort=date&statsPeriod=14d

The CODE env currently reports to the media-atom-maker-dev project in sentry.

We should see errors reported in sentry + coverage of scala bits

## Accessibility


conf/application.conf
```diff
@@ -7,3 +7,8 @@ play.http.errorHandler = "util.RequestLogging"
 # 'http://localhost:5173' and 'ws://localhost:5173' are used in local development by Vite
 play.filters.headers.contentSecurityPolicy = "default-src 'self' 'unsafe-eval' 'unsafe-inline' data: https: wss: http://localhost:5173 ws://localhost:5173"
 play.filters.headers.frameOptions = null
+
+faultInjection {
+	enabled = false
+	enabled = ${?FAULT_INJECTION_ENABLED}
```
lindseydew: How would this be set in the CODE env? Would there need to be an env variable set in the CDK? 

If this is for testing the sentry config in general, perhaps it could be scoped to the `dev` env rather than making it a config variable?

public/video-ui/src/app.tsx
```diff
@@ -14,16 +14,12 @@ import '../styles/main.scss';
...
-    dsn: ravenUrl,
-    environment: stage,
-    attachStacktrace: true
-  });
+const { stage, ravenUrl, faultInjectionEnabled } = getAppConfig();
+const sentryEnvironment = stage.toLowerCase();
+
+// publish uncaught errors to sentry.io
+if (sentryEnvironment === 'code' || sentryEnvironment === 'prod') {
+  Sentry.init({ dsn: ravenUrl, environment: sentryEnvironment });
```
lindseydew: Looks like previously we were sending the upper case version of the env to sentry? But I'm assuming this won't really matter for filtering etc?

lindseydew (commented):
Looking great, thanks for picking this up. 

Just a question about how the switch would work in other environments.

public/video-ui/src/app.tsx
AlexAscherson: No it does!  Let me check if this is still at thing

AlexAscherson:
Tidied this up and tested this on Code today

conf/application.conf
AlexAscherson: I have moved the Sentry enablement config + settings for trace sampling and anything else with cost implications to SentryConfig.scala.  These now set defaults that can be overridden by env vars to allow for config changes without a code change.

From what I understand this doesn't require a CDK change, but I might be missing something.

AlexAscherson: So at present, sentry will:

Be enabled in CODE or Prod by default
Send traces from code or prod at 1.0 and 0.1 rates respectively
Only send replays on error in both envs

vite.config.ts
```diff
@@ -1,13 +1,45 @@
...
+    // including this unconditionally makes `yarn client-dev` reach out to
+    // Sentry on every start.
+    ...(command === 'build'
+      ? [
+          sentryVitePlugin({
+            org: 'the-guardian',
+            // Single project for all stages, matching the DSN in the
+            // [STAGE].public.conf files; stages are separated by `environment`
+            // rather than by project. Overridable for one-off builds.
+            project: process.env.SENTRY_PROJECT ?? 'media-atom-maker',
```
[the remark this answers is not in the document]
AlexAscherson: We have fixed this already

andrew-nowak (approved):

PR #1669, opened 2026-07-29
Add loading state when uploading

hk-15:
## What does this change?

After clicking on 'upload' / 'upload to YouTube' for a new asset, there's a lag before a new asset container appears. It'd be good to have an interim loading state to give more immediate user feedback.

This PR introduces changes two changes to indicate to the user that the upload is happening: the upload buttons are disabled straight after click and a placeholder asset container appears instantly before being populated with the real data when that becomes available.

This PR also removes the progress bar from the header. The progress bar was appearing in the UI before but only as a light grey (empty) progress bar that disappeared when the asset container appeared. After these changes, due to reordering the steps in the s3Upload slice, it was 'working' by showing as a yellow 100% bar very briefly before disappearing again. Since it doesn't seem necessary any more, removing it seemed reasonable. Happy to discuss further.

Tested locally on this branch using YouTube and self-hosted workflows, with throttling.

## Images
upload

## Accessibility

### DIFF
--- a/public/video-ui/src/components/Header.jsx
+++ b/public/video-ui/src/components/Header.jsx
@@ -45,27 +45,6 @@ export default class Header extends React.Component {
     });
   };
 
-  renderProgress() {
-    if (this.props.s3Upload.status === 'uploading') {
-      // Start prompting the user about reloading the page
-      window.onbeforeunload = () => {
-        return false;
-      };
-
-      return (
-        <progress
-          className="topbar__progress"
-          max={this.props.s3Upload.total}
-          value={this.props.s3Upload.progress}
-        />
-      );
-    } else {
-      // Stop prompting the user. The upload continues server-side
-      window.onbeforeunload = undefined;
-      return false;
-    }
-  }
-
   renderHome() {
     return (
       <div className="flex-container topbar__global">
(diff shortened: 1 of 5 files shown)


public/video-ui/src/components/Header.jsx
```diff
@@ -45,27 +45,6 @@ export default class Header extends React.Component {
...
-          max={this.props.s3Upload.total}
-          value={this.props.s3Upload.progress}
-        />
-      );
-    } else {
-      // Stop prompting the user. The upload continues server-side
-      window.onbeforeunload = undefined;
-      return false;
-    }
-  }
```
junyuanxue, line 67: Nice one cleaning this up! We already have upload progress scoped to each asset anyways so this makes sense :)

public/video-ui/src/components/VideoUpload/VideoTrail.tsx
```diff
@@ -13,13 +13,15 @@ type Props = {
...
+  showPendingUpload: boolean;
 };
 
 export const VideoTrail = ({
   video,
   uploads,
   setAsset,
-  activatingAssetNumber
+  activatingAssetNumber,
+  showPendingUpload
```
junyuanxue: Perhaps we could rename this to `hasPendingUpload` instead, as it takes the place of an upload that'll eventually pop up?

junyuanxue (approved):
Nice one! I think this looks great and is a lovely UX improvement 😍 Just a wee naming suggestion

PR #1678, opened 2026-08-13
Enable strictNullChecks and fix resulting type errors

davidfurey:
> [!NOTE]  
> This PR was opened by Claude Opus 4.8 based on an interactive session but I have reviewed the output

## What

Enables `strictNullChecks` in `tsconfig.json` and resolves all resulting type errors.

## How

Minimal changes across the affected files. The `strictNullChecks` flag is flipped in the final commit so every intermediate commit passes `yarn typecheck`:

- **Actions** — guard against undefined asset/video in `deleteAsset`/`updateVideo`
- **DraggableTagList** — handle null `over`/`parentElement`
- **PlutoProjectPicker** — use narrowed `commissionId`
- **DatePicker** — add JSDoc prop types for nullable fields
- **VideoAsset (+ test)** — null/undefined handling; `activatingAssetNumber` optional
- **VideoTrail** — guard undefined asset; optional `activatingAssetNumber`
- **Modal** — optional chaining for `dialogRef`
- **Search** — nullable presence state
- **VideosApi** — filter undefined CAPI content; type reduce accumulators
- **Slices (s3Upload, search, video) + UploadsApi** — nullable value handling
- **Test model** — `blockAds` set to boolean

## Verification

`yarn typecheck` passes (0 errors, down from 60).

### DIFF
--- a/public/video-ui/src/services/VideosApi.ts
+++ b/public/video-ui/src/services/VideosApi.ts
@@ -173,7 +173,7 @@ function getUsages({
 }
 
 function splitUsages({ usages }: { usages: CapiContent[] }) {
-  return usages.reduce(
+  return usages.reduce<{ video: CapiContent[]; other: CapiContent[] }>(
     (all, usage) => {
       if (usage.type === 'video') {
         all.video = [...all.video, usage];
@@ -182,7 +182,7 @@ function splitUsages({ usages }: { usages: CapiContent[] }) {
       }
       return all;
     },
-    { video: [] as CapiContent[], other: [] as CapiContent[] }
+    { video: [], other: [] }
   );
 }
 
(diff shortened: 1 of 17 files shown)


hk-15 (approved):
This looks good to me

public/video-ui/src/services/VideosApi.ts
```diff
@@ -181,7 +182,7 @@ function splitUsages({ usages }: { usages: CapiContent[] }) {
       }
       return all;
     },
-    { video: [], other: [] }
+    { video: [] as CapiContent[], other: [] as CapiContent[] }
```
junyuanxue: Instead of using `as`, can we type the accumulator instead by doing `reduce<{ video: CapiContent[], other: CapiContent[] }>`

junyuanxue (commented):
Thanks for this! Just a smol comment

public/video-ui/src/services/VideosApi.ts
davidfurey: Thanks, this is nicer

PR #1680, opened 2026-08-18
Add confirmation step for activating an asset when there is already an active asset

hk-15:
## What does this change?

If a video has an active asset, this could be published and live on the site, so activating another asset could have an impact on live content. Currently, a user can activate a different asset with just one click. This PR adds a confirmation step: when there is already an active asset and the video is published, the user must click the `Activate asset` button twice, and a tooltip informs the user why this is necessary.

I have added test coverage for the new behaviour. I have also checked locally that:
- when there is no active asset, an asset can be activated with one click only
- when there is an active asset and the video's status is `published`, the `Confirm activation` step must be completed and the tooltip appears explaining the reason

It could be annoying for users who are used to having to click only once. On the other hand, it is only a small change in behaviour and helpfully guards against assets being accidentally activated and live content being affected.

## Images

Screenshot 2026-08-19 at 16 58 04

## Accessibility

-   [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/public/video-ui/src/components/ActivateButton.tsx
+++ b/public/video-ui/src/components/ActivateButton.tsx
@@ -1,0 +1,70 @@
+import React, { useEffect, useRef, useState } from 'react';
+import ReactTooltip from 'react-tooltip';
+
+type ActivateButtonProps = {
+  className: string;
+  onActivate: () => void;
+  disabled?: boolean;
+  videoHasActiveAsset: boolean;
+};
+
+export const ActivateButton = ({
+  className,
+  onActivate,
+  disabled,
+  videoHasActiveAsset
+}: ActivateButtonProps) => {
+  const [confirmActivate, setConfirmActivate] = useState(false);
+  const resetTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const handleClick = () => {
+    setConfirmActivate(true);
+
+    if (resetTimerRef.current) {
+      clearTimeout(resetTimerRef.current);
+    }
+
+    resetTimerRef.current = setTimeout(() => {
+      setConfirmActivate(false);
+      resetTimerRef.current = null;
+    }, 4000);
+  };
+
+  useEffect(() => {
+    return () => {
+      if (resetTimerRef.current) {
+        clearTimeout(resetTimerRef.current);
+      }
+    };
+  }, []);
+
+  if (confirmActivate) {
+    return (
+      <>
+        <ReactTooltip />
+        <button
+          className={className}
+          onClick={onActivate}
+          data-tip="There is already an active asset. Continue activating?"
+          data-testid="activate-button"
+          disabled={disabled}
+        >
+          Confirm activate
+        </button>
+      </>
+    );
+  }
+
+  return (
+    <>
+      <button
+        className={className}
+        onClick={videoHasActiveAsset ? handleClick : onActivate}
+        data-testid="activate-button"
+        disabled={disabled}
+      >
+        Activate
+      </button>
+    </>
+  );
+};
--- a/public/video-ui/src/components/ActivateButton.tsx
+++ b/public/video-ui/src/components/ActivateButton.tsx
@@ -46,7 +46,7 @@ export const ActivateButton = ({
           className={className}
           onClick={onActivate}
           data-tip="There is already an active asset. Continue activating?"
-          data-testid="activate-button"
+          data-testid="confirm-activate-button"
           disabled={disabled}
         >
           Confirm activate
--- a/public/video-ui/src/components/ActivateButton.tsx
+++ b/public/video-ui/src/components/ActivateButton.tsx
@@ -5,14 +5,14 @@ type ActivateButtonProps = {
   className: string;
   onActivate: () => void;
   disabled?: boolean;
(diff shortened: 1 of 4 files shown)


public/video-ui/src/components/VideoUpload/VideoAsset.test.tsx
```diff
@@ -127,6 +125,88 @@ describe('VideoAsset', () => {
...
+            id: 'AAAAAAAAAAA'
+          },
+          {
+            ...defaultVideoAsset,
+            version: 2,
+            id: 'BBBBBBBBBBB'
+          }
+        ],
+        activeVersion: 2
+      } as Video)
```
junyuanxue: Why do we need the `as` typecast? 👀

public/video-ui/src/components/ActivateButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </button>
+      </>
+    );
+  }
+
+  return (
+    <>
+      <button
+        className={className}
+        onClick={videoHasActiveAsset ? handleClick : onActivate}
```
junyuanxue: I think we potentially mainly want to surface this if a video is already published, but I'm not against adding this for draft videos too

junyuanxue (approved):
Nice! I have a couple of comments but I don't think either is blocking ✨

public/video-ui/src/components/VideoUpload/VideoAsset.test.tsx
hk-15: We're not setting all of the required fields for Video here. They're provided by `blankVideoData` in the reducer. So I suppose the cast is suppressing the error here. https://github.com/guardian/media-atom-maker/blob/82e504c274468807b33d45c9fb9d160bf6e30dac/public/video-ui/src/slices/video.ts#L25-L26

To be perfectly honest, I was just copying the existing implementation 😅 I could import `blankVideoData` instead and use it in `setVideo` to be clearer

hk-15: Changed in [4b33ca1](https://github.com/guardian/media-atom-maker/pull/1680/commits/4b33ca1711cb958d64df10432a1970834209291f) ☺️

public/video-ui/src/components/ActivateButton.tsx
hk-15: Thank you, it's a really good point! I admit I don't really understand why a video can be `published` but not have any usages or even an active asset, but that's not really relevant here. Just something I find a bit mystifying...

Now checking both whether a video has an active asset and is published, so the only time you will have to confirm activation is when there is another active asset **and** the video is marked as `published` ✌️

Implemented in [2335464](https://github.com/guardian/media-atom-maker/pull/1680/commits/233546441e08da141d0f0c82f35c8b20a95a9833)

public/video-ui/src/components/VideoUpload/VideoAsset.test.tsx
```diff
@@ -95,7 +97,7 @@ describe('VideoAsset', () => {
         </Provider>
       );
 
-      const activateButton = screen.getByRole('button', { name: 'Activate' });
+      const activateButton = screen.getByTestId('activate-button');
```
lindseydew: Might be good to update the naming of the test and be specific that the video is not published and not active. Or even better have two tests for both states

lindseydew (approved):
Looks really good, nice work ✨

public/video-ui/src/components/VideoUpload/VideoAsset.test.tsx
```diff
@@ -127,6 +127,95 @@ describe('VideoAsset', () => {
...
+          deleteAtom: true,
+          setVideosOnAllChannelsPublic: true,
+          pinboard: true,
+          addSelfHostedAsset: true
+        }
+      })
+    );
+    storeWithActiveAsset.dispatch(
+      setVideo({
+        ...blankVideoData,
```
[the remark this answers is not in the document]
hk-15: Excellent suggestion! I have made a fixtures file and added all the video and asset variables used in this suite.

public/video-ui/src/components/VideoUpload/VideoAsset.test.tsx
```diff
@@ -127,6 +127,95 @@ describe('VideoAsset', () => {
...
+    it('does not call selectAsset when activate button is clicked once', async () => {
+      const user = userEvent.setup();
+
+      render(
+        <Provider store={storeWithActiveAsset}>
+          <Asset {...defaultProps} upload={inactiveAsset} />
+        </Provider>
+      );
+
+      const activateButton = screen.getByTestId('activate-button');
```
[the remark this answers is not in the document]
hk-15: I've done this backwards by asserting that, after the first click, the `activate` button is not shown!

